PR #10, opened 2018-10-16
Delete configure

dlebauer:
because of https://github.com/ebimodeling/biocro/issues/9

not sure why the ./configure files are required, but with recent updates to roxygen it seems to cause problems with the build.

### DIFF
--- a/configure
+++ b/configure
@@ -1,2 +1,0 @@
-R --silent --vanilla -e "roxygen2::roxygenize('.', roclets='rd')"
-
--- a/configure.win
+++ b/configure.win
@@ -1,2 +1,0 @@
-R --silent --vanilla -e "roxygen2::roxygenize('.', roclets='rd')"
-


ashiklom:
They were used to automatically build the documentation, rather than commit it manually. I support this change, but you'll also need to manually build the documentation (`devtools::document` or `roxygen2::roxygenize`). Right now, you have an empty `man` directory. https://github.com/ebimodeling/biocro/tree/master/man

ashiklom:
You'll probably also want to delete the `man/.gitkeep` file, since that's currently preventing you from committing any `.Rd` files.

justinmcgrath:
This is a problem with roxygen. The solution is to fix it in roxygen or remove roxygen from BioCro. I prefer the second option.

dlebauer:
replaced by #12

PR #11, opened 2018-10-16
Merge New framework into Master

dlebauer:
@justinmcgrath is there a reason not to merge this branch into master? My understanding is that it retains the existing functionality of v 0.95 and adds the new framework that you added in the biocro-dev version


dlebauer:
@justinmcgrath can we merge these? Could you take a look at the conflicts ... I think you are in a much better position to figure out what should stay / go

justinmcgrath:
I'm going to work with Scott to organize some things and then bring those changes to here.

dlebauer:
@justinmcgrath do you have a timeline for doing this? We are trying to determine how to proceed with our work and it isn't clear if we should be using the new framework, and if so, to which fork or branch we should contribute changes.

justinmcgrath:
It's going to take at least another week, but probably two. What changes are people planning to make? If there is little chance of merge conflicts, they could modify new-framework on ebimodiling/biocro.

dlebauer:
@justinmcgrath thanks ... collaborators are interested in using BioCro to simulate the impact of NPQ dynamics on photosynthesis.

justinmcgrath:
If they write new modules, there will be little if any chance of a merge conflict, so they could use new-framework in this repository.

That being said, the models in BioCro are poorly suited for looking at NPQ. There is no ray tracer; the canopy model has no way of tracking state related to kinetics; and the smallest time step you can use is 1 hour, which is far longer than NPQ dynamics. There are plans to improve all of those things, but each one is a large undertaking.

Yu Wang is currently developing a dynamical model to examine NPQ dynamics independently of BioCro. It's not a full crop model, but it doesn't have the restrictions BioCro has.

dlebauer:
> the smallest time step you can use is 1 hour

I wasn't aware of this limitation ... 
 
> Yu Wang is currently developing a dynamical model to examine NPQ dynamics independently of BioCro. It's not a full crop model, but it doesn't have the restrictions BioCro has.

Thanks, I'll follow up w/ Yu. I'm also considering statistical approaches that could get around some of these limitations.

dlebauer:
@justinmcgrath any updates on the status of putting the new framework into this repository?

justinmcgrath:
It has been merged into the `new-framework` branch, if you want to try it.

Merging it into master will break old versions. I recall that some other repositories are pulling directly from this, so we could either merge and force those people to convert, or try to notify everyone so they can migrate before we merge.

dlebauer:
@justinmcgrath by 'Merging it into master will break old versions' I assume that you mean that the old functions have been deprecated? I think that will be fine - I can update the PEcAn tools using an old release (e.g. 0.95?) and also can add a note to the readme about using the old version.

Also worth updating the authors and maintainers list in the description file

dlebauer:
@justinmcgrath after we merge this is the intent to regularly merge changes, bug fixes, and etc from the biocro-dev repository?

justinmcgrath:
There's another pull request with the version in the publication.

PR #13, opened 2018-10-18
run Roxygen at packaging not installation

infotroph:
* Move responsibility for running Roxygen from user to developer. This is the [standard](https://github.com/r-lib/devtools/issues/229) and widely-accepted practice: Roxygen is a tool that *helps developers* do their job of writing and delivering documentation, not a replacement for them.
* Commit all Rd files. Yes, this means we should have a commit hook or CI check to insure that they're always updated in sync with the source files, but for users this removes a *major* installation headache.
* Autogenerate the NAMESPACE file, which was a pain to do by hand.
* Update tests to work with new versions of testthat .

Known deficiencies:
* Building the `biocro-manual` vignette currently fails (true before this PR too), so may need to build/check with `--no-build-vignettes`
* Parameter sensitivity tests are failing both here and in the master branch!


justinmcgrath:
Generating the docs is currently the developers responsibility, and is
handled by the configure script, but the R build system.

Committing generated files is not standard practice.  It violates best
practices of version control, because it causes problems.

Several people have updated packages that have broken their builds. Those
errors are being reported here, even though nothing in this repository has
changed. Nothing done here  is non standard or  in  error. These things are
described in the R documentation. The  error  lies  in whatever packages
you have updated. It seems like that packages is roxygen, and it
specifically seems like the problem is that roxygen calls another package's
configure script, which is completely unacceptable behavior. Submit reports
to them.

The only change that would be made here is to remove roxygen. It does not
help developers. It is a nonstandard and unnecessary step in R's build
system. Anything that must be done, such as defining a name space file,
must also be done in roxygen. Entering dozens of export statements in a
dozen files is no easier than entering a dozen export statements into a
single file.  And then you must also run roxygen every time you make a
change.

infotroph:
> Committing generated files is not standard practice.

I'm with you that it's good to *generally* lean toward not committing generated files, but I don't know where you're getting that it's nonstandard *for Roxygen outputs*. I just searched GitHub for "RoxygenNote", sorted by most recently indexed, clicked through on the first 75 R packages in the results, and [74 of them](https://gist.github.com/infotroph/4067c8371236baefdff7c4fe5e3150a1) had their Rd files committed. Even Dirk Eddelbeuttel, as retro-grouch a best practices stickler as you'll find, consistently keeps his Rd files in Git.

> Nothing done here  is non standard or  in  error.

1. If it's so standard, show me another package that builds documentation in its configure file.
2. My reading of the R extensions manual is that the configure file is expected to be used only for detecting system-specific details that get passed along to Make (See the first paragraph of the [configure and cleanup](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Configure-and-cleanup) section), and that it is specifically required to clean up any files it writes: "N.B.: If the configure script creates files, e.g. src/Makevars, you do need a cleanup script to remove them."

> it specifically seems like the problem is that roxygen calls another package's configure script, which is completely unacceptable behavior

So what's happening here is that Roxygen now calls `pkgtools::load_all`, which tries to simulate loading the package by, among other things, compiling source files. So if it's compiling, I'd say calling configure is totally expected. What exactly is unacceptable about that?

I have seen some debate (https://github.com/klutometis/roxygen/issues/771, https://github.com/klutometis/roxygen/issues/784) whether it's necessary or advisable for Roxygen to call `load_all`, but I predict this will not change -- after all, Roxygen modifies your package, so it has a legitimate reason to load the package code and make sure everything works as expected after it does the modifications.

> The only change that would be made here is to remove roxygen. It does not help developers

I mean, it's your package and you can hand-write Rd files if you like that better, but that really doesn't square with your claim, consistent up until now, that it's vitally important to keep the code and docs in sync. What are you actually trying to say here?

justinmcgrath:
Roxygen users or a notable person committing their generated files does not make it standard or good practice. Roxygen is not part of the build system, and thus is nonstandard. The community of computer scientists in general does not commit generated files, including users of the far more popular doxygen software, because it often results in errors in the repository.

Configure files are standard in that they are a defined part of the build process, and the documentation says they are for configuration that happens before installation, which is exactly what building documents with roxygen is.

This is to keep source and generated files in sync. It's easier to do that if you remove the unnecessary step of generating documents that are just as easily written in the language that is needed. Tangentially, nothing in R or roxygen keeps code and documentation from getting out of sync.

dlebauer:
How about if we check these in and from now on the Rd files can be maintained by hand?

justinmcgrath:
That's good as long as roxygen is removed at the same time.

Rd files generated, and configure and configure.win removed.
man/.gitkeep removed.
man/* and !man/.gitkeep removed from .gitignore.
Roxygen comments removed from Rd files and DESCRIPTION.
Roxygen removed from Suggests in DESCRIPTION.
Roxygen markdown removed from R files.

dlebauer:
Before completely removing roxygen, I think it would be prudent to have feedback from other maintainers including @djaiswal and @femiguez. Do either of you have preferences on using roxygen to manage documentation vs. editing .Rd files by hand? Do you have any preferences or recommendations about other points made in this thread?

djaiswal:
@dlebauer @danwang2008 @justinmcgrath @femiguez 
I am fine with removing roxygen, and manually editing Rd files.

dlebauer:
@justinmcgrath do you want to merge the new framework before making all of these changes https://github.com/ebimodeling/biocro/pull/11 ? I think it would be much easier to do now than it will be after editing every file to remove the roxygen markup.

infotroph:
@justinmcgrath This was probably clear from my comments above, but for completeness: I think removing Roxygen will cause you more work in both the short and long run, and I will not volunteer any of my time to do the removal. After this thread, I doubt I'll be interested in helping with future changes either.

If you're going to proceed, I second @dlebauer's suggestion to merge the new framework first. That will leave the Rd files in this PR out of date and its namespace changes irrelevant, but the test updates can probably still be cherry-picked after the documentation surgery is done.

justinmcgrath:
Why do you think that Roxygen is less work? You must make exactly the same
changes to Roxygen markdown that you would need to make to Rd files or
NAMESPACE, but then you must also build Rd files from the markdown, which
adds to the work.

Merging in the new framework first is fine.

dlebauer:
From [Wilson et al "Best Practices for Scientific Computing"](https://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.1001745):

> The best way to create and maintain reference documentation is to embed the documentation for a piece of software in that software (7c). Doing this increases the probability that when programmers change the code, they will update the documentation at the same time.

Indeed, it shouldn't be too difficult to keep two text files open at the same time, but the reality is that it is enough that it makes it less likely - for many people - that they will write or update the documentation when the code is written / changed. Plus, the markdown syntax of roxygen is easier to learn than the mix of latex and bespoke conventions used in Rd files. Building the .Rd files is relatively easy using the `devtools::document()`  function.

I don't think that it is necessary to argue about the merits of roxygen here, but I think that @infotroph has a good point that it shouldn't be his responsibility to revert from roxygen to using Rd files. As the person who originally converted from using Rd to roxygen I am obviously biased, but also experienced enough to understand that it is non-trivial, and will suggest that this up-front investment should also be considered when making the choice.

justinmcgrath:
Generating documentation from comments in the code is nice if the generation program is well designed. Roxygen is not well designed though, and it incurs a cost that is larger than its benefits.

The cost is not so large that I consider it even a remote priority, or think that someone should spend time now on it. If someone is going to commit the Rd files though, then Roxygen should be removed, because committing generated files and keeping Roxygen is a worse state than having Roxygen generate them.

dlebauer:
As an update, roxygen is considered best practice and ‘requested’ by rOpenSci package guidelines, which are used by the Journals Methods in Ecology in Evolution, and Journal of Open Source Software. So independent of how well it is designed, it is clearly considered a ‘best practice’ by the R community.

PR #15, opened 2019-04-09
BioCro Docker

kimberlyh66:
Please review changes made. I have created a folder named `Docker` that contains all the files needed to dockerize BioCro.


kimberlyh66:
@dlebauer @KristinaRiemer Please review when available. I am unable to assign anyone to this pull request for some reason.

justinmcgrath:
BioCro does not depend on Docker, so unfortunately, this is not the appropriate repository for this code. From the changes listed here, this set of code seems to be standalone, and thus could be in its own repository.

FYI, two of these files do not have new line characters at the end of the file, which is nonstandard, and possibly will break some programs. You likely want to fix that and ensure that your editor is properly configured.

dlebauer:
This is intended as a feature rather than a dependency. We can put it in a separate repository though.

kimberlyh66:
Closing this pull request. Files have been put into another repository.

PR #18, opened 2022-04-11
Merge BioCro II into master

justinmcgrath:
The merge-ii branch has the BioCro 2.0 version that is included in the in silico Plant manuscript. This pull request isn't for review per se, as we do not want to make any changes to this. However, before going ahead with this, I wanted to check that everything looks good and give people a chance to point repositories to specific commit IDs if needed. After merging, we'll create a 2.0 tag, so that we have a reference to the version in the manuscript.

After this, I'd like to move from `master` to `main` so that we are in line with the new inclusive naming practice.

Shortly after this, there will be another major update that allows models to be written in separate repositories. With that in place, people will need to move their models from biocro-dev to their own repositories. We will then do all "framework" development on this branch and cease using biocro-dev.


justinmcgrath:
Unless someone has a reason to delay this, I'll merge this on Thursday.

eloch216:
I'm still checking this out, but one thing I noticed is that this branch still contains a bunch of files in the `R`, `man`, and `test` directories from BioCro I. I'll make a separate branch for removing them, which can be merged into this branch when it's ready.

justinmcgrath:
It looks like a merge problem. I'll fix it and unless I can figure out how to do this cleaning on this branch I'll make a new branch and new pull request.

eloch216:
Hmm, there are still a bunch of extra files in the `man` directory

justinmcgrath:
Those look like they were readded here after being removed, so need to be manually removed. Could you check that I removed them all and that I didn't remove any I shouldn't have?

justinmcgrath:
@gsrohde will the GitHub actions need to be updated for the new repository name?

eloch216:
> @gsrohde will the GitHub actions need to be updated for the new repository name?

I think there might be some required changes. I found "biocro-dev" in the following files:
- `.github\workflows\document.yml`
- `.github\workflows\R-CMD-check.yaml`
- `.github\workflows\README.md`
- `.github\workflows\run_bookdown_and_pkgdown.yml`

eloch216:
> Those look like they were readded here after being removed, so need to be manually removed. Could you check that I removed them all and that I didn't remove any I shouldn't have?
> 
> I found some other differences too and fixed those. Hopefully that's the last of it.

I haven't been able to find anything else that looks out of place. I ran all the tests and examples, built the vignettes, ran `R CMD check`, built the `Doxygen` documentation, and built the `bookdown` documentation, and everything seemed to work without any serious complaints.

gsrohde:
I really don't understand why you are doing this as a merge.  Don't you want the resulting version to look exactly like some version from biocro-dev (except with references to biocro-dev changed to biocro)?

justinmcgrath:
What do you suggest instead?

justinmcgrath:
This was my plan after the meeting. I'm sure I've got some git syntax issues, so these are approximate. However, after some reading, I think this may not work like we expect.

On my local biocro repository, run these commands.
git checkout  biocro-dev/yet-another-biocro-for-public
git push # To get a copy on GitHub to work with.
Scott updates this branch so the GitHub actions work.
I pull the changes.
git branch -m yet-another-biocro-for-public main
git push -u origin main
I update the default branch in GitHub.
git push origin --delete master


The description here makes me think that may cause problems: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch. It mentions updating links to match the new name. If we use the approach above, I think issues that reference master won't work anymore. That's not a big problem, but it would be a little nice to have it work.

If we do this pull request and then rename master using the GitHub interface, then the links will be updated correctly. We have the few changes here that are discarded in the history, which isn't ideal, but that seem less a problem than issue history that doesn't work correctly.

justinmcgrath:
Eh, well there are only 10 and issues and 8 pull request in total, so it's unlikely to matter. I don't feel that a bit doing and undoing in the history is too large a problem though. That sort of stuff is common enough in the history anyway.

justinmcgrath:
Maybe instead of the steps starting at `git branch -m yet-another main` this would work.

git branch -m master old-master
git branch -m yet-another master
git push -f origin master

Rename master on GitHub

justinmcgrath:
One other problem is that the most recent tag, 0.951, does not exist on yet-another-biocro. I'm again leaning back to using a merge and renaming. We would have the messy history, but git is made to record history, not modify it, and trying to think of all details involved in modifying history is getting confusing.

gsrohde:
> This was my plan after the meeting. I'm sure I've got some git syntax issues, so these are approximate. However, after some reading, I think this may not work like we expect.

> On my local biocro repository, run these commands.
```
git checkout biocro-dev/yet-another-biocro-for-public
git push # To get a copy on GitHub to work with.
```
For the push, is the default push repository here `ebimodeling/biocro`?  (And if so, where in the repository does it get pushed to, since `ebimodeling/biocro` doesn't have a `yet-another-biocro-for-public` branch?)  And haven't you already essentially done these steps?

> Scott updates this branch so the GitHub actions work.
> I pull the changes.

I take it you mean I would do this on a branch branched off whatever branch `yet-another-biocro-for-public` got pushed to.

eloch216:
> I'm again leaning back to using a merge and renaming.

I'm fine with this. There are very few "irrelevant" commits (ones that were made to the `biocro` repository since `biocro-dev` diverged from it in 2017) and I think it's unlikely that anyone will notice they are there.

gsrohde:
> Eh, well there are only 10 and issues and 8 pull request in total, so it's unlikely to matter. 

I'm not sure what you are referring to here, but I'm guessing it has something to do with problems associated with `master` no longer referring to what is currently the master branch.  But I can't see this as being a problem for the issues, and this PR is the only one of the 10 that isn't closed.

> I don't feel that a bit doing and undoing in the history is too large a problem though. That sort of stuff is common enough in the history anyway.

Not at all clear what you are talking about here.

gsrohde:
> Maybe instead of the steps starting at `git branch -m yet-another main` this would work.
> 
> ```
> git branch -m master old-master 
> git branch -m yet-another master 
> ```
Again, there is not currently a `yet-another ...` branch in `ebimodeling/biocro`.

> ```
> git push -f origin master 
> ```

What state is your working directory in here?  What is `origin` here?

> Rename master on GitHub.

What is it being renamed to?

> 
> Then we don't have the unnecessary history and we get the updated links.

gsrohde:
> One other problem is that the most recent tag, 0.951, does not exist on yet-another-biocro.

How is this a problem?

> I'm again leaning back to using a merge and renaming. We would have the messy history, but git is made to record history, not modify it, and trying to think of all details involved in modifying history is getting confusing.

The only messy history I'm proposing modifying is those last three commits (d2f6977, 64899e5, and 31c14c0)—namely, by removing them, so that commit cb635d9 (whose content exactly matches the last of the three, commit 31c14c0) is the new main development branch head.

justinmcgrath:
> How is this a problem?

After making the end of yet-another-biocro the new master, it leaves a hanging commit that is messier than having changes that undo previous changes.

> The only messy history I'm proposing modifying is those last three commits (https://github.com/ebimodeling/biocro/commit/d2f6977ba9e6c559da8e61a46f88ef8a91129017, https://github.com/ebimodeling/biocro/commit/64899e516788213bb2de3dd1cb85809624042b53, and https://github.com/ebimodeling/biocro/commit/31c14c0096fb72d5552c79714b974eb0d0353f51)—namely, by removing them, so that commit https://github.com/ebimodeling/biocro/commit/cb635d9cdc5360e8437df8433511d091db0db7e9 (whose content exactly matches the last of the three, commit https://github.com/ebimodeling/biocro/commit/31c14c0096fb72d5552c79714b974eb0d0353f51) is the new main development branch head.

These are part of the history though. The branches diverged in these regards, so after merging these changes need to be made. Messy changes happen as a consequence of diverging branches, and everything I've tried to get around that results in something arguably worse than these commits. I tried rebasing back to the merge commit thinking I could include those changes in the merge, but that records every commit on the same day and results in a lot of empty commits. I also tried branching master here, cherrying picking these commits that remove files, then rebasing to yet-another-biocro, but that also puts all of the commits on one day. As before, the options above would leave a commit without decedents (and I think maybe without a parent), which although not terribly important, is less likable than some commits that undo previous changes. Typically the recommendation is to record reversions rather than trying to hide them anyway.

Maybe there is some way to do this, but I can't figure it out, and the effort is turning out to not be worth removing two commits. Although it would be nice to not have things like this in the history, the solution is to not have independent development on branches so that these things are less likely to happen, and this merge brings us a big step closer to that.

justinmcgrath:
Combining the last two is easy! Now at least there's only one objectionable commit.

gsrohde:
> > How is this a problem?
> 
> After making the end of yet-another-biocro the new master, it leaves a hanging commit that is messier than having changes that undo previous changes.
> 
> >

If it's got a tag on it (in this case, 0.951), it's not a hanging commit; there's no danger it will be garbage collected.

It is true, however, that since there is one further commit along the current master branch _after_ the one tagged 0.951, that commit will be left hanging if we simply remove or move the master branch pointer.  That problem is very simple to solve, however, by either tagging that commit as well (e.g. do `git checkout master; git tag old_master` before moving or removing the master branch pointer), or by simply renaming the master branch.

gsrohde:
Here is what the commit tree currently looks like:

Screen Shot 2022-04-13 at 5 44 18 PM

and here is what I would like it to look like:

Screen Shot 2022-04-13 at 5 44 39 PM


(or whatever name you want to use for the branch; merge-ii no longer seems appropriate since it's no longer a merge, but renaming it is not my concern here).

To effect this doesn't seem that difficult.  Starting with a working tree that is up to date with the current `merge-ii` branch on `ebimodeling/biocro`, you would simply do
```
git reset --hard cb635d9cdc5360e8437df8433511d091db0db7e9
```

(cb635d9 is the hash id for the commit with summary "Use 2.0 instead of 2.00.", the last commit before the unnecessary merge.  As I said before, it exactly matches the current tip of the `merge-ii` branch, i.e.
```
git diff cb635d9 merge-ii
```
returns no differences (which also means the revert command will not change the working tree at all).

Then, assuming `origin` refers to `ebimodeling/biocro`, follow this by
```
git push -f origin merge-ii
```

justinmcgrath:
yet-another-biocro already points to cb635d, so the hard reset just makes merge-ii identical to that. We are then still left with the same two options: (1) merge that into master, which requires the changes in 77926 and which is also the most common way this is handled or (2) rename yet-another-biocro to master, which leaves tag 0.951 a dead end and which is not typically how changes are brought to other branches.

I too would like to not have undo changes, but that is what happens when branches diverge in ways that are not outright conflicts that can be handled during the merge. I'm assume we already have things like this in the history. I have never seen anyone handle this by renaming a branch to make it look like the files were never changed. I am reasonably sure we have nothing like that in the history. One couldn't tell of course, but I'm not aware of anyone engaging in that sort of git-fu (it would require a force push, so I think someone would notice), so it seems unlikely.

The history of master on this repository is what it is, and I don't see any reason to change it so that instead master looks like it was always yet-another-biocro, which isn't even master on biocro-dev. If anyone looks back on this, it will be confusing. There will be no merge commit for the single largest change to this code base, and they'll wonder how 0.951 got stranded. Will anyone ever look back on it? I don't know. I have occasionally had reasons to sift through the history, so it seems possible.

This also all ignores that we are only two small steps away from being done with this. Delaying this over a single commit is frustrating.

justinmcgrath:
Maybe there is just a misunderstanding about why that commit is there, so I've made this poor figure. 26e0 is biocro/master and cb635 is biocro-dev/yet-another-biocro-for-public.

SwiftScan 2022-04-13 21 02 20

gsrohde:
> yet-another-biocro already points to cb635d, so the hard reset makes merge-ii identical to that.

Yes, I know, this is what I intended.  (`merge-ii` is already identical to `yet-another-biocro...` as far as file content goes; it simply has a different hash id, different parents, and a different commit message.  So the two commits after `cb635d`—the _merge_ commit d2f6977, and the commit to undo the effects of that merge on the file content—simply bring one back to the same place, at the file content level.  Only now you have a merge history that, in the absence of any clear explanatory log comments, is, in my view, misleading as to the line of development (but see below as to how to ameliorate this).)

While I'm still not convinced it is necessary (see below), If you really want to merge `master` with `yet-another-biocro...`, there are better ways.  I'll mention just one.

First of, in spite of the log message "Merge BioCro II into master", the commit really does the opposite: it merges master into BioCro II (represented by `yet-another-biocro...`, or `merge-ii` as you call it in this repository); after the merge, master is exactly what is was before the merge.  Usually the reason to merge master into a development branch is to bring the latest changes from master into that branch.  But here, you bring in those changes only to immediately undo all of them.  

Another reason to merge master into a development branch is to make master a direct ancestor of the development branch so that later, when the development branch is merged into master, it will be a fast forward merge and there will be no conflicts.  If this was your intention, the best command to use would have been
```
git checkout merge-ii [i.e., cb635d]
git merge -s ours master
```
This merge strategy (-s ours) is a way to completely ignore the content of the master branch.  Now `git diff cb635d merge-ii` returns zero differences without the need for the subsequent "undo" commit you had to use.  The commit message for this merge should be written to clearly state that the `master` branch content was ignored and that the result of the merge has exactly the same content as `merge-ii` did before the merge.

> We are then still left with the same two options: (1) merge that into master, which requires the changes in 77926 and which is also the most common way this is handled or (2) rename yet-another-biocro to master, which leaves tag 0.951 a dead end and which is not typically how changes are brought to other branches.

A third option would be to leave master as it is, use "biocro-ii" in place of "merge-ii", and make "biocro-ii" the new default branch.  The commit tagged 0.951 will then still be on a branch and so won't be a dead end, though perhaps you object to it not being a direct ancestor of any subsequent versions that may happen to get version tags.  (I may be wrong about this, but I think in some Git workflows, it is not uncommon to have tagged versions on a "spur".)

> 
> I too would like to not have to undo changes, but that is what happens when branches diverge in ways that are not outright conflicts that can be handled during the merge. I assume we already have things like this in the history. I have never seen anyone handle this by renaming a branch to make it look like the files were never changed. I am reasonably sure we have nothing like that in the history. It wouldn't actually be "in the history", so one couldn't tell of course, but I'm not aware of anyone engaging in that sort of git-fu (it would require a force push, so I think someone would notice), so it seems unlikely.

I normally wouldn't recommend a _force push_, as I did earlier to reset the merge-ii branch head to undo the merge, but I thought in this case, it was unlikely anyone had pulled and begun any development work on branch `merge-ii`.

Renaming a branch in the repository, by the way, either by using a force push or directly on GitHub, presents some of the same problems, and so perhaps it is best to avoid any branch renaming unless absolutely necessary.

> 
> The history of master on this repository is what it is, and I don't see any reason to change it so that instead master looks like it was always yet-another-biocro, which isn't even master on biocro-dev. If anyone looks back on this, it will be confusing. There will be no merge commit for the single largest change to this code base, and they'll wonder how 0.951 got stranded. Will anyone ever look back on it? I don't know. I have occasionally had reasons to sift through the history, so it seems possible.
> 
> This also all ignores that we are only two small steps away from being done with this. Delaying this over a single commit is frustrating.

Since you are in a hurry to get this done with, I simply ask one thing: When you do the commit to merge `merge-ii` into `master` (the _pull_ of the pull request), make sure that you use a commit message that makes clear that this replaces everything on the master branch (or at least every thing between f92834e (the merge base of the _current_ master and cb635d) and the current master (26e0125)).  I don't remember if GitHub _always_ allows you to customize the commit log message when you use the _Merge pull request_ button.  I'm fairly sure that fast-forward command line merges don't by default.

I assume the second step you are referring to is updating the GitHub actions.  Correct me if I'm wrong.

There are some other issues with the content of `yet-another-biocro ...` branch that will need to be addressed.  The one I noticed is that some items were left out of the _bookdown_ book.  It would have been nice to have had the chance to thoroughly look over the content of the `yet-another-biocro ...` branch as a separate issue preceding this one to make sure that there are not other problems with it.  Perhaps you did bring this up as some point earlier and I have forgotten.  I just hope that other problems with it, if any, are, like the _bookdown_ problem, relatively minor.

eloch216:
> The one I noticed is that some items were left out of the _bookdown_ book.

What was missing?

>  It would have been nice to have had the chance to thoroughly look over the content of the `yet-another-biocro ...` branch as a separate issue preceding this one to make sure that there are not other problems with it.

It might be worth pointing that the `yet-another-biocro-for-public` branch _has_ been looked over: https://github.com/ebimodeling/biocro-dev/pull/396

justinmcgrath:
> git checkout merge-ii [i.e., cb635d]
git merge -s ours master

`-s ours` chooses files from the current branch when they conflict with the other branch. If there is not a conflict, then the changes are ... merged. If those changes are unwanted, they need to be removed. Aside from doing something incredibly unusual, merge strategies do not result in a branch fully overwriting a different branch when there are nonconflicting changes.

If a changes are committed, there are two ways to undo the: (1) a new commit that reverts the changes, which is by far the most common way to do it, because it makes the history of changes clear or (2) some altering of the history to make it look like it never happened, which is generally shunned in public branches because it's confusing, it requires a force push which is at the least annoying and possibly an incredible pain, and because sorting out how to do it is complex in all but the simplest cases.

What is being done here is the same workflow as dozens of other merges. On April 8, there is an example of the exact thing in the biocro-dev repository (https://github.com/ebimodeling/biocro-dev/network). master and separate-module-library were merged. Changes continued on separate-module-library. Then they were merged again.

master and yet-another-biocro were merged (d2f6). Changes were made (7792). We'd like more changes so that the GitHub actions work. Then we can merge it into master. That would be an identical workflow to what occurred starting on April 8 in biocro-dev. We don't want to do this stuff on the master branch. It would be an least unorganized if not unusable state until it was finished.

justinmcgrath:
> First of, in spite of the log message "Merge BioCro II into master",

The title of this pull request is what it does when it is completed.

gsrohde:
> It might be worth pointing that the `yet-another-biocro-for-public` branch _has_ been looked over: [ebimodeling/biocro-dev#396](https://github.com/ebimodeling/biocro-dev/pull/396).

Yes, I think I vaguely remembered this.  I may have stopped looking it over when it was changed to a draft PR.

gsrohde:
> What was missing?

I think I was mistaken about this.  I had built the _bookdown_ book from a version of `yet-another-biocro-for-public` that wasn't up to date.

gsrohde:
> `-s ours` chooses files from the current branch when changes between braches conflict. If there is not a conflict, then the changes are ... merged. If those changes are unwanted, they need to be removed. Aside from doing something incredibly unusual, merge strategies do not result in a branch fully overwriting a different branch when there are nonconflicting changes.

Not true.  The documentation from `get help merge` states
```
ours
  This resolves any number of heads, but the resulting tree of the merge
  is always that of the current branch head, effectively ignoring all 
  changes from all other branches. It is meant to be used to supersede
  old development history of side branches. Note that this is different 
  from the -Xours option to the recursive merge strategy.
```
I tested this out myself on a copy of `ebimodeling/biocro` with `merge-ii` set to where it was before the merge with master to verify this is indeed what happens.

You may be getting this confused with the _recursive_ ours strategy, which is explained as follows:
```
ours

  This option forces conflicting hunks to be auto-resolved
  cleanly by favoring our version. Changes from the other tree
  that do not conflict with our side are reflected in the merge
  result. For a binary file, the entire contents are taken from
  our side.

  This should not be confused with the ours merge strategy, which
  does not even look at what the other tree contains at all. It
  discards everything the other tree did, declaring our history
  contains all that happened in it.
```

gsrohde:
[Some of this may be moot since I see you've already closed this, but some of it you may want to read for future reference and please at least read the last paragraph.]

> If a changes are committed, there are two ways to undo them: (1) a new commit that reverts the changes, which is by far the most common way to do it, because it makes the history of changes clear and git commands are specifically designed to facilitate that workflow or (2) some altering of the history to make it look like it never happened, which is generally shunned in public branches because it's confusing, it requires a force push which is at the least annoying and possibly an incredible pain, and because sorting out how to do it is complex in all but the simplest cases.

My point was that I saw no really good reason to merge `master` into the `yet-another-biocro-for-public` branch (commit cb635d9) to begin with since that commit already contained mostly what was wanted.  But what is done is done, and I do see that there are entirely valid arguments for not altering the repository history once it has been pushed to a public repository.

I can't imagine that merge was easy to do, since when I tried duplicating it on a copy (with `merge-ii` reset to where it was before the merge), I got numerous conflicts.  That is why I pointed out (for future reference, I guess) that if you really were adamant about merging in the `master` branch, the _ours_ strategy would have made it much much easier.

Even after doing this merge, however, you could have saved yourself some trouble in restoring `merge-ii` to how it looked before the merge by using the `git revert` command.  Specifically, at the point where `merge-ii` pointed to commit d2f6977 (the merge commit), you could have followed this by
```
git revert d2f6977b -m 2
```
This would undo the effects of the merge on the file content in `merge-ii`, just as the commit you _did_ do (77926ba) does.  I am presuming that you went to at least a little trouble to "Remove old files" and "Update showSoilType.R", but perhaps I am presuming too much.  In any case, the revert command would do the same thing automatically.

The "2" in this command, by the way, signifies which side of the merge we want to restore.  In this case, it is 2 because as the `git log` output for `d2f6977b`,
```
commit d2f6977ba9e6c559da8e61a46f88ef8a91129017
Merge: 26e0125b cb635d9c
Author: Justin McGrath <justinmcgrath@users.noreply.github.com>
Date:   Tue Apr 12 11:02:39 2022 -0500

    Merge BioCro II into master
```
shows, the commit content we want to keep (cb635d9c) is the second one listed.

Again, I mention this for future reference, since I'm no longer advocating making any "force pushes" to the GitHub repository.  I will be reasonably content as long as the log comment for the (fast-forward) merge of `merge-ii` into master clearly indicates that this merge brings the master branch into conformity with the last commit on the `yet-another-biocro-for-public` branch of the `biocro-dev` repository, or perhaps even if it just says something like "replace recent changes to the master branch with the new BioCro-II version of the code".

justinmcgrath:
> I can't imagine that merge was easy to do, since when I tried duplicating it on a copy (with merge-ii reset to where it was before the merge), I got numerous conflicts

git merge -X ours dev/yet-another-biocro-for-public
git rm `git status | grep delete | awk '{print $4}'`
git merge --continue

Without doubt, knowing that `git merge -s ours` will accomplish the same thing with one fewer commit will never cumulatively save the time spent on this.

justinmcgrath:
I don't want to sound ungrateful. I appreciate the work that you do. However, this much effort to save one commit was not a good use of anyone's time. Whether we used this method or the other one, the version on master that gets tagged 2.0 is the same. This much effort to turn a 2022 commit merge into a 2021 one is not sensible.

justinmcgrath:
Looking at the other branch, it does odd things to the history. When using `git log --graph --online --simplify-by-decoration` the tags are no longer in the order they were created.

Trying to sort out goes one with you modify histories is too complicated. Let's just do this the standard way, get the actions to work on this branch, and merge it.

gsrohde:
> Looking at the other branch, it does odd things to the history. When using `git log --graph --online --simplify-by-decoration` the tags are no longer in the order they were created.

That's odd.  I get exactly the same output from both branches (except for the first line of course).  In both cases, the tags appear in reverse numerical order (which I assume to also be the reverse of the creation order).

From `origin/for-merge`:
```
% git log --graph --oneline --simplify-by-decoration origin/for-merge
* 9e6dee48 (HEAD -> for-merge, origin/for-merge) Merge branch 'master' into for-merge
* 26e0125b (origin/master, origin/HEAD, master) Update README.md
* fb1f88ce (tag: 0.951) Update AuxBioCro.c
* d9301557 (tag: 0.95) minor cleanup of functions to prevent errors
* f164c1eb (tag: v0.93) bumped version
* c303949c (tag: v0.91) phenology brack of Dan and master branch are merged manually
*   dc4a49fc Merge branch 'phenology2'
|\  
| * 7e5dbeb2 (tag: v0.9) add leaf death rate and temp threshold
| * 107b23ce (tag: v0.81) add growth resp terms
* | 94c4d905 (tag: v0.8) add growth resp terms
|/  
* 5f0bb2fb (tag: v0.75) changing the initial conditions and enable parenating; adding Tbase
* bdcee4c4 (tag: v0.7) This version is calling willowGro.C in willowGro.R
* b840a7c7 (tag: v0.6) willowGro, as a copy of MisGro, is added. Documentation is missing.
* e428b4e4 Initial commit
```
From `origin/merge-ii`:
```
% git log --graph --oneline --simplify-by-decoration origin/merge-ii
* 77926ba6 (origin/merge-ii) Remove old files. Update showSoilType.R
* 26e0125b (origin/master, origin/HEAD, master) Update README.md
* fb1f88ce (tag: 0.951) Update AuxBioCro.c
* d9301557 (tag: 0.95) minor cleanup of functions to prevent errors
* f164c1eb (tag: v0.93) bumped version
* c303949c (tag: v0.91) phenology brack of Dan and master branch are merged manually
*   dc4a49fc Merge branch 'phenology2'
|\  
| * 7e5dbeb2 (tag: v0.9) add leaf death rate and temp threshold
| * 107b23ce (tag: v0.81) add growth resp terms
* | 94c4d905 (tag: v0.8) add growth resp terms
|/  
* 5f0bb2fb (tag: v0.75) changing the initial conditions and enable parenating; adding Tbase
* bdcee4c4 (tag: v0.7) This version is calling willowGro.C in willowGro.R
* b840a7c7 (tag: v0.6) willowGro, as a copy of MisGro, is added. Documentation is missing.
* e428b4e4 Initial commit
```

gsrohde:
> However, this much effort to save one commit was not a good use of anyone's time.

Saving one commit was really only a very small part of this for me.  It was more about trying to understand what you were doing, especially, why you thought it necessary to merge the BioCro II and the master branches together, and, given that you did think this necessary, why you weren't using the simpler built-in methods to do so, or to make you aware that there are such methods if you weren't already.  I also wanted to become more comfortable with some of the Git commands that I don't routinely use so that doing more complicated things would seem less daunting.

justinmcgrath:
Ok, I think we're ready. I also changed the version to 2.0.0 to match what Ed is using on biocro-dev.

PR #21, opened 2022-04-15
Port GitHub workflow actions

gsrohde:
This PR changes all references to `biocro-dev` in the workflows and in the documentation built by the workflows  to `biocro`.  Since all of these references were in documentation, documentation-related links, comments, or portions of code where the name used is arbitrary, these changes were not really necessary to get the GitHub actions working.  What _was_ necessary was to add public and private keys on GitHub to allow this repository's actions to push to `ebimodeling/biocro-documentation`.

The only reference to `biocro-dev` that I did not eliminate was the one in `LICENSE`, since its use there is not related to the workflows in any way and since its use there makes sense, even if it may now be unnecessary.  In any case, changes to `LICENSE`, if needed, should be part of a separate issue and PR.

In addition, I changed all workflows so that they are triggered by a push to any branch.  This was necessary in order to test them (I think), because it seems that workflows on a non-default branch cannot be triggered manually until they have been merged into the default branch.  (I'm not absolutely positive this is true, but in any case, I couldn't figure out how to trigger them manually.)  Since (if I'm not mistaken) workflow runs on public repositories are not time-limited, perhaps this change can be permanent.

All three workflows ran successfully.

This all brings up some issues about the workflows, such as when exactly they _should_ be triggered, whether we want to check all platforms now by default, whether we should now disallow `biocro-dev` to overwrite the documentation at `ebimodeling/biocro-documentation`, and so on.  I will perhaps make a separate issue for these.

The target branch for this PR is `merge-ii`, but if that gets merged into `master` first, the target can be changed to `master`.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -7,7 +7,7 @@
 # time ("workflow_dispatch" in the YAML file).
 
 # The workflow can be run manually from
-# https://github.com/ebimodeling/biocro-dev/actions?query=workflow%3AR-CMD-check.
+# https://github.com/ebimodeling/biocro/actions?query=workflow%3AR-CMD-check.
 # For a manual run, the user can specify different eight yes/no (or
 # y/n) options.  (Note that these input values are case insensitive.)
 
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -78,8 +78,6 @@ name: R-CMD-check
 
 on:
   push:
-    branches:
-      - master
   workflow_dispatch:
     # When running manually, allow users to customize the run:
     inputs:
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -24,7 +24,7 @@ files to a _different_ repository (namely,
 "ebimodeling/biocro-documentation"); see steps 2 and 6 below.
 
 2. The private key was added as a _secret_ in the
-`ebimodeling/biocro-dev` repository (*this* repository) under the key
+`ebimodeling/biocro` repository (*this* repository) under the key
 `PRIVATE_SSH_KEY`.  (See
 https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository.
 The key name matches the reference `secrets.PRIVATE_SSH_KEY` used in
@@ -44,8 +44,8 @@ https://ebimodeling.github.io/biocro-documentation/.
 
 6. The *public* SSH key was added as a deploy key to the
 "ebimodeling/biocro-documentation" repository under the name "Access
-from biocro-dev action".  (See
+from ebimodeling/biocro actions".  (See
 https://docs.github.com/en/developers/overview/managing-deploy-keys#setup-2.
-The name "Access from biocro-dev action" is for informational purposes
+The name "Access from biocro action" is for informational purposes
 only and has no programmatic significance.)
 
--- a/.github/workflows/document.yml
+++ b/.github/workflows/document.yml
@@ -136,7 +136,7 @@ jobs:
         cd target
 
         # user.name is somewhat arbitrary but must be provided:
-        git config user.name biocro-dev-action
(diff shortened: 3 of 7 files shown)


justinmcgrath:
Great. Thank you.

> The only reference to biocro-dev that I did not eliminate was the one in LICENSE, since its use there is not related to the workflows in any way and since its use there makes sense, even if it may now be unnecessary. In any case, changes to LICENSE, if needed, should be part of a separate issue and PR.

We can leave as-is for now and change this when biocro-dev is out of use.

> This all brings up some issues about the workflows, such as when exactly they should be triggered, whether we want to check all platforms now by default, whether we should now disallow biocro-dev to overwrite the documentation at ebimodeling/biocro-documentation, and so on. I will perhaps make a separate issue for these.

I suspect we want only this repository to update the documentation. That probably makes it more difficult to check changes before they go to the main page, so I'll defer to those who do most of the doc writing.

It would be nice to check all platforms so that we don't need to do it manually. Checking pull requests after any changes would be nice, but not the other branches. Then we'd probably get a lot of notifications about failed checks for works in progress.

gsrohde:
> I suspect we want only this repository to update the documentation. That probably makes it more difficult to check changes before they go to the main page, so I'll defer to those who do most of the doc writing.

There are two different but related issues here.  One is that when the documentation action is run on a branch, the generated documents replace whatever was there before but may not be consistent with the latest released version.  The second, _new_, issue is that documentation workflows for `biocro-dev` and `biocro` write to the same place.

The latter problem could easily be solved by making `biocro-dev` write to a different place—say, a repository called `ebimodeling/biocro-dev-documentation`, with corresponding deployment site `https://ebimodeling.github.io/biocro-dev-documentation/`.

The main use, it seems to me, for the ability to generate documentation from a branch other than the primary one is to be able to easily test out revisions to the documentation itself: it's slightly easier to simply run a GitHub workflow action and generate everything than to run the various commands required to generate the documents on one's own machine.  Probably simply having one place for the "official" documentation, corresponding to the latest released BioCro version, and another place for "changes-in-progress" documentation would suffice.  The latter documentation could either be written to a dedicated subdirectory of https://ebimodeling.github.io/biocro-documentation/ or to a separate repository altogether.

For the sake of completeness, I think it's worth pointing out that the PEcAn project maintains multiple versions of documentation.  See https://github.com/PecanProject/pecan-documentation and its corresponding deployment site https://pecanproject.github.io/documentation.html.  Unlike the "branch" documentation case I just mentioned, these are all "official" versions being documented.  This may go beyond what we need, at least for right now.

eloch216:
> Since we want https://ebimodeling.github.io/biocro-documentation/ to be the documentation for `main` of this repository, let's not have it automatically replace that when other branches are modified. Can you have a different URL for automatic generation from other branches, and have changes to `main` update the "official" URL?

If we need to make a new repository to accomplish this, this could be a good opportunity to create a new BioCro GitHub organization.

gsrohde:
> Can you have a different URL for automatic generation from other branches, and have changes to `main` update the "official" URL?

Yes.

gsrohde:
> > Since we want https://ebimodeling.github.io/biocro-documentation/ to be the documentation for `main` of this repository, let's not have it automatically replace that when other branches are modified. Can you have a different URL for automatic generation from other branches, and have changes to `main` update the "official" URL?
> 
> If we need to make a new repository to accomplish this, this could be a good opportunity to create a new BioCro GitHub organization. Ultimately, it would be nice to have all the "official" BioCro-related repositories in one organization.

Agreed.  Is this relatively simple to do?  Can `ebimodeling/biocro` just be transferred to a new organization, keeping issues intact and with automatic URL forwarding?

justinmcgrath:
> Agreed. Is this relatively simple to do? Can `ebimodeling/biocro` just be transferred to a new organization, keeping issues intact and with automatic URL forwarding?

The entire GitHub repository gets relocated, so we'd keep issues. Here it says that URLs are redirected: https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository.

PR #22, opened 2022-05-05
Fix issues related to new versions of R and knitr

eloch216:
Since the releases of `R` version `4.2.0` and `knitr` version `1.39`, there are a few small changes required to get the package to compile and the vignettes to build.

As a side note, when I ran the `pkgdown` workflow on `main`, the vignettes were not successfully produced (probably because BioCro didn't compile) but the workflow claimed that it had no problems.

As another side note, when I published this branch, three workflows were triggered: the one that runs `R CMD check`, the one that generates `doxygen` documentation, and the one that generates `pkgdown` documentation. I don't think this is the behavior we want. Typically we wouldn't want the documentation for a feature branch to automatically overwrite the documentation for the stable version.

### DIFF
--- a/src/state_map.h
+++ b/src/state_map.h
@@ -5,6 +5,7 @@
 #include <vector>
 #include <set>
 #include <string>
+#include <stdexcept>  // for std::out_of_range
 #include <algorithm>  // for std::sort
 
 /**
--- a/vignettes/a_practical_guide_to_biocro.Rnw
+++ b/vignettes/a_practical_guide_to_biocro.Rnw
@@ -2,13 +2,23 @@
 %\VignettePackage{BioCro}
 %\VignetteEngine{knitr::knitr}
 
-% See
-% https://tex.stackexchange.com/questions/148188/knitr-xcolor-incompatible-color-definition
-% for why this is needed:
-<<preliminary-hooks,include=FALSE,error=TRUE>>=
-knitr::knit_hooks$set(document = function(x) {
-    sub('\\usepackage[]{color}', '\\usepackage[dvipsnames,table]{xcolor}', x, fixed = TRUE)
-})
+% In this document, the `hyperref` latex package requires the `xcolor` package,
+% which causes conflicts with `knitr`. Prior to version 1.39 of `knitr`, `knitr`
+% automatically loaded the `color` package, which introduced incompatible color
+% definitions. Starting with version 1.39, `knitr` loads the `xcolor` package
+% instead, which can cause problems because the default `xcolor` options
+% specified by `knitr` do not match the `xcolor` options required by `hyperref`.
+% See the following pages for more information:
+% - https://tex.stackexchange.com/questions/148188/knitr-xcolor-incompatible-color-definition
+% - https://github.com/yihui/knitr/commit/b31228012419e0365ff33e18ef53fd7be7148e38
+<<solve_latex_color_problems,include=FALSE,error=TRUE>>=
+if (packageVersion('knitr') >= "1.39") {
+  knitr::opts_knit$set(latex.options.xcolor = 'dvipsnames')
+} else {
+  knitr::knit_hooks$set(document = function(x) {
+    sub('\\usepackage[]{color}', '\\usepackage[dvipsnames]{xcolor}', x, fixed = TRUE)
+  })
+}
 @
 
 \documentclass[10pt,letterpaper,oneside]{article}
--- a/vignettes/an_introduction_to_biocro.Rnw
+++ b/vignettes/an_introduction_to_biocro.Rnw
@@ -2,13 +2,18 @@
 %\VignettePackage{BioCro}
 %\VignetteEngine{knitr::knitr}
 
-% See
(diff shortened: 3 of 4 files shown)


eloch216:
Hopefully we can merge this PR ASAP because the version on `main` does not currently compile using the latest version of R.

justinmcgrath (approved):

gsrohde:
[I couldn't get to this earlier but want to make a few comments.]

> Since the releases of `R` version `4.2.0` and `knitr` version `1.39`, there are a few small changes required to get the package to compile and the vignettes to build.

These changes are fine for now, but in my comments to [ebimodeling/biocro-dev PR #433](https://github.com/ebimodeling/biocro-dev/pull/433), I plan to advocate a simpler fix.
 
> As a side note, when I ran the `pkgdown` workflow on `main`, the vignettes were not successfully produced (probably because BioCro didn't compile) but the workflow claimed that it had no problems.

Which workflow run was this?

> 
> As another side note, when I published this branch, three workflows were triggered: the one that runs `R CMD check`, the one that generates `doxygen` documentation, and the one that generates `pkgdown` documentation. I don't think this is the behavior we want. Typically we wouldn't want the documentation for a feature branch to automatically overwrite the documentation for the stable version.

I made an issue for this.

eloch216:
> > As a side note, when I ran the `pkgdown` workflow on `main`, the vignettes were not successfully produced (probably because BioCro didn't compile) but the workflow claimed that it had no problems.
> 
> Which workflow run was this?

This one: https://github.com/ebimodeling/biocro/actions/runs/2278273540. Its output has since been replaced. It successfully rendered the "BioCro User Manual" web page, but the links to the vignette PDFs were all dead. I am guessing this is because they couldn't be built, which I would guess was due to a compilation error, but I'm not sure.

gsrohde:
> > > As a side note, when I ran the `pkgdown` workflow on `main`, the vignettes were not successfully produced (probably because BioCro didn't compile) but the workflow claimed that it had no problems.
> > 
> > 
> > Which workflow run was this?
> 
> This one: https://github.com/ebimodeling/biocro/actions/runs/2278273540. Its output has since been replaced. It successfully rendered the "BioCro User Manual" web page, but the links to the vignette PDFs were all dead. I am guessing this is because they couldn't be built, which I would guess was due to a compilation error, but I'm not sure.

Yes, the PDF didn't get built because of the `knitr` upgrade.  The commands to build them are in `try` statements (for no good reason, I've decided), which is why the workflow succeeded.  The workflow _truly_ succeeded (including building all vignettes) when it ran on the same source tree version on April 18, which was before the release of `knitr` 1.39.

eloch216:
> Yes, the PDF didn't get built because of the `knitr` upgrade. The commands to build them are in `try` statements (for no good reason, I've decided), which is why the workflow succeeded.

I think those `try` statements are important for the "trial build" involved in the `R-CMD-check` workflow. The trial happens before the package is built, so it always results in an error.

https://github.com/ebimodeling/biocro/blob/c6608371983fb104516a921f2293b2d031ef3b99/.github/workflows/R-CMD-check.yaml#L356-L361

gsrohde:
> > Yes, the PDF didn't get built because of the `knitr` upgrade. The commands to build them are in `try` statements (for no good reason, I've decided), which is why the workflow succeeded.
> 
> I think those `try` statements are important for the "trial build" involved in the `R-CMD-check` workflow. The trial happens before the package is built, so it always results in an error.
> 
> https://github.com/ebimodeling/biocro/blob/c6608371983fb104516a921f2293b2d031ef3b99/.github/workflows/R-CMD-check.yaml#L356-L361

The way things used to work, after I wrapped the `pdflatex` commands in `try` calls, is that the workflow would build the LaTeX files from the `Rnw` files, and then `pdflatex` would fail to compile these LaTeX files, but not before installing any needed LaTeX packages.  This made everything ready and all hunky-dory for the _real_ vignette building that (optionally) happens as part of the package _check_ step.

But then, less than a month after I _fixed_ the check workflow, I broke it again when I set `error=FALSE` on the chunks in the vignettes that load the BioCro library.  (The reasons for doing this are explained in the portion of [this very long comment](https://github.com/ebimodeling/biocro-dev/pull/389#issuecomment-1005084684) that begins "If you don't have lattice installed ...".)  This had the effect of making the "BioCro not found" error a fatal error, printed to the console, instead of just an error message appearing in the finished vignette.  Consequently, the LaTeX files weren't even getting built, and so the needed LaTeX packages didn't get installed, resulting in the "missing package" errors later on in the workflow.

It took me the longest time to figure out what had been the point of the `try(...)` calls if the trial vignette builds always failed before reaching those calls.

The `try(...)` calls, as far as I can tell, were necessitated by the `\includegraphics` calls in the _Quantitative Comparison_ vignette.  This is because those `pdf_print` calls that depended on BioCro being installed failed.  These weren't fatal errors, but the attempt at the LaTeX compilation stage to include a graphic that didn't exist was.

I've come up with what I think is a better solution to all this and will make a PR to implement it soon.

PR #25, opened 2022-06-14
Update installation instructions

eloch216:
Since this is a public repository, it's now very simple to install BioCro directly from GitHub using `remotes::install_github`, as was pointed out in issue https://github.com/ebimodeling/biocro/issues/24.

I think it makes sense to make this change here rather than on `biocro-dev` since it is only relevant on this repository, and it would be beneficial to get it in place sooner instead of waiting for the next official version to become public.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,21 +27,22 @@ There are parameters and modules for soybean (_Glycine max_), miscanthus (_Misca
 - On MacOS, Xcode.
 
 #### Installation steps
-Download the BioCro source code from GitHub, unzip the file, and install from the either the command line or from within R using one of the following sets of commands.
 
-Do one of the sets of instruction below. These assume that the source files are in a directory named "biocro".
+- BioCro can be installed directly from GitHub by typing the following in the R command line: `remotes::install_github('ebimodeling/biocro')`. Note that this method requires the `remotes` R package.
 
-- From the command line
-```
-cd path_to_unzipped_directory
-R CMD INSTALL biocro
-```
+- Another option is to download the BioCro source code from GitHub, unzip the file (if necessary), and install the package using one of the following sets of commands, which both assume that the source files are in a directory named "biocro":
 
-- Or from within R
-```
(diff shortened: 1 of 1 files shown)


justinmcgrath:
Although I don't particularly care, I feel like this change falls under "more stuff to maintain" and "not our responsibility". When I first wrote these instructions, people asked why I didn't point out `devtools::install_github`. Now there's `remotes`. There are also other packages to install software, and this stuff will continue to change.

My thought is that these instructions are meant to work for everyone, long into the future, so that the maintenance burden is low. If someone uses `devtools` or `remotes` or one of the others, then they already know how to use them. We shouldn't need to tell them, and it is not our responsibility to document the many other packages out there or to keep up with their changes.

justinmcgrath (approved):

eloch216:
I understand your point, but I think it's also important to make it as easy as possible for people to install BioCro. I reworded the instruction a bit to make it clear that we're just providing an example of one way to directly install from GitHub, and I used `devtools` rather than `remotes` because that seems more established and less likely to change in the future.

gsrohde:
> I used `devtools` rather than `remotes` because that seems more established and less likely to change in the future.

Thinking about the relative sizes of the packages, I decided to check the installed size on my system and was surprised to find `remotes` was larger than `devtools`, 636K vs. 572K—surprised because `devtools` seemed to do so much more.  Then I found this statement in the `remotes` DESCRIPTION file: 
> _This package provides the `install_*` functions in 'devtools'. Indeed most of the code was copied over from 'devtools'._

Sure enough, `devtools` _imports_ `remotes`.  So basically, `remotes` was spun off from `devtools`.

I think I became familiar with `remotes` through writing the _R-CMD-check_ workflow: the code I modelled it after from the GitHub `r-lib/actions` repository uses it, presumably because it is cheaper than installing all of what `devtools` offers.  (Note that the `r-lib` GitHub organization also hosts the source code for `remotes` and `devtools`.)

PR #29, opened 2023-10-25
Merge iii

justinmcgrath:
This branch is a copy of biocro-dev/master. This pull request is to ensure that everything is working as expected before merging it to main here, but is not meant to undergo a typical review as that has been done already on biocro-dev.

PR #30, opened 2023-10-27
Merge resolution

gsrohde:
This PR

1.  resolves conflicts between the merge-iii branch and the main branch, mostly in favor of the former except in cases of references to the repository name;
2. changes references to "biocro-dev" to "biocro", except in one case where the reference is to a pull request in the "ebimodeling/biocro-dev" repository;
3. bumps the minimum required R version from 3.5 to 3.6, both in the DESCRIPTION file, the README.md file, and the workflow files.

(The testthat package's latest release, on Oct. 1, requires R >= 3.6.)

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -39,14 +39,14 @@ on:
         description: 'Platform to check'
         type: choice
         options:
-          - R 3.5 on Ubuntu
+          - R 3.6 on Ubuntu
           - R release version on Ubuntu
           - R devel version on Ubuntu
-          - R 3.5 on macOS
+          - R 3.6 on macOS
           - R release version on macOS
-          - R 3.5 on Windows
+          - R 3.6 on Windows
           - R release version on Windows
-        default: R 3.5 on Ubuntu
+        default: R 3.6 on Ubuntu
 
 jobs:
 
@@ -70,11 +70,11 @@ jobs:
           # This filter uses the JMESPath query language to modify the
           # strategy matrix (see https://jmespath.org/).  Note that if
           # the event is anything other than 'workflow_dispatch', the
-          # filter condition is true only when configName is 'R 3.5 on
+          # filter condition is true only when configName is 'R 3.6 on
           # Ubuntu'.
           #
           # For the public version of this repository, the "&&
-          # configName = 'R 3.5 on Ubuntu'" clause will be eliminated
+          # configName = 'R 3.6 on Ubuntu'" clause will be eliminated
           # so that when the event is anything other than
           # 'workflow_dispatch', the filter condition is always true
           # and nothing gets filtered out: all the configurations will
@@ -85,25 +85,25 @@ jobs:
           # matrix.  Thus, if the result of the filter is
           # [
           #   {
-          #     "configName": "R 3.5 on Ubuntu",
+          #     "configName": "R 3.6 on Ubuntu",
           #     "os": "ubuntu-20.04",
-          #     "r": "3.5"
+          #     "r": "3.6"
           #   }
           # ],
           # the actual output will be
           # { "include":
           #     [
           #       {
-          #         "configName": "R 3.5 on Ubuntu",
+          #         "configName": "R 3.6 on Ubuntu",
           #         "os": "ubuntu-20.04",
-          #         "r": "3.5"
+          #         "r": "3.6"
           #       }
           #     ]
           # }.       
           filter:
             '[?configName == `${{ github.event.inputs.platform-to-check }}`
                 || `${{ github.event_name }}` != ''workflow_dispatch''
-                    && configName == ''R 3.5 on Ubuntu'']'
+                    && configName == ''R 3.6 on Ubuntu'']'
 
   # For debugging:
   display_matrix:
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -21,7 +21,7 @@ using the command `ssh-keygen -t rsa -b 4096`.  (Use the `-C` option
 to include a comment.  Use an empty pass-phrase) This key pair is used
 in order to allow a workflow defined in _this_ repository to push
 files to a _different_ repository (namely,
-"ebimodeling/biocro-dev-documentation"); see steps 2 and 6 below.
+"ebimodeling/biocro-documentation"); see steps 2 and 6 below.
 
 2. The private key was added as a _secret_ in the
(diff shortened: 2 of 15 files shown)


gsrohde:
A few additional notes:

1. There are some issues with the documentation workflow that will need to be resolved.  I don't think it makes much sense to spend much time on this until we move over to `biocro/biocro`.  I don't think we need to maintain documentation for the public version both in a `ebimodeling` organization repository and in a `biocro` organization repository.
2. The reference to the pull request I mention in the note above won't be viewable by the general public since the PR itself lives in `ebimodeling/biocro-dev`.  We may want to change the note in this part of the documentation.

I expect there are many more issues that will crop up and need to be resolved, but I mention these two because the are top of mind at the moment.

eloch216:
This all looks good to me. A few small comments:
- I think we could just remove the reference to the PR from `biocro-dev`. The instructions for truncating filenames are probably clear enough without the example.
- I was hoping to retain the installation instructions that use `remotes::install_github`, but apparently that R function won't work if the repository contains a submodule.
- I agree that it's not necessary to fix up the documentation workflows until this gets moved to the new organization.
- I noticed that there are some comments like `<!-- CHANGE THIS URL WHEN MOVING FROM ebimodeling/biocro TO ebimodeling/biocro!!! -->`. I guess we are leaving them in until we move to `biocro/biocro`?

eloch216 (approved):

gsrohde:
> This all looks good to me. A few small comments:
> 
> * I think we could just remove the reference to the PR from `biocro-dev`. The instructions for truncating filenames are probably clear enough without the example.

I went ahead and removed this.

> * I was hoping to retain the installation instructions that use `remotes::install_github`, but apparently that R function won't work if the repository contains a submodule.

I think it is _supposed_ to work (as of `remotes v2.0.0`), but I couldn't get it to.  See https://github.com/r-lib/remotes/issues/82, https://github.com/r-lib/remotes/pull/103, and https://github.com/r-lib/remotes/commit/09271727977bbb791b50c5e166813218b959ce1f.  I may try to debug this.  (Where are the installation instructions that reference `remotes::install_github`?)

> * I noticed that there are some comments like `<!-- CHANGE THIS URL WHEN MOVING FROM ebimodeling/biocro TO ebimodeling/biocro!!! -->`. I guess we are leaving them in until we move to `biocro/biocro`?

Yes, I think we can leave these for now.

justinmcgrath:
GitHub Desktop seems unable to handle the submodule properly, and whatever it did broke the repository. I needed to delete it, set it up again in Desktop, then from the command line run `git checkout merge-resolution` `git submodule update --init`. Doing it any other way caused errors. Maybe that's because the submodule is only in this branch, but I'm not sure.

This now requires testthat >= 3.2.0.  I can install and test this on Windows, but not on Linux because it is an LTS and testthat isn't on CRAN with the installed R version, and the ridiculous number of dependencies in every tidyverse package make it untenable to install them all manually. The tests that do run all pass except the ones that use testthat 3.2. The summary says there are 2 warnings, but it's not listing them.

I'm happy to wait till it's move to finish up the documentation.

justinmcgrath:
This got merged before my comment. We can update DESCRIPTION in merge-iii.

eloch216:
> GitHub Desktop seems unable to handle the submodule properly

I also had some problems with GitHub desktop. It's strange because GitHub Desktop was able to handle the submodule on `biocro-dev`. I'm not sure what might be different here. Hopefully the issue will go away when the submodule is present on the `main` branch.

> I think it is _supposed_ to work (as of `remotes v2.0.0`), but I couldn't get it to. See [r-lib/remotes#82](https://github.com/r-lib/remotes/issues/82), [r-lib/remotes#103](https://github.com/r-lib/remotes/pull/103), and [r-lib/remotes@0927172](https://github.com/r-lib/remotes/commit/09271727977bbb791b50c5e166813218b959ce1f).

I found these too. Although the issues are supposedly resolved, the help menu entry for `remotes::install_github` in version 2.4.2 of `remotes` says "If the repository uses submodules a command-line git client is required to clone the submodules." I also confirmed through testing that it does not work.

> Where are the installation instructions that reference `remotes::install_github`?

They can be found in the main `README` on the `main` branch of `ebimodeling/biocro`. Also see this PR: https://github.com/ebimodeling/biocro/pull/25

gsrohde:
> I also had some problems with GitHub desktop. It's strange because GitHub Desktop was able to handle the submodule on `biocro-dev`. I'm not sure what might be different here. Hopefully the issue will go away when the submodule is present on the `main` branch.

It may make a difference whether you are cloning (i.e., starting fresh) or updating.  Working from the command line, I was always having to delete the contents of the `src/framework` directory when checking out a branch using the submodule when on a branch that didn't (and maybe vice-versa as well).  It may be a similar story when working with GitHub desktop.

gsrohde:
> This now requires testthat >= 3.2.0. 

If true, we should probably update that in `DESCRIPTION`.

> I can install and test this on Windows, but not on Linux because it is an LTS and testthat isn't on CRAN with the installed R version, and the ridiculous number of dependencies in every tidyverse package make it untenable to install them all manually. The tests that do run all pass except the ones that use testthat 3.2. The summary says there are 2 warnings, but it's not listing them.

I guess you are saying that you can't install and test on Linux because you don't _want_ to install all that is required for using `testthat 3.2`?  There are install options that make this relatively easy, I think, though the installation may take a while, and if you don't want all the dross entailed living on your machine, that's another matter.  The good news is that with the repository being public, we can run the check workflow without having to worry about going over some arbitrary monthly time limit (I think).  And the check workflow, which runs on Ubuntu by default, is currently passing on this branch.  (The documentation workflow is failing, and so I need to look into that.  I don't think that's a show stopper for merging to main or moving to `biocro/biocro` though.)

That all said, I've mulled over for a while now switching from `testthat` to `tinytest` but wasn't sure it was worth the work involved.

justinmcgrath:
I think the issue on Linux has to do with CRAN being down. I had thought that they didn't have a version built for 3.6, which would have involved installing a lot of the dependencies from source files, and there's more than 30 dependencies. When CRAN is up it will probably work.

gsrohde:
> I think the issue on Linux has to do with CRAN being down. I had thought that they didn't have a version built for 3.6, which would have involved installing a lot of the dependencies from source files, and there's more than 30 dependencies. When CRAN is up it will probably work.

Yes, I saw it was down but didn't put 2 and 2 together!

(BTW, I'm just catching up, noticing new comments on PR #29.)

PR #36, opened 2023-11-01
Changed repository to write documentation to biocro/BioCro-documentation

gsrohde:
Also: use `main` rather than `master` as the branch where pushes generate new documentation.

---

This is a stop-gap fix to get documentation up and running at the new location (under the `biocro` organization).

I'm temporarily using a landing page to ease the transition.

The old ebimodeling BioCro documentation page now points to that landing page.

If you don't like https://biocro.github.io/BioCro-documentation/ as the location, we can easily change it.  I considered just https://biocro.github.io/documentation/, thinking the extra "biocro" was redundant, but I wasn't sure if there would be other, separate documentation for other repositories under `biocro/`.  I used "BioCro-documentation" rather than "biocro-documentation" to distinguish between BioCro, the R package, and biocro, the organization; but maybe this is too fussy and will lead to unnecessary mistypings and 404 errors.  (The hostname portion of a URL is, in general, not case-sensitive, but the rest usually _is_ case sensitive.)

PR #37, opened 2023-11-02
Updated the documentation for the Generate Documentation workflow.

gsrohde:
I updated and clarified this documentation and eliminated references to `ebimodeling` repositories.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -45,7 +45,7 @@ private key, and the public key will be in a file called
     This key pair is needed in order to allow a workflow defined in
 the BioCro repository to push files to the documentation repository.
 
-1. The private key must be added as an _Actions secret_ in the BioCro
+2. The private key must be added as an _Actions secret_ in the BioCro
 repository (*this* repository) under the name `PRIVATE_SSH_KEY`.  When
 doing this, paste the entire contents of the private key file into
 text area labeled **Secret**.  (See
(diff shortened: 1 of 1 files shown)


eloch216:
Looks good. My only critique is that all the items in the numbered list are number 1.

To check my `git-flow` understanding... the target for this PR is `develop` since it isn't critical for the operation of the R package or the workflows, whereas the target for PR #36 is `main` since it _is_ critical for the operation of the workflows?

gsrohde:
> Looks good. My only critique is that all the items in the numbered list are number 1.
> 

This is the way Markdown handles numbered lists.  It spares one having to renumber everything if you add a new item in the middle.  They will be numbered correctly when they appear in the **bookdown** book.

> To check my `git-flow` understanding... the target for this PR is `develop` since it isn't critical for the operation of the R package or the workflows, whereas the target for PR #36 is `main` since it _is_ critical for the operation of the workflows?

Yes.  There will have to be many adjustments to the documentation in connection with changing locations from `ebimodeling/biocro` to `biocro/biocro`, of which this is just one.  I don't want have to bump the version number each time one of these documentation pages changes.  I started with this one because I just had to go through these steps, so that they were fresh in my mind.  But I don't see this as an urgent critical fix.  Most general users, in fact, will never look at this page.

gsrohde:
> All of our branches are the same at this point, so it doesn't particularly matter but I am wondering why this wouldn't be treated as a hotfix and merged with main and develop simultaneously. I think of a hotfix as something that you want done everywhere, immediately. Doesn't this change pertain to both main and develop?

I didn't think this was urgent enough to be a hotfix.  See [my previous comment](https://github.com/biocro/biocro/pull/37#issuecomment-1791147601).

justinmcgrath:
> This is the way Markdown handles numbered lists. It spares one having to renumber everything if you add a new item in the middle. They will be numbered correctly when they appear in the **bookdown** book.

It does render them properly, but one of the points of Markdown is that is that it is formatted nicely in the source in addition to any way it's rendered.

gsrohde:
> It does render them properly, but one of the points of Markdown is that is that it is formatted nicely in the source in addition to any way it's rendered.

OK, I've renumbered the items sequentially.

justinmcgrath (approved):

PR #38, opened 2023-11-03
Workflow fix to v3.0.0

eloch216:
In `git-flow`, a hotfix should be merged into `main` and `develop`. So, I think the branch from PR #36 also needs to be merged into `develop`.

Alternatively, we could merge `main` into `develop`, which seems like it should generally produce the same end result. We should probably follow the stated rules for `git-flow`, but this alternative could be handy in case a hotfix branch is deleted after merging into `main` but before merging into `develop`.

### DIFF
--- a/.github/workflows/automatically-call-document.yml
+++ b/.github/workflows/automatically-call-document.yml
@@ -6,11 +6,8 @@ name: Automatically generate documentation
 
 on:
   push:
-    branches:
-      - main
-
     tags:
-      - '*'
+      - v*.*
 
   # This event creates an artifact for download but does not write to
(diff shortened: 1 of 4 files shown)


justinmcgrath (approved):

gsrohde:
I'm just going to go ahead and merge this: it should have been done as a matter of course when I merged into main.  (I'm still getting used to this workflow.)

Also, since we are (I think) going with the policy of tagging and versioning every commit on the main branch, this branch is really no different from a hotfix, and I would, in retrospect, have called it `hotfix-v3.0.1`.

PR #40, opened 2023-11-03
Update publication location and check out source with submodules.

gsrohde:
This updates the _Document quantities_ workflow to publish to the new location under `biocro/` and to automatically run when a new BioCro version is released.

For now, the quantities documentation isn't versioned.  Any updates will replace what was there before.


justinmcgrath (approved):

gsrohde:
I changed the base (branch to merge into) to `develop`, as it should be.  I think the default branch should now be `develop` to make this kind of mistake less likely.  Most PRs will be requests to merge feature branches into `develop`.

Issue #43, opened 2023-11-06
adding missing references


djaiswal:
We have two (old) missing references: (1) [sugarcane](https://www.nature.com/articles/nclimate3410) and (2) [Miscanthus model coupling with CWRF](https://onlinelibrary.wiley.com/doi/full/10.1111/gcbb.12937).

PR #44, opened 2023-11-07
Add references and make new version

eloch216:
This PR adds a few missing references to the BioCro README, in response to Issue #43 (@djaiswal). Are there any other papers missing from the list? I know of a few others using WIMOVAC, but not sure if they should be included here.

Since this is a hotfix, I will also merge the branch into `develop` once it has been approved here. (Only mentioning this since I think we are still getting used to `git-flow`.)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: BioCro
-Version: 3.0.1
+Version: 3.0.2
 Date: 2023-10-30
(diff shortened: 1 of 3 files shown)


eloch216:
I found two more references.

At some point we might want to make a dedicated page for references. Then we could split them into a few groups, such as one for WIMOVAC references. I don't think we need to do that right now, though.

justinmcgrath (approved):

PR #50, opened 2023-11-11
Changes to check workflow

gsrohde:
This PR makes one substantive change and two (for now) transparent ones:

The substantial change is that now, the checks are run for all specified platforms when the triggering event is a pull request.  Up until now, only R 3.6 on Ubuntu was tested by default to avoid running up against the workflow run limits imposed on private repositories; to test other platforms or R versions required manually dispatching the workflow.

The first transparent change was to replace the deprecated `set-output` command calls with the use of environment files, the recommended equivalent.  See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ for details.  (The `set-output` command was originally scheduled to be disabled on May 31, 2023, but that change has been delayed, and so this workflow has continued to work up until now.  Presumably, it would eventually start to fail without this change in place.)

The second transparent change is simply a simplification of the syntax, replacing usages of the `github.event.inputs` context with the `inputs` context.  Besides shortening the code, it allows inputs declared as boolean to be treated as such instead of being treated as strings.  (See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs.)

### DIFF
--- a/.github/workflows/R-CMD-check-backend.yaml
+++ b/.github/workflows/R-CMD-check-backend.yaml
@@ -94,21 +94,21 @@ jobs:
           args <- paste0('c(\\"', paste(args_value, collapse = '\\", \\"'), '\\")')
 
           # store these in the environment
-          system(sprintf('echo "::set-output name=build_args::%s"', build_args))
-          system(sprintf('echo "::set-output name=args::%s"', args))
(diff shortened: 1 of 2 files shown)


justinmcgrath (approved):

eloch216 (approved):

gsrohde:
I added a small bug fix (and revised a comment) before merging.

PR #52, opened 2023-11-15
Upgraded the versions of actions used by R-CMD-check workflow

gsrohde:
We now use the latest (as of this writing) versions of the `JoshuaTheMiller/conditional-build-matrix` and `actions/checkout` actions.

This was motivated by the deprecation warning about the `set-output` command coming from the `JoshuaTheMiller/conditional-build-matrix` action—the same warning about _our_ code addressed by PR #50.


gsrohde:
Superseded by PR #58.

Issue #51, opened 2023-11-15
Debug R CMD check failures


gsrohde:
Now that we are running automatic R package checks on multiple platforms, not just R 3.6 on Ubuntu, failures on all the other platforms have come to light: see [workflow run #98](https://github.com/biocro/biocro/actions/runs/6871602974).

Some or all of these may be trivial problems.  Nevertheless, they should be investigated and eliminated if possible.

eloch216:
I went through and found the relevant error messages from the different checks. Some of them are pretty baffling. I will start looking into the potential `sprintf` issue.

There are also a few NOTEs that could be addressed, but maybe not at this time.

----
**Windows-latest using R version 3.6**

```
/usr/bin/sh: -c: line 10: syntax error: unexpected end of file

make: *** [C:/R/share/make/winshlib.mk:13: BioCro.dll] Error 2

ERROR: compilation failed for package 'BioCro'
```

----

**MacOS-latest using R version 3.6**

```
 ❯ checking PDF version of manual ... WARNING
  LaTeX errors when creating PDF version.
  This typically indicates Rd problems.
  LaTeX errors found:
```

----


**Ubuntu-20.04 using R version release**

``
 ❯ checking compiled code ... WARNING
  File ‘BioCro/libs/BioCro.so’:
    Found ‘__sprintf_chk’, possibly from ‘sprintf’ (C)
      Objects: ‘framework/ode_solver_library/ode_solver_factory.o’,
        ‘framework/ode_solver_library/boost_ode_solvers.o’
  
  Compiled code should not call entry points which might terminate R nor
  write to stdout/stderr instead of to the console, nor use Fortran I/O
  nor system RNGs nor [v]sprintf.
  
  See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual.
```

----

**Windows-latest using R version release**

```
ERROR
cannot change to directory ''
Error: Error in proc$get_built_file() : Build process failed
```

----

**MacOS-latest using R version release**

```
 ❯ checking whether package ‘BioCro’ can be installed ... [76s/78s] WARNING
  See below...

❯ checking compiled code ... WARNING
  File ‘BioCro/libs/BioCro.so’:
    Found ‘_sprintf’, possibly from ‘sprintf’ (C)
      Objects: ‘framework/ode_solver_library/boost_ode_solvers.o’,
        ‘framework/ode_solver_library/ode_solver_factory.o’
  
  Compiled code should not call entry points which might terminate R nor
[...]

PR #55, opened 2023-11-17
Address R CMD check issues

eloch216:
This is a first attempt at addressing some of the issues identified and discussed in issue #51. Here, I've replaced `sprintf` by `snprintf` in `odeint`, as was done by the `BH` package. Obviously it's not a great solution, but I don't think it's feasible to wait until the `boost` library stops using `sprintf`. We have no idea how long that might take.

As we decide on solutions to the other issues, they can also be implemented on this branch.

Update: I originally made these changes on a hotfix branch because these changes are urgent. But, when `main` is the target, the checks do not run on all operating systems and R versions, because those modifications were introduced as a feature and only exist on `develop`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -37,6 +37,9 @@ be directly added to this file to describe the related changes.
 - This version adds a description of the BioCro git branching model to
   `contribution_guidelines.Rmd` and clarifies the process of updating `NEWS.md`.
 
+- Replaced instances of `sprintf` in boost files with `snprintf` to avoid
+  `R CMD check` warnings
+
 # CHANGES IN BioCro VERSION 3.0.2
 
 ## MINOR CHANGES
--- a/developer_documentation/adding_the_boost_libraries.md
+++ b/developer_documentation/adding_the_boost_libraries.md
@@ -51,10 +51,16 @@ operating systems without it. Hence the need for
 
 4. Update the path to the Boost license in the package `LICENSE` file.
 
-5. Run `R CMD check` and truncate any boost file paths that are flagged as
-   exceeding 100 characters. Be sure to update any associated `#include`
-   directives that reference these files; otherwise, compilation errors will
-   occur.
+5. Run `R CMD check` and address any new warnings or errors related to the boost
+   library. It is likely that the following issues will occur:
+
+   - Some file paths will exceed the 100 character limit. Truncate any such file
+     names, and be sure to update any associated `#include` directives that
+     reference these files; otherwise, compilation errors will occur.
+
+   - As of boost version 1.83, some of the boost libraries included with BioCro
+     call `sprintf`, which is not allowed by CRAN. Replace any such instances of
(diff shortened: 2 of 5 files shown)


justinmcgrath (approved):

eloch216:
I forgot about `lexical_cast` and only removed `sprintf` from `odeint`. We _could_ address the `lexical_cast` issue by updating our boost version. That's a lot of work though, and it still wouldn't fix the `odeint` issue. I think the best route for right now is to just make another "band-aid" fix for `lexical_cast`, as was done in the `BH` package.

justinmcgrath:
Going off of your comment about not know what CRAN will reject, do you know that they would reject a package because of this warning? I've been operating under that assumption. If they're not going to reject based on this, then I'm happy to ignore it entirely. All of these are false positives, and indiscriminately flagging every use of sprintf is stupid enough that hopefully they'll stop doing it entirely and we don't need to fix things that aren't broken. If it's going to be rejected, though, then there's little option.

justinmcgrath:
Tangentially, the overall sense I get from Boost is that they're OK making the change, but ODEINT seems to be orphaned. It hasn't been touched in 4 years. I don't know how they would handle that. The change is simple and if incorporated would be in the next release in April I think. If we get someone's attention, then there's also hope of getting shorter file names and not needing to do any manual edits to Boost.

gsrohde:
> Going off of your comment about not knowing what CRAN will reject, do you know that they would reject a package because of this warning? I've been operating under that assumption. If they're not going to reject based on this, then I'm happy to ignore it entirely. All of these are false positives, and indiscriminately flagging every use of sprintf is stupid enough that hopefully they'll stop doing it entirely and we don't need to fix things that aren't broken. If it's going to be rejected, though, then there's little option.

Here's what the O'Reilley [R Packages](https://r-pkgs.org/release.html#double-r-cmd-checking) book has to say about this:

> When running R CMD check for a CRAN submission, you have to address any problems that show up:
>
> * You must fix all ERRORs and WARNINGs. A package that contains any errors or warnings will not be accepted by CRAN.
>
> * Eliminate as many NOTEs as possible. Each NOTE requires human oversight, which creates friction for both you and CRAN. If there are notes that you do not believe are important, it is almost always easier to fix them (even if the fix is a bit of a hack) than to persuade CRAN that they’re OK. See our [online-only guide to R CMD check](https://r-pkgs.org/R-CMD-check.html) for details on how to fix individual problems.
>
> * If you can’t eliminate a NOTE, list it in cran-comments.md and explain why you think it is spurious. We discuss this file further in [Section 22.6](https://r-pkgs.org/release.html#sec-release-cran-comments).
>
>     Note that there will always be one NOTE when you first submit your package. This reminds CRAN that this is a new submission and that they’ll need to do some extra checks. You can’t eliminate this NOTE, so just mention in cran-comments.md that this is your first submission.

PR #56, opened 2023-11-17
Make not checking the manual the default.

gsrohde:
`--no-manual` is only the default when the check workflow is run automatically.  Checking the manual is the default when the check workflow is run manually, but the user can always uncheck the "Check the manual" checkbox to skip checking of the manual.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -12,7 +12,7 @@ on:
       check-manual:
         description: 'Check the manual'
         type: boolean
-        default: true
+        default: false
       check-vignettes:
         description: 'Check the vignettes'
         type: boolean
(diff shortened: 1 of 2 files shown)


gsrohde:
This make all the configurations except the two Windows ones pass.  In other words, it eliminates the failure on macOS when using R 3.6.  Interestingly, there is no failure on macOS with R release (4.3.2) since the sprintf problem has been eliminated.

justinmcgrath:
> Interestingly, there is no failure on macOS with R release (4.3.2) since the sprintf problem has been eliminated.

sprintf is in ODEINT though. Why would --no-manual affect that?

gsrohde:
> > Interestingly, there is no failure on macOS with R release (4.3.2) since the sprintf problem has been eliminated.
> 
> sprintf is in ODEINT though. Why would --no-manual affect that?

The two macOS configurations tested failed for different reasons: The release version failed because of the sprintf problem.  The R 3.6 version I think wasn't as picky about sprintf but it did have a problem with the manual.  Why the problem with the manual went away when R 4.3.2 was used instead of R 3.6 is unknown to me.

eloch216 (approved):

PR #60, opened 2023-11-28
Update framework

eloch216:
This PR updates the C++ framework to v1.1.1. This will remove a few more instances of `ebimodeling` from `biocro/biocro`.


gsrohde (approved):
I'm still getting used to working with submodules.  Is it worth mentioning that users will have to run `git submodule update` after pulling this change?  Perhaps they "should" just _know_ this, and it might be a pain and inappropriate to put a reminder in the release notes.

(Alternatively, they can run `git pull` with the `--recurse-submodules` flag.  Or they can set the configuration option `submodule.recurse` to `true`.  I don't know how GitHub desktop handles this.)

justinmcgrath:
> Perhaps they "should" just _know_ this, and it might be a pain and inappropriate to put a reminder in the release notes.

I remember almost none of these details about git and would appreciate the notes. I assume those commands need to be run from the biocro repository and not the framework repository. Is a submodule entirely independent from the repository it's contained in? Somehow I assumed that if you make a commit in the containing repository that it would include a reference to the current state of the subrepository. Thus pulling the parent would include making any changes to the submodule. Apparently that's not the case though. Maybe it's possible to set it up that way though?

gsrohde:
> > Perhaps they "should" just _know_ this, and it might be a pain and inappropriate to put a reminder in the release notes.
> 
> I remember almost none of these details about git and would appreciate the notes. I assume those commands need to be run from the biocro repository and not the framework repository.

Yes, I'm assuming you're not inside the `src/framework` subtree.

> 
> Is a submodule entirely independent from the repository it's contained in? 

Not entirely.  (It doesn't know anything about the containing repository, but the containing repository knows about it.)

> Somehow I assumed that if you make a commit in the containing repository that it would include a reference to the current state of the subrepository. 

The containing repository can specify what version of the subrepository should be used.  But you have to take special pains to change this version: just doing a commit of any old change won't automatically make the containing repository point to the currently-checked-out version of the submodule.  And when a user pulls your commit, even if you have correctly updated the submodule version pointed to, their update of the containing repository doesn't automatically update their version of the subrepository unless they have configured things that way.

If a user updates their containing repository and the submodule version it expects changes, then, if they run `git status` in the containing repository, they will get a note something like `modified:   src/framework (new commits)` if the subrepository is out of sync.  This is one sense in which the submodule is not entirely independent of the repository it's contained in.

> Thus pulling the parent would include making any changes to the submodule. Apparently that's not the case though. Maybe it's possible to set it up that way though?

This is what I meant by my note about setting the configuration option `submodule.recurse` to `true`.  It will "set it up that way."  I'm not sure why this isn't the default, but there may be a sensible reason.  Whether this will only automate syncing to the correct submodule version when doing a `git pull`, or whether it also works when switching between branches is something I haven't tested out.  (Perhaps there's a separate setting for the latter.)

An intermediate option would be to set the `status.submodulesummary` option to "true" with
```
git config status.submodulesummary 1
```
Then, if you run `git status`, you get a slightly more informative message, something containing something like
```
	modified:   src/framework (new commits)

Submodules changed but not updated:

* src/framework a562c43...50cee78 (1):
  < Merge pull request #5 from biocro/hotfix-v1.1.1
```

If you have the time, you might find it worthwhile reading [section 7.11 of the Pro Git book](https://git-scm.com/book/en/v2/Git-Tools-Submodules).  Or perhaps the section on submodules in Chapter 14 of Brent Laster's book _Professional Git_ (available on-line from the UIUC library).

justinmcgrath:
"Completely independent" was not good word choice. I guess more what I mean is does it treat the submodule like other tracked files. Updating he submodule would be seen as a change, that change is then committed in the parent repository. Then pulling the parent would include setting the submodule to the correct version. Also switching between branches on the parent would set the submodule to the correct version. If I understand, settings `submodule.recurse` to true will do part of that, but what are the "special pains" so that the parent knows which submodule version to use?

gsrohde:
> what are the "special pains" so that the parent knows which submodule version to use?

I probably made this sound like a bigger deal than it is.  If the state of your submodule has changed, either because you went into it an committed new work or because you pulled down new work from the remote repository, it suffices to "git add" the submodule directory before committing, just as you would do for a changed file.  (You can also use the `git commit -a` option, but I tend to avoid this because I tend to accumulate changes in my tracked files that I'm not ready to commit yet.)

gsrohde:
> This is what I meant by my note about setting the configuration option `submodule.recurse` to `true`. It will "set it up that way." I'm not sure why this isn't the default, but there may be a sensible reason. Whether this will only automate syncing to the correct submodule version when doing a `git pull`, or whether it also works when switching between branches is something I haven't tested out.

I checked this, and the `--recurse-submodules` option (or the `submodule.recurse` configuration option) works with both `git pull` and `git checkout <branch>` (or `git switch <branch>`).

eloch216:
> I'm still getting used to working with submodules. Is it worth mentioning that users will have to run `git submodule update` after pulling this change? 
> 
> Alternatively, they can run `git pull` with the `--recurse-submodules` flag. Or they can set the configuration option `submodule.recurse` to `true`.

> I remember almost none of these details about git and would appreciate the notes. I assume those commands need to be run from the biocro repository and not the framework repository.

> I checked this, and the `--recurse-submodules` option (or the `submodule.recurse` configuration option) works with both `git pull` and `git checkout <branch>` (or `git switch <branch>`).


I added some basic instructions to the release notes. Let me know if you think they are okay. GitHub desktop takes care of this stuff for me, so I'm not sure about the best way to do it manually with command line git. I did find this post, which seems to prefer using `--recurse-submodules` with `git pull`: https://superuser.com/questions/1332024/need-for-git-submodule-update-over-git-pull-recurse-submodules.

gsrohde:
> I added some basic instructions to the release notes. Let me know if you think they are okay. GitHub desktop takes care of this stuff for me, so I'm not sure about the best way to do it manually with command line git.

The instructions look okay.

Probably setting the config option `submodule.recurse` to `true` is the most convenient method of keeping submodules up to date if you don't want to have to type `--recurse-submodules` each time you run `update`, `switch`, or `pull`.  If you want to maintain a little more control, set the `status.submodulesummary` option to "true" instead, and then run `git submodule update` if and when the result of a `git status` check tells you you need to.  (I tend to run `git status` frequently, and so this is viable option for me.)

Note that the post you refer to often writes `--recurse` in place of `--recurse-submodules`.  As best I can tell, there _is_ no `--recurse` option, but Git lets you shorten option specifications up to the shortest prefix that has a unique completion.  If at some future date Git adds some other option called `--recurse-something-else`, then writing simply `--recurse` will presumably no longer work.

Regarding the section [7.11 of the Pro Git book](https://git-scm.com/book/en/v2/Git-Tools-Submodules): If you don't want to read the whole thing, it might still be worth reading the short bit entitled "Issues with Submodules" near the end, particularly the first subsection, called "Switching branches" and maybe also the next one, "Switching from subdirectories to submodules".

justinmcgrath:
> Probably setting the config option `submodule.recurse` to `true` is the most convenient method of keeping submodules up to date if you don't want to have to type `--recurse-submodules` each time you run `update`, `switch`, or `pull`. If you want to maintain a little more control, set the `status.submodulesummary` option to "true" instead, and then run `git submodule update` if and when the result of a `git status` check tells you you need to. (I tend to run `git status` frequently, and so this is viable option for me.)

I added those options to my config file and it seems like that makes everything work the way one would expect. I was looking for a place in the guidelines to add this, but it appears we don't have a section for getting starting with development.

justinmcgrath (approved):
Everything seems to work for me.

gsrohde:
> I was looking for a place in the guidelines to add this, but it appears we don't have a section for getting starting with development.

While it might be appropriate to add a few hints about getting started with development, I don't think we want the guidelines to turn into a tutorial about using Git, or, for that matter, coding best practices.  Rather, I think the main focus should be on conventions we have adopted _for this project_.

eloch216:
> > I was looking for a place in the guidelines to add this, but it appears we don't have a section for getting starting with development.
> 
> While it might be appropriate to add a few hints about getting started with development, I don't think we want the guidelines to turn into a tutorial about using Git, or, for that matter, coding best practices. Rather, I think the main focus should be on conventions we have adopted _for this project_.

That's fair. I think a restriction to our conventions is too limiting, though. We also want to make sure that users and developers can accomplish critical basic tasks like accessing the code. So a few short instructions about updating git submodules are probably reasonable. If it's something that's not obvious to one of the main BioCro developers, it's probably something worth adding.

justinmcgrath:
> > > I was looking for a place in the guidelines to add this, but it appears we don't have a section for getting starting with development.
> > 
> > 
> > While it might be appropriate to add a few hints about getting started with development, I don't think we want the guidelines to turn into a tutorial about using Git, or, for that matter, coding best practices. Rather, I think the main focus should be on conventions we have adopted _for this project_.
> 
> That's fair. I think a restriction to our conventions is too limiting, though. We also want to make sure that users and developers can accomplish critical basic tasks like accessing the code. So a few short instructions about updating git submodules are probably reasonable. If it's something that's not obvious to one of the main BioCro developers, it's probably something worth adding.

I was thinking of adding this somewhere: "Our repository uses git submodules. For historical reasons, by default git does not automatically update submodules when pulling or checking out parent branches, but this can be set up by running `git config summary.submodulesummary true` and `git config submodule.recurse true`. If these options are not set, you should use `git pull --recurse-submodules`, when pulling and `git submodule update` after switching branches".

gsrohde:
> I was thinking of adding this somewhere ...

I'd go even more succinct (surmising that most users are not interested in Git history):

"This repository uses Git submodules.  To ensure the submodules are populated and kept up to date, use the `--recurse-submodules` option when first cloning the repository, when pulling new changes with `git pull`, and when switching branches with `git checkout` or `git switch`.  (If you happen to forget the `--recurse-submodules` option when pulling changes or switching branches, run `git submodule update` to bring the submodules up to date.)

"You may configure Git to apply the `--recurse-submodules` flag automatically to these commands (excepting `git clone`) by setting the configuration option `submodule.recurse` to `true` with `git config submodule.recurse true`."

I guess that wasn't more succinct.  But I think it's a little clearer.

By the way, for what it's worth, I did happen upon a discussion of reasons why one might _not_ want to set this option: https://stackoverflow.com/questions/53514521/what-are-the-drawbacks-to-setting-gits-submodule-recurse-config-option-to-true.  But I don't think we need to discuss this in our documentation.  Moreover, I don't think these reasons really apply in our case.

PR #59, opened 2023-11-22
Update docs for new biocro location

gsrohde:
This removes most references to the ebimodeling GitHub organization and updates references to `ebimodeling/biocro` to point to `biocro/biocro` instead.  Most of these occur in various places in the documentation.  Most links to the online documentation have been replaced with links to https://biocro.github.io, with (in some cases) instructions on how to navigate to the particular section of the documentation of interest.  This decreases dependence on the precise layout of the online documentation.  Some other changes and clarifications to the documentation have been made as well.

PR #61, opened 2023-11-29
Removed obsolete example.

gsrohde:
This PR is in response to a [comment on PR 59](https://github.com/biocro/biocro/pull/59#pullrequestreview-1753237726) that I neglected to deal with before merging that PR.

I refer to the example as "obsolete" simply because, now that BioCro has a submodule, the installation command `devtools::install_github("biocro/biocro")` used in the example will fail.

Another example could perhaps be substituted, but I didn't think it necessary.  At most, perhaps a reference to some exposition of the topic "separating control flow from data" could be included (though there are certainly many other common sources of code duplication).

(As an aside, a workaround for `devtools::install_github`'s lack of support for submodules is to define as short helper function `install_submodule_git` as outlined in [a comment in PR #751 for the r-lib/devtools repository](https://github.com/r-lib/devtools/pull/751#issuecomment-215540807):
```
install_submodule_git <- function(x, ...) {
   install_dir <- tempfile()
   system(paste("git clone --recursive", shQuote(x), shQuote(install_dir)))
   devtools::install(install_dir, ...)
}
```
Then we may install BioCro using
```
install_submodule_git("https://github.com/biocro/biocro")
```

But this seems too complicated and involved to use in the example.)

### DIFF
--- a/developer_documentation/contribution_guidelines.Rmd
+++ b/developer_documentation/contribution_guidelines.Rmd
@@ -391,59 +391,7 @@ discuss the following questions:
   prefixes, except that kg is the coherent unit of mass, not g.
 
 * Do not copy and paste code, changing only small parts. Choose a
-  design that eliminates the duplication. Duplication is often the
-  result of not separating control flow from data. Consider the
-  following R code.
-
-    ```r
-    if (!("lattice" %in% installed.packages()[,"Package"])) {
-      install.packages("lattice", repos="http://R-Forge.R-project.org", type="source")
-    }
-
-    if (!("BioCro" %in% installed.packages()[,"Package"])) {
-      devtools::install_github("biocro/biocro")
-    }
-
-    if (!("boot" %in% installed.packages()[,"Package"])) {
-      devtools::install_url("http://cran.r-project.org/src/contrib/Archive/boot/boot_1.3-7.tar.gz")
-    }
-    ```
-
-    The data and behavior can be separated.
-
-    ```r
-    required_packages = list(
-        # package name  # installation function   # function arguments
-        list("lattice", install.packages,         list("lattice", repos="http://R-Forge.R-project.org", type="source")),
-        list("BioCro",  devtools::install_github, list("biocro/biocro")),
-        list("boot",    devtools::install_url,    list("http://cran.r-project.org/src/contrib/Archive/boot/boot_1.3-7.tar.gz"))
-    )
-
-    install_packages_if_missing = function(package_table) {
-        installed_packages = installed.packages()[,"Package"]
-        for (row in package_table) {
-            if (!(row[[1]] %in% installed_packages)) {
-                do.call(row[[2]], row[[3]])
-            }
-        }
-    }
-
-    install_packages_if_missing(required_packages)
-    ```
-
-    In the first example, the behavior is spread throughout all of the
-    code, and there is not an obvious indication of what is being
-    done. Once one understands what is being done, one must still read
-    all of the code to be sure some different behavior is not hidden
-    somewhere.
-
-    In the second example, it is clear that the same task is performed
-    repeatedly because control flow is in a single place, and the
-    place has a meaningful name: `install_packages_if_missing`. For a
-    long list, this is succinct, and easier to understand and
-    maintain. If one wanted, one could devise a way to use the
-    meaningful names of the columns of the `required_packages` table
-    within the function.
+  design that eliminates the duplication.
 
 * Make an effort to write unit tests.  (See the vignette [_An
   Introduction to BioCro for Those Who Want to Add Models_][unit tests


justinmcgrath:
I took Ed's comment to mean that BioCro can't be installed that way now, not that the example of duplication was obsolete. I don't know that the example needs to remain, but copying and pasting chunks of code in sequence is the most common type of duplication that I see, and one of the easiest types to address.

eloch216:
I had originally thought we could just replace the BioCro package in the example with another one that would work. If we remove the example, I think we should at least replace it with a link to another source discussing duplication and loops. I don't think the short version will be helpful to people with less coding experience. This one seems pretty good, but there might be others: https://r4ds.had.co.nz/iteration.html

gsrohde:
> I had originally thought we could just replace the BioCro package in the example with another one that would work. 

I had thought of simply replacing `BioCro` in the example with some other R module on GitHub—one whose repository doesn't use submodules.  I think, though, I simply thought, "Of all the coding best-practices we could discuss and even provide examples for, why single out this one?", and so I didn't bother.

Ideally, we could just expect our contributors to be experienced programmers who know and follow best coding practices so that our contribution guidelines could be limited to discussing the somewhat arbitrary stylistic conventions we have decided we wish to adhere to, as well as other conventions unique to our project, such as the use and documentation of units.

If you feel that more is warranted here than just an admonition to keep code DRY (and I sort of feel that even that should be unnecessary), I invite you to edit this PR and add what you think is needed.

gsrohde:
I'm not going to do more on this PR.  Right now, all it does is fix a link and remove an example.

If you really want to provide a new example, be my guest: the simplest would be to approved this PR and make a new PR to add a new example.  If you really want to add an example as part of this PR, then (unless you have something ready to go) convert this to a draft and assign it to yourself, and then, when ready, mark it as ready to be reviewed.  In any case, the link should get fixed one way or the other.

eloch216:
> If you really want to provide a new example, be my guest

I just added one here. It's shorter and simpler than the previous one.

eloch216 (approved):

developer_documentation/contribution_guidelines.Rmd
```diff
@@ -391,7 +391,40 @@ discuss the following questions:
...
+  In the first example, the behavior is spread throughout all of the
+  code, and there is not an obvious indication of what is being
+  done. Once one understands what is being done, one must still read
+  all of the code to be sure some different behavior is not hidden
+  somewhere.
+
+  In the second example, it is clear that the same task is performed
+  repeatedly because control flow is in a single place, and the
+  place has a meaningful name: `get_col_max`. For a long list, this
+  is succinct, and easier to understand and maintain.
```
gsrohde, line 427: To my mind, the original code makes it very clear that the same task is being performed repeatedly because we call the function that performs that task, `max`, repeatedly, in three successive lines that have a parallel structure.  True, it _is_ repetitive.  But it isn't unclear.

`get_column_max` or even `get_column_maximum` might be more meaningful names than `get_col_max`, but even these names are a little misleading: the function doesn't take a column, but rather, the name of a column; moreover (in particular), the name of a column of whatever the name `biocro_result` is attached to at the moment when `get_column_max` is called.  (If we want to be really perverse, we can do something like

```r
col_for_max <- c('Leaf', 'Stem', 'Root')

get_col_max <- function(colname) {
    max(biocro_result[[colname]])
}

max_vals <- sapply(col_for_max, get_col_max)

biocro_result <- mtcars

col_for_max <- c('mpg', 'wt', 'hp')

max_vals <- sapply(col_for_max, get_col_max)
```
Probably we shouldn't be encouraging the writing of functions that use global variables.)

Which code is better is highly dependent upon the context.  If we are going to use `max_leaf`, `max_stem`, and `max_root` later on in the code and we don't anticipate needing maximums for any of the other columns, then the original code is probably the best; even though it is a little repetitive, it is probably shorter than any of the alternatives and it saves us from having to do, say,
```r
max_leaf <- max_vals[['Leaf']]
max_stem <- max_vals[['Stem']]
max_root <- max_vals[['Root']]
```
after we are done calling `sapply`.

On the other hand, if all we want to do is display a labeled list of maximums for various columns, the most succinct way of doing that is simply to call
```r
sapply(biocro_result[c('Leaf', 'Stem', 'Root')], max)
```

If we really think we are going to want to repeatedly get maximums for various sets of columns, then maybe it would be worthwhile to define a function like `get_col_max`, but preferably, a version that doesn't use a global variable.  In fact, if we expect to want to do this with various result data frames, it would be worthwhile defining a closure function.

justinmcgrath: I agree that in this case there isn't anything unclear about the code, although it is repetitive. I would have used `max_values = sapply(biocro_results[c('Leaf', 'Stem', 'Root')], max)` or a for loop.

Maybe we can change the example to that? I don't think we need to spend a lot of time on this.

justinmcgrath (approved):

PR #62, opened 2023-11-29
Address wformat warnings

eloch216:
The `R CMD check` workflow started showing some warnings on ubuntu-20.04 using R version devel related to "format not a string literal and no format arguments [-Wformat-security]". I looked into it, and this is a pretty simple security fix: https://stackoverflow.com/questions/9707569/c-array-warning-format-not-a-string-literal

Update: both of the ubuntu checks now pass!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -81,7 +81,7 @@ be directly added to this file to describe the related changes.
   precise layout of the online documentation.  Some other changes and
   clarifications to the documentation have been made as well.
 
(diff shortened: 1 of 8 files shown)


justinmcgrath:
This is another case of flagging things that are not a problem, but it gets rid of the warnings. The rule should be something like "a format specifier should be used printing the value of a string value *with unknown content*", but it seems like the people adding these warnings to the compilers aren't thinking too critically about it.

justinmcgrath (approved):

PR #63, opened 2023-11-29
Address wdeprecated warnings

eloch216:
This PR attempts to address some `wdeprecated` warnings that occur when running the `R CMD check` workflow on windows-latest using R version release. The solution follows https://github.com/boostorg/functional/commit/6a573e4b8333ee63ee62ce95558c3667348db233


eloch216:
Looks like there is still a problem:
```
  ../inc/boost/container_hash/hash.hpp:130:33: warning: 'template<class _Arg, class _Result> struct std::unary_function' is deprecated [-Wdeprecated-declarations]
```

I'm not quite sure yet how to fix this one.

NEWS.md
```diff
@@ -81,6 +81,11 @@ be directly added to this file to describe the related changes.
   precise layout of the online documentation.  Some other changes and
   clarifications to the documentation have been made as well.
 
+- Removed usage of deprecated C++ standard library functions
+  `std::unary_function` and `std::binary_function` to address `R CMD check`
+  warnings; this is a temporary manual fix should not be required after
```
gsrohde, line 86: Either "this is a temporary manual fix that should not be required after ..." or "this temporary manual fix should not be required after ... ."

gsrohde:
> It looks like boost typically releases new versions in April, August, and December, so 1.84 might come out soon.

Yes, from the conversations I had last year about BH, I was led to believe that Boost updates three times a year.

I guess I will hold off on further review until this latest issue is solved.

By the way, given the maintenance we are having to do on the Boost code we are using, I think the `inc` subtree is a prime candidate for being made into a submodule.

gsrohde:
Not sure why [call-R-CMD-check / Check R package on macOS-latest using R version 3.6](https://github.com/biocro/biocro/actions/runs/7036062796/job/19147875878#logs) is suddenly failing in run #135 of this branch.  Trying a manual re-run.

eloch216:
> By the way, given the maintenance we are having to do on the Boost code we are using, I think the `inc` subtree is a prime candidate for being made into a submodule.

That seems like it could be a good idea. Otherwise we will have to separately update the copy of boost included in the `skelBML` repository.

I can't figure out how to solve the `std::unary_function` issue with `container_hash`, so I am thinking we should ignore this `R CMD check` error until boost 1.84 is released. Then we can make a `biocro/boost` repository for the new version, and include it in `biocro/biocro` and `biocro/skelBML` as a submodule. I'm sure some new boost-related compilation warnings and errors will occur, and we can address them at that time.

I will go ahead and merge this branch into `develop` for now, since it does remove some of the compiler warnings.

gsrohde:
> I can't figure out how to solve the `std::unary_function` issue with `container_hash`, so I am thinking we should ignore this `R CMD check` error until boost 1.84 is released.

@eloch216 This doesn't seem any harder than your changes to `functional.hpp`, and the solution is the same: simply eliminate the #if/#else and always use the #if code.  See PR #64.  Is there something I'm missing?

inc/boost/functional.hpp
```diff
@@ -21,7 +21,6 @@ namespace boost
     namespace functional
     {
         namespace detail {
-#if defined(_HAS_AUTO_PTR_ETC) && !_HAS_AUTO_PTR_ETC
```
gsrohde, line 24: The commenting out of this line and of the _else_ code below are the only significant changes to this file.  The other 28 changes are simply whitespace changes.

I don't know if you are adopting the philosophy "as long as I'm changing this file, I may as well eliminate extraneous whitespace", or if your editor simply does this automatically.  Mainly, I just want to make sure you are aware of this.

Fortunately, both my editor's ediff utility and GitHub's "Diff view" settings contain options to hide whitespace changes, so this isn't too annoying.

justinmcgrath (approved):
With Scott's PR against this branch, I don't get any warnings with R CMD check --as-cran.

eloch216:
> > I can't figure out how to solve the `std::unary_function` issue with `container_hash`, so I am thinking we should ignore this `R CMD check` error until boost 1.84 is released.
> 
> @eloch216 This doesn't seem any harder than your changes to `functional.hpp`, and the solution is the same: simply eliminate the #if/#else and always use the #if code. See PR #64. Is there something I'm missing?

I just didn't think of doing that.

inc/boost/functional.hpp
eloch216: My editor does this automatically, to enforce BioCro style guidelines about trailing whitespace.

PR #64, opened 2023-12-10
Commented out use of std::unary_function as base class for hash_base.

gsrohde:
With these changes, the checks pass on the [call-R-CMD-check / Check R package on windows-latest using R version release](https://github.com/biocro/biocro/actions/runs/7160194022/job/19494297283#logs) platform.


justinmcgrath:
With this, I only get warnings about vignettes because I don't build them, so it seems to fix all the CRAN warnings.

eloch216 (approved):

PR #65, opened 2023-12-12
Fix and simplify c3photoC

eloch216:
I installed R v3.6.0 and the appropriate Rtools (v3.5) in an attempt to debug the `R CMD check` failures on that version of R. The package compiled on my machine, but there were some test failures related to `src/module_library/c3photoC`.

I tried looking into the failures using the following code:

```r
library(BioCro)

cases <- cases_from_csv('BioCro:c3_assimilation', 'biocro/tests/module_test_cases/')

evaluate_module('BioCro:c3_assimilation', cases[[1]]$inputs)

cases[[1]]$expected_outputs
```

When running the test case, the module outputs indicated that 0 loop iterations had occurred, and that `Rp` was `NA`. Ultimately, this was due to two issues with the code.
- The convergence check was made using `abs`, not `fabs`. `abs` is for integers, so it must have been reporting 0 whenever the difference in assimilation rates was between -1 and 1. For these inputs, that happened to be the case on the first guess.
- The photorespiration rate was calculated after the loop converged from `Ci`, not `Ci_pa`. `Ci` actually didn't get updated until the start of each loop, so its value was from the previous loop iteration. For the test case inputs, since the loop "converged" right away, `Ci` was still set to 0, so we were dividing by zero.

The loop was tricky to diagnose because there were two different `Ci` values (with different units) being used. So, I simplified it.

I don't think this will address the `R CMD check` failures, but it's a simple fix. I wonder why the loop seemed to behave properly in newer versions of R / other compilers; I wonder if `abs` behaves differently when compiling the code with different compilers?


justinmcgrath:
It looks right to me. Do you know what causes the small differences - the use of fabs, using the updated Ci, or both?

I thought the R CMD check errors were only on the testing platform. I recall everything working when I did the check on my computer.

justinmcgrath (approved):

eloch216:
> It looks right to me. Do you know what causes the small differences - the use of fabs, using the updated Ci, or both?

I think the test errors were caused by using `abs` instead of `fabs`. The stored module test cases and soybean simulation results were originally calculated with code that included the `Ci` error, so that shouldn't have caused any test failures.

> I thought the R CMD check errors were only on the testing platform. I recall everything working when I did the check on my computer.

Did you actually use R v3.6.0 and Rtools v3.5? I only get check errors with this setup -- not with the newer one that I normally use.

eloch216:
There is a test failure on Ubuntu here but it is just in a harmonic oscillator test. This test uses random oscillator parameters and sometimes fails, but it's not repeatable. I'm not worried about this.

justinmcgrath:
I used R v3.6.3, which is what's listed on the test machine logs. I used RTools 3.5.0.4, I don't see specifics on the test logs - just 3.5. Are you seeing errors in our tests or things that R checks?

eloch216:
> I used R v3.6.3, which is what's listed on the test machine logs. I used RTools 3.5.0.4. I don't see specifics on the test logs - just 3.5. Are you seeing errors in our tests or things that R checks?

I saw test errors, but no compilation errors, and no other R CMD check errors.

I didn't notice that the test machine uses R v3.6.3; I had assumed it uses v3.6.0 based on the action name, so I used that. (Our `DESCRIPTION` file also states 3.6.0 as the minimum required version.) I am using the same version of Rtools as you.

eloch216:
I just checked using R v3.6.3 on my machine, and `rcmdcheck::rcmdcheck` didn't report any errors. (R v3.6.0 also doesn't report any errors.) It seems that the test failure from the workflow is impossible to reproduce...

gsrohde:
> * The convergence check was made using `abs`, not `fabs`. `abs` is for integers, so it must have been reporting 0 whenever the difference in assimilation rates was between -1 and 1. For these inputs, that happened to be the case on the first guess.

In C++, `abs` is overloaded in the cmath library to accept and return floating-point types.  The only reason to use `fabs` is if you are writing C code (C doesn't allow overloading of functions).  At least, that _should_ be the case.  But perhaps the Windows version of R 3.6 uses a non-standards-compliant compiler.  If so, perhaps we should be requiring a later R version on Windows rather than using functions geared toward C code.

> so it must have been reporting 0 whenever the difference in assimilation rates was between -1 and 1

Did you actually check what gets reported by `abs()`, either using `cout` or a debugger?

justinmcgrath:
> In C++, `abs` is overloaded in the cmath library to accept and return floating-point types. The only reason to use `fabs` is if you are writing C code (C doesn't allow overloading of functions). At least, that _should_ be the case. But perhaps the Windows version of R 3.6 uses a non-standards-compliant compiler.

That version of Rtools uses gcc 4.9.3 from mingw. I've had a problem with noncompliance in mingw before. It does seem somewhat unlikely though.

Does this mean that the differences in the results are entirely from using last instead of second-to-last value of Ci?

eloch216:
> Does this mean that the differences in the results are entirely from using last instead of second-to-last value of Ci?

On all operating systems and versions of R, there was a change in the output of `c3photoC`, which was related to using the second-to-last value when calculating the photorespiration rate `Rp`. Fixing the issue necessitated a change in the stored results, but not to any of the important crop model outputs such as the canopy assimilation rate or any biomass values.

On Windows with R version 3.6.0, there was another issue in `c3photoC`, where the check for convergence was not being implemented properly. It seems that on this OS/R version, an unqualified `abs` is interpreted as the integer version of `abs`, so the loop falsely indicated convergence whenever the difference in assimilation rates was smaller than 1.

PR #67, opened 2023-12-13
Make R 4.0.0 instead of R 3.6.0 the minimum supported R version

gsrohde:
Make R 4.0.0 instead of R 3.6.0 the minimum supported R version, and update the test workflow accordingly.

The package check passes on all platform + R version configurations except Windows with the Release version of R (see https://github.com/biocro/biocro/actions/runs/7198925592/job/19609456910).  This is because of a failing harmonic oscillator test failure that @eloch216 alluded to earlier.

UPDATE: All checks passed in [Run #163](https://github.com/biocro/biocro/actions/runs/7199430426).  This really is an intermittent bug.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -52,14 +52,14 @@ on:
         description: 'Platform to check'
         type: choice
         options:
-          - R 3.6 on Ubuntu
+          - R 4.0.0 on Ubuntu
           - R release version on Ubuntu
           - R devel version on Ubuntu
-          - R 3.6 on macOS
+          - R 4.0.0 on macOS
           - R release version on macOS
-          - R 3.6 on Windows
+          - R 4.0.0 on Windows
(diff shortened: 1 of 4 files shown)


justinmcgrath:
Looks good. It would be nice to figure out the oscillator thing, but maybe for another day.

justinmcgrath (approved):

gsrohde:
Converted to a draft while we consider merging PR #70 (which fixes a problem with the test of R 3.6.0) instead.

eloch216:
I guess we don't need to do this since PR #70 was merged?

gsrohde:
> I guess we don't need to do this since PR #70 was merged?

That is correct.  My thought was to leave this here as a draft for if and when we decide to require R 4.0.0.  I also just now added the "tabled for now" label.

eloch216:
The minimum version is now 4.1.0, so I don't think we need this PR anymore.

PR #68, opened 2023-12-13
Eliminate the use of C99 designated initializers.

gsrohde:
See [my comment in issue #51](https://github.com/biocro/biocro/issues/51#issuecomment-1813828278) concerning this non-standard (in C++11) syntax.


eloch216 (approved):
Looks good to me! I also locally confirmed that all tests pass on Windows using R v3.6.0.

PR #70, opened 2023-12-20
stop echoing OBJECTS variable when running workflow on windows

gsrohde:
This fixes the problem with the R-CMD-check workflow running on Windows with R 3.6.0.  It also explicitly sets the R version for R 3.6 checks on _all_ platforms to 3.6.0 to match the minimum required R version specified in the DESCRIPTION file.

This is an alternative to PR #67.

### DIFF
--- a/.github/workflows/R-CMD-check-backend.yaml
+++ b/.github/workflows/R-CMD-check-backend.yaml
@@ -226,6 +226,17 @@ jobs:
 
         shell: Rscript {0}
 
+      - name: Modify winshlib.mk
+        if: ${{ matrix.os == 'windows-latest' && matrix.r == '3.6.0' }}
+        run: |
+          # Remove echoing of the OBJECTS variable, which seems to
+          # cause problems when using R 3.6 on Windows
+          sed '/echo.*\$(OBJECTS) /s/ \$(OBJECTS) / /' \
+              "C:/R/share/make/winshlib.mk" > "C:/R/share/make/winshlib.mk-tmp"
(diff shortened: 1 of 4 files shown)


gsrohde:
The test failure on Ubuntu using the R release version is a little worrisome: for one run of the harmonic oscillator test in Attempt 1 (https://github.com/biocro/biocro/actions/runs/7269612636/job/19807505491), the value of total energy was NA for five rows sampled.  I did rerun all the failed jobs, and the check succeeded for all platforms in Attempt 2 (https://github.com/biocro/biocro/actions/runs/7269612636).

This is likely a problem in the with the design of the test: random values are chosen for the various parameters, and perhaps certain choices result in degenerate systems.  In any case, it is very doubtful this test failure has anything to do with the changes in this PR.

The failure of the harmonic oscillator test on Windows with the R release version is something we've seen intermittently.  It is puzzling, though, that the test failure on Attempt 1 here issues the message "actual != expected but don't know how to show the difference", whereas in other cases, we see a much more specific message.  For example the failure in run #161 (https://github.com/biocro/biocro/actions/runs/7198925592/job/19609456910) shows
```
── Failure ('test.HarmonicOscillationModeling.R:164:13'): Harmonic oscillator position and velocity values match the expected values (random parameters and initial values) ──
  result$velocity[index] (`actual`) not equal to result$expected_velocity[index] (`expected`).
  
    `actual`: -0.0252
  `expected`: -0.0254
```

justinmcgrath:
I double checked this on my computers and everything seems to work, other than also getting the spurious error about the harmonic oscillator.

justinmcgrath (approved):

PR #71, opened 2024-01-02
Replace all fabs with std::abs

eloch216:
This PR replaces `fabs` and unqualified `abs` with `std::abs`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,11 +39,13 @@ be directly added to this file to describe the related changes.
   when using `git pull`, `git checkout`, or `git switch` to update a local copy
   of the BioCro repository, or to move to or from this branch.
 
-- Two bugs were corrected in `src/module_library/c3photoC.cpp`:
-  - Convergence is now tested using `fabs` rather than `abs`, since we need the
-    absolute value of a floating point number, not an integer.
-  - The photorespiration value `Rp` is now calculated using the value of `Ci`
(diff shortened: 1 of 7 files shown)


NEWS.md
```diff
@@ -39,11 +39,13 @@ be directly added to this file to describe the related changes.
...
-    absolute value of a floating point number, not an integer.
-  - The photorespiration value `Rp` is now calculated using the value of `Ci`
-    from the current loop iteration (rather than the previous loop iteration).
+- All instances of `fabs` or unqualified `abs` have been replaced by `std::abs`.
+  The use of unqualified `abs` in `src/module_library/c3photoC.cpp` had been
+  causing test failures when running BioCro on Windows using R version 3.6.0.
+
+- Another bug was corrected in `src/module_library/c3photoC.cpp`: The
+  photorespiration value `Rp` is now calculated using the value of `Ci` from the
+  current loop iteration (rather than the previous loop iteration).
```
gsrohde, line 48: Just a nit, which you can ignore if you like: I think this sentence reads better without the parentheses: "The photorespiration value `Rp` is now calculated using the value of `Ci` from the current loop iteration rather than the previous loop iteration."

gsrohde (approved):
See comment.

NEWS.md
eloch216: I think I'll ignore this for now but we can always fine-tune the wording in `NEWS` when making a release branch.

PR #72, opened 2024-01-02
Change some testing behavior

eloch216:
If an external module library based on `skelBML` uses a different version of the BioCro C++ framework than the version used by the `BioCro` R package, a warning will occur whenever a module is evaluated. The warning will say something like "The `edBML` module library R package uses a newer version of the BioCro C++ framework than the `BioCro` R package (1.2.1 vs. 1.1.1)." Such warnings should not be considered failures when testing modules as part of a package, since they are not related to a module's functionality. 

However, these warnings have been triggering test failures. (Thanks to @uphone87 for pointing this out.) The reason is that the `tryCatch` call in `BioCro::test_module` has been catching conditions (which includes errors _and_ warnings). Here I changed this behavior so `test_module` now only catches errors.

I also changed one of the crop model testing functions to use `expect_no_error` rather than `expect_silent` for the same reason.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -109,6 +109,12 @@ be directly added to this file to describe the related changes.
   `Rf_error` and `Rprintf` without a format specifier; a format specifier of
   `"%s"` should always be used when printing the value of a string variable.
 
+- Modified some testing-related functions so that warnings due to mismatched
+  framework versions do not trigger test failures: the `tryCatch` call in
+  `test_module` now only catches errors (not warnings) when evaluating the
+  module, and `test_plant_model` (in `crop_model_testing_helper_functions.R`)
+  now uses `expect_no_error` instead of `expect_silent`.
+
 # CHANGES IN BioCro VERSION 3.0.2
 
 ## MINOR CHANGES
--- a/R/module_testing.R
+++ b/R/module_testing.R
@@ -7,7 +7,7 @@ test_module <- function(module_name, case_to_test)
     msg <- character()
     actual_outputs <- tryCatch(
         evaluate_module(module_name, case_to_test[['inputs']]),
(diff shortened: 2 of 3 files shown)


gsrohde (commented):
I suppose this is okay as far as it goes.  But something seems less than ideal when a framework version mismatch triggers tons of test warnings; it seems like one should suffice.

I tried this out using a test module library that only has a single test, the automatically-generated "All modules are functioning correctly" in `test.Modules.R`.  If there is a framework version mismatch, then in addition to the _1_ in the "PASS" column, I also get _12_ in the "WARN" column.

This brings up a related issue: I had always assumed that if the tests remained constant, then the sum FAIL + WARN + SKIP + PASS would remain constant as well, even if R function code or C++ code changes resulted in changes to the numbers in the respective columns.  But that assumption doesn't hold up here.  Whether that is due to our violating some tacit assumptions on the way one should write test code or whether it is due to my own misapprehensions about the way `testthat` works is something I don't know at this point.

Perhaps it would be better if the testing framework simply ignored this particular warning.  Instead, we could add another automatically-generated test that would test if the framework version matches that of the version of BioCro that has been loaded and issue a warning—or error—if the two don't match.

justinmcgrath (approved):

eloch216:
> Perhaps it would be better if the testing framework simply ignored this particular warning. Instead, we could add another automatically-generated test that would test if the framework version matches that of the version of BioCro that has been loaded and issue a warning—or error—if the two don't match.

This sounds like it would probably require a bit of time to figure out. While it might not be ideal to have large numbers of warnings in the test results, these warnings do not actually pose a problem, since the tests still count as successes. I don't think this should be a priority right now. You can make an issue about it if you'd like, though.

PR #75, opened 2024-01-08
Par energy content

uphone87:
This PR makes the following changes,

- Change the par_energy_content constant to 0.219
- Update the related test file
- Update the NEWS

The current value of 0.235 is from an exercise example in the Campbell & Norman textbook, which may only represent an extremely clear sky. Here the more realistic 0.219 is the inverse of 4.57. 4.57 is a commonly used constant to convert PAR in W/m2 to &mu;mol m-2 s-1 as you would find in many references (e.g., [in Table 2 here](https://www.controlledenvironments.org/wp-content/uploads/sites/6/2017/06/Ch01.pdf)). No significant changes were found on assimilation and yield due to this change. About 1% of change was seen in the cumulative transpiration for soybean in 2002. This makes sense because the par_energy_content is used in the soil evaporation and evapotranspiration modules. The two modules need the radiation to be in W/m2, and thus we need the par_energy_content to make this conversion.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,8 @@ merged in since the previous release.
 Subsequent commits will then include a new "Unreleased" section in preparation
 for the next release.
 -->
+# Unreleased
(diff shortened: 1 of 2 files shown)


eloch216:
@uphone87 It looks like this branch was based off `main`, but it should have been based off `develop`. Because of this, it's missing some changes to `c3photoC` that exist on `develop` but not `main`. That will make it difficult to merge this as-is. Since the changes here are relatively straightforward, it may be easiest to make a new branch based off `develop`, re-implement the changes, and then make a new PR.

One other comment: this constant should probably be changed in all of the crop model definitions, not just soybean.

uphone87:
A new branch was created based off develop and a new PR was created. This is closed without merges.

PR #76, opened 2024-01-10
Par energy content2

uphone87:
This PR makes the following changes,

- Change the par_energy_content constant to 0.219 for Soybean, Miscanthus and Willow

- Update the related test files for Soybean, Miscanthus and Willow

- Update in the NEWS

The current value of 0.235 is from an exercise example in the Campbell & Norman textbook, which may only represent an extremely clear sky. Here the more realistic 0.219 is the inverse of 4.57. 4.57 is a commonly used constant to convert PAR in W/m2 to μmol m-2 s-1 as you would find in many references (e.g., [in Table 2 here](https://www.controlledenvironments.org/wp-content/uploads/sites/6/2017/06/Ch01.pdf)). No significant changes were found on assimilation and yield due to this change. About 1% of change was seen in the cumulative transpiration for soybean in 2002. This makes sense because the par_energy_content is used in the soil evaporation and evapotranspiration modules. The two modules need the radiation to be in W/m2, and thus we need the par_energy_content to make this conversion.


justinmcgrath:
Thanks. I'm not clear on whether this is being used consistently everywhere. The note in NEWS.md says that if the conversion is being done outside BioCro to be sure it's the same. Is this regarding weather data that report irradiance instead of PPFD?

eloch216 (approved):
Looks good! Sorry there were a few snags along the way.

> I'm not clear on whether this is being used consistently everywhere. The note in NEWS.md says that if the conversion is being done outside BioCro to be sure it's the same. Is this regarding weather data that report irradiance instead of PPFD?

To be clear, all calculations within a single BioCro crop model are done with the same value of `par_energy_content`. This value is no longer hard-coded in any BioCro modules.

gsrohde:
@uphone87 I have at least a few comments to make on this before it is merged.

data/miscanthus_x_giganteus.R
```diff
@@ -118,7 +118,7 @@ miscanthus_x_giganteus <- list(
         nRdb1                       = 0.1247,
         nvmaxb0                     = -16.25,
         nvmaxb1                     = 0.6938,
-        par_energy_content          = 0.235,
+        par_energy_content          = 0.219,  # W * s / micromole. Also J / micromole. Conversion from photon flux density to to irradiance for PAR. Equals 1/4.57. Plant Growth Chamber Handbook. CHAPTER 1 – RADIATION– John C. Sager and J. Craig McFarlane. Table 2, Pg 3 (https://www.controlledenvironments.org/wp-content/uploads/sites/6/2017/06/Ch01.pdf)
```
gsrohde, line 121: This line is excessively long, over five times the recommended line length (generally 80 characters, though up to 100 or possibly more is acceptable in specific circumstances).  It could easily be broken up as
```
        par_energy_content          = 0.219,  # W * s / micromole. Also J / micromole. Conversion
                                              # from photon flux density to to irradiance for
                                              # PAR. Equals 1/4.57. Plant Growth Chamber
                                              # Handbook. CHAPTER 1 – RADIATION– John C. Sager and
                                              # J. Craig McFarlane. Table 2, Pg 3
                                              # (https://www.controlledenvironments.org/wp-content/uploads/sites/6/2017/06/Ch01.pdf)
```
or possibly better,
```
        nvmaxb1                     = 0.6938,
        par_energy_content          = 0.219,  # W * s / micromole.
            # Conversion from photon flux density to to irradiance for
            # PAR. Equals 1/4.57. Plant Growth Chamber Handbook. CHAPTER 1 –
            # RADIATION– John C. Sager and J. Craig McFarlane. Table 2, Pg 3
            # (https://www.controlledenvironments.org/wp-content/uploads/sites/6/2017/06/Ch01.pdf)
        par_energy_fraction         = 0.5,
```

But I question whether this documentation really belongs here at all.  With the exception of `soybean.R`, none of the quantities in crop model data files is documented except for this quantity and `atmospheric_transmittance`.  It seems to me a more appropriate place to document this would be in the `crop_model_definitions.Rd` file in the `man` directory.  Alternatively, if it is deemed desirable to keep the documentation close to the data and not have the documentation appear for general users, and if these data files are going to continue to be rather sparsely documented, it might be more appropriate to simply put a leading comment at the top of the data file explaining a few quantities of interest with a note on the quantity line itself saying "see comments at top of file".

Also note that the exact same comment appears here and in two other crop model data files.  Just as with code, documentation should follow the DRY ("don't repeat yourself") principle, though perhaps not as strictly as for code.  The problem here, of course, is that not just the documentation, but the quantity itself (which _is_ part of the code) is repeated three times.  Ideally, the quantities that remain constant from crop to crop would be factored out into a separate data file and each crop model would use these common values and supplement them with the quantities that are crop-specific.  In other words, the data would be modularized, analogously to the way we modularize computational components of models in the BioCro module library.  This would be relatively easy to do using the `utils::modifyList` function.  (I am not, of course, suggesting doing this as part of this PR.)

Note that (for what it's worth) the unit for `par_energy_content` is uniformly listed in the module library files as "J / micromol" (see https://biocro.github.io/BioCro-documentation/quantity_docs/quantities.html).  The code that scrapes the library files to deduce the unit for each quantity and display it in the _Quantities_ documentation file does not also look at the data files, so there is no enforcement of consistency with regard to what is listed for units in the data files.  (But again, most all units in the data files are currently undocumented there, and I'm not recommending changing this.)

eloch216: >  Ideally, the quantities that remain constant from crop to crop would be factored out into a separate data file and each crop model would use these common values and supplement them with the quantities that are crop-specific.

We have tried doing something like this with the soil parameters, since they depend on location and not on crop. They have been moved to a separate data structure (`soil_parameters`) that also has its own documentation.

Ultimately we weren't able to remove repetition in the data. We wanted to keep things simple for users, and didn't want to make them do something like `c(soybean$parameters, soil_parameters$clay_loam)` to run a basic model. And R didn't seem to let us append the soil parameters in the soybean model definition file. So we ended up simply repeating soil parameter values in the soybean model definition file. More info can be found here: https://github.com/ebimodeling/biocro-dev/issues/424.

Another similar situation was the inclusion of the time zone. Ultimately we decided to include it as part of the weather data. It doesn't change so it could be a parameter, but conceptually it is related to a particular set of weather data. We could also consider doing something similar with `par_energy_content`.

I think a "general parameter" data object with documentation would be a good idea. It doesn't fix all issues, but it helps with documentation. It doesn't need to be done for this PR, though, since it would likely include a lot of other parameters besides `par_energy_content`.

NEWS.md
```diff
@@ -117,6 +117,15 @@ be directly added to this file to describe the related changes.
   module, and `test_plant_model` (in `crop_model_testing_helper_functions.R`)
   now uses `expect_no_error` instead of `expect_silent`.
 
+- Change the energy_par_content constant to 0.219. This is a conversion rate 
+  from photons to W/m2, which equals 1/4.57. 4.57 is a commonly used constant 
+  to convert PAR in W/m2 to micromol m-2 s-1. Users should be cautious when 
+  performing this conversion outside of BioCro to make sure the two are the same,
+  for example, pre-processing the radiation data. See more details in
```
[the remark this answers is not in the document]
justinmcgrath: I had to ask some follow up questions to be sure I understood it. Overall, I suspect there's a clearer way to handle this in general, but I don't feel like it's a priority.

NEWS.md
```diff
@@ -117,6 +117,15 @@ be directly added to this file to describe the related changes.
   module, and `test_plant_model` (in `crop_model_testing_helper_functions.R`)
   now uses `expect_no_error` instead of `expect_silent`.
 
+- Changed the energy_par_content constant to 0.219. This is a conversion rate 
+  from photons to W/m2, which equals 1/4.57. 4.57 is a commonly used constant 
+  to convert PAR in W/m2 to micromol m-2 s-1. Users should be cautious when 
+  performing this conversion outside of BioCro to make sure the two are the same,
+  for example, pre-processing the radiation data. See more details in
```
gsrohde, line 124: I'm going to try rewording the middle two sentences, partly to make sure I'm really understanding this correctly, and (if I am) partly to make it clearer to readers.

"This is a conversion rate from photons (in micromoles) to watt-seconds.  It equals 1/4.57, 4.57 being a commonly used constant to convert PAR in W/m2 to micromol m-2 s-1.  Users should take care to ensure that if preprocessing of radiation data is required to prepare it for use with BioCro, the same conversion factor is used."

Probably this information should go somewhere else in addition to (or instead of?) being hidden away in the news file.  But that can be a separate issue.

eloch216: If we are going to modify this, I would also like to explain the parameter's meaning more clearly. To me a conversion factor is something that only changes the units, e.g. gallons of water to liters of water. The `par_energy_content` factor is different, because it relates the energy of light to the number of photons. In the water analogy, it would be more like the density of water, which could be used if we wanted to specify a weight of water rather than a volume. We're still specifying the amount of water, but using a totally different type of measurement.

"This factor represents the average energy per photon in photosynthetically active radiation (PAR), expressed in units of Joules (or, equivalently, Watt-seconds) per micromole of photons. This value of this ratio depends on the spectrum of radiation in the PAR range. The value used here equals 1/4.57, 4.57 being a commonly used constant to convert PAR in W/m2 to micromol m-2 s-1. Users should take care that this factor is used consistently if radiation data is preprocessed for use with BioCro."

justinmcgrath: Here's a further change to that using more exact terminology. I find that a phrase like "converting photons to W m^-2" doesn't make sense.

"This is a conversion rate from photon flux density (in micromoles per square meter per second) to energy flux (joules per square meter per second, also watts per square meter) for photosynthetically active radiation (PAR). It equals 1/4.57, 4.57 being a commonly used constant to convert PAR in W m^-2 to micromole m^-2 s^-1. Users should take care to ensure that if processing of radiation data is required to prepare it for use with BioCro, the same conversion factor is used."

data/miscanthus_x_giganteus.R
gsrohde: > And R didn't seem to let us append the soil parameters in the soybean model definition file. So we ended up simply repeating soil parameter values in the soybean model definition file.

Unless there is something I'm not understanding, this is relatively easy to do: simply source the soil parameters file in `soybean.R` and then set parameters to
```
parameters = utils::modifyList(soil_parameters$clay_loam,
      list(
            [other parameters]
      )
)
```
See the branch I made called `proof-of-concept_using_soil_parameters_in_crop_models`.

One thing I'm not sure of is whether sourcing the soil parameters file somehow circumvents lazy loading of data.  But this is a relatively small data set, so it doesn't seem like a huge concern.

NEWS.md
justinmcgrath: I like Ed's, but "joules" and "watt" aren't capitalized and "^" for exponentiation.

data/miscanthus_x_giganteus.R
eloch216: From https://github.com/ebimodeling/biocro-dev/pull/422:

> I eliminated two undocumented code objects (`SoilType` and `showSoilType`), replacing them with a new data set called `soil_parameters` that contains the same information. This also allowed me to remove the `soil_type_selector` module and its related C++ code. I attempted to keep including the soil parameters as part of the crop-specific parameter lists, but this required me to call `source` within the crop-specific data files; this didn't cause a problem when installing the package, but it did cause a problem when using `devtools::run_examples`. So, I changed my mind and just kept the soil parameters separate from the crop parameters.

NEWS.md
uphone87: Thanks for all the suggestions. I find Justin's version clearer. It's Watt (=Joule/second), but not watt-seconds.

gsrohde: > It's Watt (=Joule/second), but not watt-seconds.

Well, for what it's worth, it seems to me you could either say

1 μmol (photons) = 0.219 J = 0.219 watt-seconds

or

1 μmol/m^2/s = 0.219 J/m^2/s = 0.219 W/m^2

In any case, the new wording you committed seems entirely suitable (unless review fatigue has caused me to overlook something), and I will submit a new review giving explicit approval in case that's what I need to do to remove the block on merging.

gsrohde (approved):

data/miscanthus_x_giganteus.R
gsrohde: > > I attempted to keep including the soil parameters as part of the crop-specific parameter lists, but this required me to call `source` within the crop-specific data files; this didn't cause a problem when installing the package, but it did cause a problem when using `devtools::run_examples`. So, I changed my mind and just kept the soil parameters separate from the crop parameters.
> 
> I use `devtools::run_examples` to run examples that are designated as "don't run" or "don't test" since they don't get checked by `R CMD check`. Otherwise I would need to find and run them manually. So it's important to be able to use `devtools::run_examples`.

What sort of problem did you have running the examples?  I tried this out using an installation from my `proof-of-concept_using_soil_parameters_in_crop_models` branch, calling `run_examples('.', run_dontrun = TRUE)` and didn't notice any problems.  Perhaps I don't know what I'm looking for.  What version of `devtools` are you using?

uphone87:
There is a failed check for the windows platform. I'm not sure why it happened just based on the error message.

gsrohde:
> There is a failed check for the windows platform. I'm not sure why it happened just based on the error message.

The error message in this check is baffling.  But I did try manually rerunning the check of this platform on the `develop` branch (now containing your merges), and I instead get a known error, one failed test having to do with the harmonic oscillator system.  So I'm not too worried about this.

data/miscanthus_x_giganteus.R
eloch216: I can't find a record of the error I was getting. I tried using `devtools:run_examples` on your proof of concept branch and didn't encounter any issues. Maybe it was a bug in an older version of `devtools`. It's good to know that including `source` in a data file doesn't seem to be an issue anymore!

PR #82, opened 2024-02-08
Document default solver

eloch216:
There is an R CMD check message about using `:::` in our own code, which is related to the `default_ode_solver`. Here, I converted that list into a function that provides default settings for each ODE solver type. Now there is no need to use `:::` anymore. There was also a message about the use of `\item` in the documentation for one function, so I addressed that as well.


justinmcgrath:
I find that having only "default" in the name hides what a call does, and tangentially something is a default because it's given as the default argument, not because of the name.

Rather than grouping the solvers, they could be separate lists, and then set the argument accordingly: `ode_solver=homemade_euler`. Maybe it's helpful to group them into a single object rather than have a separate object for each. I would require an argument to the function then though, so it's obvious what's done: ` default_ode_solver('homemade_euler')`. The function is reimplementing a list, so it could be done as a list.

```r
default_ode_solvers <- list(
        homemade_euler = list(
                    type = 'homemade_euler',
                    output_step_size = NA,
                    adaptive_rel_error_tol = NA,
                    adaptive_abs_error_tol = NA,
                    adaptive_max_steps = NA
                ),

        boost_euler = list(
                    type = 'boost_euler',
                    output_step_size = 1,
                    adaptive_rel_error_tol = NA,
                    adaptive_abs_error_tol = NA,
                    adaptive_max_steps = NA
                ),

        boost_rk4 = list(
                    type = 'boost_rk4',
                    output_step_size = 1,
                    adaptive_rel_error_tol = NA,
                    adaptive_abs_error_tol = NA,
                    adaptive_max_steps = NA
                ),

        boost_rkck54 = list(
                    type = 'boost_rkck54',
                    output_step_size = 1.0,
                    adaptive_rel_error_tol = 1e-4,
                    adaptive_abs_error_tol = 1e-4,
                    adaptive_max_steps = 200
                ),

        boost_rosenbrock = list(
                    type = 'boost_rosenbrock',
                    output_step_size = 1.0,
                    adaptive_rel_error_tol = 1e-4,
                    adaptive_abs_error_tole = 1e-4,
                    adaptive_max_steps = 200
                )
)
```

eloch216:
My first attempt was to just move the original `default_ode_solver` list to the `data` directory. I got some errors, and thought that maybe I couldn't use a package data object as the default value of a function's input argument. I may have made a mistake somewhere, though. I can try again with your list.

eloch216:
I tried implementing the default ODE solver settings as a list in the `data` directory called `default_ode_solvers`. Then the default value of the `ode_solver` input argument of `run_biocro` becomes `default_ode_solvers$homemade_euler`. In this case I get an `R CMD check` warning:

```
run_biocro: no visible binding for global variable
  'default_ode_solvers'
```

It seems that functions in the package namespace don't know about data sets within the package. I was able to address the issue by using the fully-qualified name `BioCro::default_ode_solvers$homemade_euler`.

justinmcgrath:
Eh, I don't know how that works. This looks good. The function is fine too it you want it to be a bit shorter in the argument. Mostly I was concerned about not being able to tell what the default was without searching elsewhere.

PR #83, opened 2024-02-08
Use boost submodule

eloch216:
This PR replaces the `inc/boost` directory with a git submodule pointing to the new `biocro/boost` repository. This will make it easier to keep the copies of the boost libraries included in `biocro/biocro` and `biocro/skelBML` in sync with each other.


eloch216:
Got some weird errors related to graphs when running `R CMD check` on macOS-latest using R version 3.6.0. This doesn't seem to be related to `boost` so I'm re-running those tests.

gsrohde:
> Got some weird errors related to vignette graphs when running `R CMD check` on macOS-latest using R version 3.6.0. This doesn't seem to be related to `boost` so I'm re-running those tests.

Were you seeing these errors on your own machine?  I don't see any errors in the GitHub Actions logs.

I can't imagine why using a submodule for the Boost library would have any bearing on what check errors one would get.

gsrohde:
[My browser gets all wonky when I try to add a review: the "Add your review" button opens the "Files changed" tab with its 3762 changes, and then I can barely type anything.  So I'm just adding a comment here instead.]

In principle, this seems like a good idea.  But I can't really sign off on the version numbering system you are using for the submodule (see issue https://github.com/biocro/boost/issues/1), and I have suggestions for slightly different structuring of the submodule inclusion (see issue https://github.com/biocro/boost/issues/2).

eloch216:
> > Got some weird errors related to vignette graphs when running `R CMD check` on macOS-latest using R version 3.6.0. This doesn't seem to be related to `boost` so I'm re-running those tests.
> 
> Were you seeing these errors on your own machine? I don't see any errors in the GitHub Actions logs.
> 
> I can't imagine why using a submodule for the Boost library would have any bearing on what check errors one would get.

No, they were in the GitHub action checks. The errors are gone because I reran the checks and they passed the second time.

eloch216:
> [My browser gets all wonky when I try to add a review: the "Add your review" button opens the "Files changed" tab with its 3762 changes, and then I can barely type anything. So I'm just adding a comment here instead.]
> 
> In principle, this seems like a good idea. But I can't really sign off on the version numbering system you are using for the submodule (see issue [biocro/boost#1](https://github.com/biocro/boost/issues/1)), and I have suggestions for slightly different structuring of the submodule inclusion (see issue [biocro/boost#2](https://github.com/biocro/boost/issues/2)).

I will make a `biocro/inc` repository rather than `biocro/boost`. This will preserve the file structure we currently have, and it will be more flexible in the future in case we need to include other libraries besides boost.

PR #84, opened 2024-02-09
Use boost submodule 2

eloch216:
This is an alternative to PR https://github.com/biocro/biocro/pull/83. It uses a different directory structure for the boost libraries, as implemented in the `biocro/inc` repository.

If this PR is approved, I will delete the `biocro/boost` repository.


justinmcgrath:
This find that this organization is nicer, but what happens if we want to include another library?

eloch216:
> I find that this organization is nicer, but what happens if we want to include another library?

Couldn't we just add another directory in the `inc` repository? E.g. if we wanted to add `sundials`, we would just have a new `sundials` directory within `inc`. We wouldn't even need to change anything in the `makevars` file since it already includes `inc`.

justinmcgrath:
Ah, I see. I was not thinking of this the right way.

justinmcgrath (approved):

PR #85, opened 2024-02-10
Use C++14 to compile code on systems that support it.

gsrohde:
This PR is meant to replace PR #81.

This PR keeps the requirement that a compiler supporting C++11 is available; but it makes changes so that C++14 is used during compilation on most platforms that support it.

This is done by adding the variable setting `CXX_STD = CXX14` in the `Makevars` and `Makevars.ucrt` files.  `Makevars.urct` is recognized and prefered over `Makevars.win` on Windows platforms when versions of R >= 4.2.0 are being used.  It is assumed that versions of R < 4.2.0 are also less likely to support the C++14 standard, and so the setting `CXX_STD = CXX11` is used in the `Makevars.win` file.

Since `Makevars.ucrt` and `Makevars.win` would be identical except for the `CXX_STD` setting, we have factored out the common code into a third file, `Makevars.win.mk`; this file is then included by the two Windows _Makevars_ files using Make's _include_ directive.

One wrinkle in using the _include_ directive in a Makevars file is this: Although normally one would expect a file with an include directive to act equivalently to a file that simple substituted the contents of the included file for the directive, this doesn't seem to be entirely the case in the context of an R package.  In particular, when the variable `OBJECTS` is set directly inside a Makevars file, the setting overrides the default setting provided by R, but when `OBJECTS` is set in an included file, **the value of `OBJECTS` set by R is not overridden unless the _override_ directive is used!**


gsrohde:
I'm still getting a note about specifying a standard (this time, C++14): "Specified C++14: please drop specification unless essential."

I'm converting this to a draft while I test out some tweaks.

gsrohde:
Superseded by PR #86.

PR #86, opened 2024-02-13
Remove the C++11 requirement from DESCRIPTION

gsrohde:
[This PR replaces PRs #81 and #85.]

This PR does three things:
 
* It removes the C++11 SystemRequirement from the DESCRIPTION file since this now prompts a NOTE during R CMD check on R-devel.  (Moreover, it is desirable to allow compilation with later versions of C++ on systems that support it.)
* CXX_STD = CXX11 is now explicitly set in `Makevars.win`, the version of the Makevars file that Windows systems using older versions of R use.  Windows systems using more recent versions of R will use the newly-added `Makevars.ucrt` file, which does not set CXX_STD.  (To avoid tons of repeated code, these two files include a third file, `Makevars.win.mk`, which includes the Make code that they have in common.)
* Since newer versions of R will use C++17 now that C++11 is being removed from the SystemRequirements field, and since C++17 is a lot pickier about the use of some deprecated declarations in the Boost code, this PR specifies the use of a new version of the framework submodule that uses pragma directives to turn off the resulting warnings.

@eloch216 I explicitly set the framework submodule to the `get_rid_of_deprecations_warnings` branch, but clearly this should change once this branch has been merged into `main`.  I'm unclear about the best way to handle a PR that depends upon changes in a pending PR for a submodule.


justinmcgrath:
This is most of what's needed for #103. Can we change this to use C++11 everywhere? CRAN doesn't want to ignore deprecation (a silly policy that we can't change), so until Boost is C++17 compliant we can use 11 everywhere.

At some point it's likely that we'll want to use newer C++ features. If there's anything in mind, let's make an issue pointing out what we'd like, and we can use that as a place to document what needs to be done to get there.

gsrohde:
> This is most of what's needed for #103. Can we change this to use C++11 everywhere? CRAN doesn't want to ignore deprecation (a silly policy, since you can specify the standard, but we can't change this), so until Boost is C++17 compliant we can use 11 everywhere.

I'm making a PR (much simpler than this one) to move the C++11 standard specification from the DESCRIPTION file to the Makevars files.  I'm not sure this will prevent the NOTE during R CMD check on R-devel, however.  If we discontinued support for R 3.6.0 on Windows, we could probably remove the C++11 specification altogether, though this might bring back C++17-related warnings.  Specifying C++14 might be an option if the other options aren't viable.

justinmcgrath:
At this point the only goal is to remove the NOTE and since we know that C++11 works, it will be easiest to stick with that. Having any notes kicks the package out of the automated system, requiring intervention from me and someone on CRAN. It's not the biggest thing on its own, but I suspect that a typical interaction with CRAN will involve some back and forth, and this step at least doubles the time needed for each submission. Separately, we can working on moving to a new standard, which I'm sure will take a while to sort out.

gsrohde:
> At this point the only goal is to remove the NOTE and since we know that C++11 works, it will be easiest to stick with that. Having any notes kicks the package out of the automated system, requiring intervention from me and someone on CRAN. It's not the biggest thing on its own, but I suspect that a typical interaction with CRAN will involve some back and forth, and this step at least doubles the time needed for each submission. Separately, we can working on moving to a new standard, which I'm sure will take a while to sort out.

Removing the NOTE by removing the C++11 specification from DESCRIPTION and moving it to the Makevars files doesn't seem possible.  See my note in PR #105.

gsrohde:
Current plan is to keep the C++11 requirement in DESCRIPTION (even though it will require our justifying the resulting NOTE) with a longer-term goal of moving to C++17-compliant code.

PR #90, opened 2024-03-01
Update documentation examples

eloch216:
One of the requirements listed in the "R Packages" advice page (https://r-pkgs.org/release.html) is to make sure all exported functions have documentation that includes `value` and `examples`. I found a few functions that didn't meet these requirements, and also standardized some formatting (just indentation).

Functions related to creating and testing module case files were a bit tricky. I looked at the documentation for `write.csv` to see how its examples work. They use `tempdir`, so I tried that out and it seems to work well (pending the results from the GitHub action checks).

----

We might need to still do something about the examples for `system_derivatives`, which are currently wrapped in `dontrun` because they require the `deSolve` library and are also very slow. I'm a bit unsure about this. Apparently the CRAN policy is that:

> Examples/code lines in examples should never be commented out.
> Ideally find toy examples that can be regularly executed and checked.
> Lengthy examples (\> 5 sec), can be wrapped in `\donttest{}`.
> If you don't want your code to be executed but still visible to the user, use `\dontrun{}`.

It seems that `donttest` may be more appropriate for these examples, but then I think we would need to add `deSolve` to the `Suggests` for our package. That might be reasonable, since the documentation for `system_derivatives` mentions `deSolve` several times. Alternatively, we could remove these examples altogether, or use a really simple Euler method to solve them (which wouldn't require any other packages). Or we can just leave them as-is and see if they are a problem for CRAN. What do you think?

### DIFF
--- a/man/module_case_files.Rd
+++ b/man/module_case_files.Rd
@@ -142,7 +142,7 @@ initialize_csv(
   overwrite = TRUE
 )
 
-file.show(file.path(td, 'BioCro_thermal_time_linear.csv'))
+writeLines(readLines(file.path(td, 'BioCro_thermal_time_linear.csv')))
 
 add_csv_row(
   'BioCro:thermal_time_linear',
@@ -151,7 +151,7 @@ add_csv_row(
   'temp above tbase'
 )
 
-file.show(file.path(td, 'BioCro_thermal_time_linear.csv'))
+writeLines(readLines(file.path(td, 'BioCro_thermal_time_linear.csv')))
 
 update_csv_cases('BioCro:thermal_time_linear', td)
 }
--- a/man/module_case_files.Rd
+++ b/man/module_case_files.Rd
@@ -127,28 +127,31 @@
 
 \examples{
 # First, we will initialize a test case file for the 'BioCro' library's
-# 'thermal_time_linear' module, which will be saved in the current directory as
+# 'thermal_time_linear' module, which will be saved in a temporary directory as
 # 'BioCro_thermal_time_linear.csv'. Then, we will add a new case to the file.
 # Finally, we will update the file. Note that the call to `update_csv_cases`
 # will not actually modify the file unless it is manually edited beforehand to
 # change an input or output value.
 
-\dontrun{
+td <- tempdir()
 
 initialize_csv(
   'BioCro:thermal_time_linear',
-  '.',
+  td,
   nonstandard_inputs = list(temp = -1),
   overwrite = TRUE
 )
 
+file.show(file.path(td, 'BioCro_thermal_time_linear.csv'))
+
 add_csv_row(
   'BioCro:thermal_time_linear',
-  '.',
+  td,
   list(time = 101, sowing_time = 100, tbase = 20, temp = 44),
   'temp above tbase'
 )
 
-update_csv_cases('BioCro:thermal_time_linear', '.')
-}
+file.show(file.path(td, 'BioCro_thermal_time_linear.csv'))
+
+update_csv_cases('BioCro:thermal_time_linear', td)
 }
--- a/man/system_derivatives.Rd
+++ b/man/system_derivatives.Rd
@@ -83,9 +83,7 @@ soybean_system <- system_derivatives(
 
 derivs <- soybean_system(0, unlist(soybean$initial_values), NULL)
 
-# Example 2: a simple oscillator with only one module (requires deSolve)
-
-\dontrun{
+# Example 2: a simple oscillator with only one module
 
 times = seq(0, 5, length=100)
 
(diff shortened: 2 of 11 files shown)


justinmcgrath:
Thanks. This small stuff isn't fun to get to.

I'm fine with any of the ways to handle "dontrun". Marking them "donttest" and suggesting deSolve seems both easiest and most useful for users, but go with whatever you think is best.

eloch216:
Agreed. I am not having fun working on this 😂

I will try switching to `donttest` and adding `deSolve` to `DESCRIPTION` as necessary.

man/module_case_files.Rd
```diff
@@ -127,28 +127,31 @@
...
 
 initialize_csv(
   'BioCro:thermal_time_linear',
-  '.',
+  td,
   nonstandard_inputs = list(temp = -1),
   overwrite = TRUE
 )
 
+file.show(file.path(td, 'BioCro_thermal_time_linear.csv'))
```
gsrohde, line 145: I'm thinking it would be better to show the file contents inline, since it's short.  I think
```
writeLines(readLines(file.path(td, 'BioCro_thermal_time_linear.csv')))
```
would work.  The resulting presentation is clearer, I think.

On my system, at least, I was baffled that clicking "Run examples" showed me output like
```
[1] "Added new case to file `/var/folders/l1/_9rb4hdn4fzdg783664d6yt80000gq/T//Rtmpcs7aM0/BioCro_thermal_time_linear.csv`"
```
but didn't seem to show me the contents of that file.  It was only later, when I was dinking around on my system, that I noticed a small window had appeared, with the file contents shown in a tiny font.  This was when I was working in the R Gui app.  RStudio's behavior was baffling as well.  Instead of opening the file to display it, it prompted me to allow a file called "content" to be saved to my Downloads file.

gsrohde (changes requested):
Excepting the use of the `file.show` function, all looks OK to me.

man/module_case_files.Rd
eloch216: The `writeLines/readLines` method looks good. I just used `file.show` because I was following the examples for `write.csv`. I had no idea it would have such strange behavior. On my computer, it just made a window pop up with the contents of the file, which didn't seem too bad at the time.

gsrohde: > On my computer, it just made a window pop up with the contents of the file, which didn't seem too bad at the time.

Just out of curiousity, is this a Windows computer, and are you running R through a GUI app or from the command line?

eloch216: I'm using Windows and RGui. The font size used in the popup window was the same as the font size used in the terminal window.

man/system_derivatives.Rd
```diff
@@ -83,9 +83,7 @@ soybean_system <- system_derivatives(
 
 derivs <- soybean_system(0, unlist(soybean$initial_values), NULL)
 
-# Example 2: a simple oscillator with only one module (requires deSolve)
```
gsrohde, line 86: Maybe you should have just changed this to "requires patience".  I'm being a little facetious, but still, although you warn the user that Example 3 will run very slowly, it's probably worth putting a very prominent warning right at the top of the examples section that will appear next to the "Run examples" link.  Otherwise users may begin to wonder, as I did, whether the "Run examples" link really does anything.

Also note that removing "dontrun" will cause _pkgdown_ to take considerably longer to build the reference section of the pkgdown site.  I don't see this as a real problem—the pkgdown documentation really need only be built once for each new release, and it's probably a _good_ thing that people who go to the public site to view the documentation don't have to wait around to see the result of running the example.  But it's something to be aware of.

man/system_derivatives.Rd
```diff
@@ -112,13 +110,11 @@ lattice::xyplot(
   auto=TRUE,
   data=result
 )
-}
 
-# Example 3: solving 500 hours of a soybean simulation. This requires the
-# deSolve package and will run very slow compared to a regular call to
-# `run_biocro`.
+# Example 3: solving 500 hours of a soybean simulation. This will run very slow
```
gsrohde, line 114: slowly

man/module_case_files.Rd
gsrohde: > The font size used in the popup window was the same as the font size used in the terminal window.

Probably more disconcerting than the small font size (which could be made bigger after all) was the fact that the window appeared in the background where I didn't notice it.  In any case, it just makes sense to me to display the file contents "inline", just as the graphs, which take up far more screen real-estate than these small files, are presented inline.

man/system_derivatives.Rd
eloch216: The warning is a good idea. I think the extra time for building the pkgdown documentation is acceptable.

eloch216:
I would like to merge this but can't. It says: "Merging is blocked. Merging can be performed automatically once the requested changes are addressed." I tried marking the requested changes as "resolved," but I still can't do the merge. @gsrohde Perhaps you have to mark them as resolved since you requested them.

gsrohde (approved):
I think the formal way to resolve requested changes it to click "re-request review" next to my name.  But I think this should be mergeable now.

Issue #92, opened 2024-03-06
Problems generating documentation for quantities


eloch216:
There was an error in the recent automated workflow run related to the release of version 3.1.0:

https://github.com/biocro/biocro/actions/runs/8176197071/job/22355046093

It looks like another case where we have to provide a username and email address for git. I think this should probably addressed via a hotfix branch (?)

Update: maybe I misinterpreted the error message. I'm not totally sure what went wrong. For posterity, we have the following output from the "Commit updates and push to target (documentation) repository" step:

```
cd target
  
  # user.name and user.email are somewhat arbitrary but must be provided:
  git config user.name biocro-action
  git config user.email bogus@url.com
  git add .
  git commit -m "updated quantities documentation"
  git push
  shell: /usr/bin/bash -e {0}
  env:
    publish-to: biocro/BioCro-documentation
    R_LIBS_USER: /home/runner/work/_temp/Library
    TZ: UTC
    _R_CHECK_SYSTEM_CLOCK_: FALSE
    NOT_CRAN: true
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
Error: Process completed with exit code 1.
```

gsrohde:
> Update: maybe I misinterpreted the error message. I'm not totally sure what went wrong.

Judging from these two lines
```
nothing to commit, working tree clean
Error: Process completed with exit code 1.
```
I think the problem is that we are trying to commit changes when nothing has changed, because the script produced exactly the same result as it did when it was run on the source for the previous release.  This will indeed produce an exit code of 1 from the commit command.

I suppose I could test this conjecture by running the script on the previous release version of BioCro and on this one and check that the results are exactly the same.

If this is indeed what is going on, then the main issue is that we get a reported run failure when there really is nothing wrong.

gsrohde:
> I suppose I could test this conjecture by running the script on the previous release version of BioCro and on this one and check that the results are exactly the same.

I did go ahead and run the script using the source versions of BioCro used for runs 2 through 4 of the [Document quantities action](https://github.com/biocro/biocro/actions/workflows/document_quantities.yml).  The result files are identical in all three cases.

I'm still thinking about what the best way is to avoid a workflow failing result in this case.

eloch216:
I guess it might not need to be a hotfix, since it seems this won't be an issue until there is another release.

Issue #20, opened 2022-04-15
Update GitHub actions imported from biocro-dev


gsrohde:
* [x] Update references to `biocro-dev` contained in GitHub workflow action-related files
* [x] Give `ebimodeling/biocro` actions write permission to `ebimodeling/biocro-documentation`
* [ ] Provide separate locations for documentation generated from branch `main` and documentation generated from other branches, including (if needed) from `ebimodeling/biocro-dev` branches [wait on this until possible relocation of repository to a new organization has been completed]
* [x] Change _check_ workflow to test all platforms by default
* [ ] Possibly generate documentation for each major release (2.0.0 and later)

gsrohde:
This issue was written with `ebimodeling/biocro` in mind.  I think all the changes mentioned here that we want to do, suitably modified to apply to `biocro/biocro`, have been done.  Closing.

PR #93, opened 2024-03-08
Prevent bogus error result from document quantities workflow

gsrohde:
This addresses issue #92.

### DIFF
--- a/.github/workflows/document_quantities.yml
+++ b/.github/workflows/document_quantities.yml
@@ -108,14 +108,17 @@ jobs:
     - name: 12. Commit updates and push to target (documentation) repository
       run: |
         cd target
-        if git diff --quiet HEAD; then
+
+        # stage any new or changed files:
+        git add .
+
+        if git diff --quiet --cached; then
           echo "no changes to commit"
         else
           echo "committing changes"
           # user.name and user.email are somewhat arbitrary but must be provided:
           git config user.name biocro-action
           git config user.email bogus@url.com
-          git add .
           git commit -m "updated quantities documentation"
           git push
         fi
--- a/.github/workflows/document_quantities.yml
+++ b/.github/workflows/document_quantities.yml
@@ -108,10 +108,14 @@ jobs:
     - name: 12. Commit updates and push to target (documentation) repository
       run: |
         cd target
-
-        # user.name and user.email are somewhat arbitrary but must be provided:
-        git config user.name biocro-action
-        git config user.email bogus@url.com
-        git add .
-        git commit -m "updated quantities documentation"
-        git push
+        if git diff --quiet HEAD; then
+          echo "no changes to commit"
+        else
+          echo "committing changes"
+          # user.name and user.email are somewhat arbitrary but must be provided:
+          git config user.name biocro-action
+          git config user.email bogus@url.com
+          git add .
+          git commit -m "updated quantities documentation"
+          git push
+        fi


eloch216:
I Googled the `git diff --quiet HEAD;` part to see what it does, and stumbled across this: https://stackoverflow.com/questions/28296130/what-does-this-git-diff-index-quiet-head-mean

Would it be safer to use `git diff-index --quiet HEAD --;` instead?

gsrohde:
> I Googled the `git diff --quiet HEAD;` part to see what it does, and stumbled across this: https://stackoverflow.com/questions/28296130/what-does-this-git-diff-index-quiet-head-mean
> 
> Would it be safer to use `git diff-index --quiet HEAD --;` instead?

Funny you should bring up `git diff-index`.  That was the command I had originally before switching to `git diff`.  (Not sure why—I copied it from a script in another project I was working on.)  But I was getting really wonky behavior: the command was returning a non-zero exit code when nothing had changed, and then the "else" branch was erroring out when it tried to "git commit" zero changes.

But then I added a "git diff" call a few lines before the "if" test to try to debug this, and suddenly everything worked: that is, `git diff-index` now correctly returned an exit code of 0 and the "else" branch with the failing "git commit" call was avoided.  At this point, I felt I was a victim of the observer effect, threw up my hands in my attempt to debug this, and then wondered why I was using `git diff-index` to begin with when I was really much more familiar with `git diff`, so I switched to it, and the problem went away.  It's also my understanding that "porcelain" commands (`git diff` in this case) are considered more stable than "plumbing" commands (`git diff-index`), and so preferred.

(To see the effect of the addition of `git diff`, compare [Run #20](https://github.com/biocro/biocro/actions/runs/8192487193), which has the extra `git diff` call and succeeds, with [Run #21](https://github.com/biocro/biocro/actions/runs/8192636983), where I again removed it.  The respective commit trees are tagged [Run-20](https://github.com/biocro/biocro/releases/tag/Run-20) and [Run-21](https://github.com/biocro/biocro/releases/tag/Run-21).  By the way, I did manage to find a similar though clearly different bug reported about `diff-index` (see https://github.com/git/git/commit/6a044a20480a8ef56f7ddb8142f660ca01a3391e).  I don't know, however, whether that bug is also subject to the observer effect.)

You do have a good point about adding `--` at the end of the command, however.  `git diff --quiet HEAD` may not give the result we want if there happens to be a file called `HEAD` at the top level.

I opted for a slightly different solution though.  First, I moved the `git add .` call to before the `if` test.  This corrects a possible problem which might occur if we ever changed the names of the files output by the script or ever changed the location of the documentation repository: if the files copied in step 11 are not being tracked yet, the `git diff` call won't pick up any differences, and the "nothing to commit" branch will be followed.  `git add .` will ensure that every file under the current directory is being tracked and that any new or changed files are staged.  Then I use `git diff --quiet --cached` to check whether the `git add` actually staged anything.  (I could have used `git diff --quiet HEAD --` just as easily in this context, since it detects if there are any tracked files in the working copy (staged or not) that differ from what's in `HEAD`.  Or I could have used `git diff --quiet --cached HEAD --`, which is exactly equivalent to `git diff --quiet --cached`, since `<commit>` defaults to `HEAD` if not specified.)

eloch216 (approved):

gsrohde:
I realized I didn't add anything to `NEWS.md`.  But that brings up a question: Do we even want an item about this in the news file?  These sorts of changes are hardly relevant, it would seem, to general users.  The changed files, in fact, don't get included in even the source version of the BioCro R package.  (I first became aware of this issue in reading through the `NEWS.md` file while assessing PR #88 and the imminent submission of the package to CRAN.)

eloch216:
> I realized I didn't add anything to `NEWS.md`. But that brings up a question: Do we even want an item about this in the news file? These sorts of changes are hardly relevant, it would seem, to general users. The changed files, in fact, don't get included in even the source version of the BioCro R package. (I first became aware of this issue in reading through the `NEWS.md` file while assessing PR #88 and the imminent submission of the package to CRAN.)

My two cents: I think it's natural to treat `NEWS.md` as the changelog for the entire repository, not just the files included with the R package. Otherwise we would need to have separate changelogs for different parts of the repository, which seems like a lot of work for very little benefit. Things that aren't relevant to users can be included in a subheading for things that only developers would care about. At the same time, I think it's possible to have a "significance threshold" for `NEWS.md`, and we can exclude things that seem small. For example, one could probably argue that this PR, the associated issue, and the git blame will be enough of a record to explain what happened.

gsrohde:
> one could probably argue that this PR, the associated issue, and the git blame will be enough of a record to explain what happened.

I'm either persuaded by that argument or just being lazy.  In any case, I'm merging this.

PR #98, opened 2024-04-19
Replace vlas

gsrohde:
Also eliminated an unneeded cast while I was at it.

### DIFF
--- a/src/module_library/AuxBioCro.cpp
+++ b/src/module_library/AuxBioCro.cpp
@@ -6,6 +6,7 @@
 #include <stdexcept>  // for std::out_of_range, std::range_error
 #include <algorithm>  // for std::max, std::min
 #include <cmath>      // for exp, log, pow, lgamma, std::abs
+#include <vector>
 #include "c4photo.h"
 #include "BioCro.h"
 #include "water_and_air_properties.h"  // for saturation_vapor_pressure,
@@ -15,6 +16,8 @@
                                        // atmospheric_pressure_at_sea_level,
                                        // molar_mass_of_water, stefan_boltzmann
                                        // celsius_to_kelvin
+
+using std::vector;
 
 // The probability density for the Poisson distribution is
 // e^(-lambda) * lambda^x / x!
@@ -59,11 +62,11 @@ double poisson_density(int x, double lambda)
  * Preconditions:
  *     `WindSpeed` is non-negative.
  *     `LAI` is non-negative
- *     `nlayers` is at least 1 and at most MAXLAY.
- *     `wind_speed_profile` is an array of at least size `nlayers`.
+ *     `wind_speed_profile` is an array of size at most MAXLAY.
  */
-void WINDprof(double WindSpeed, double LAI, int nlayers, double* wind_speed_profile)
+void WINDprof(double WindSpeed, double LAI, vector<double> wind_speed_profile)
 {
+    auto nlayers { wind_speed_profile.size() };
     constexpr double k = 0.7;
     double LI = LAI / nlayers;
 
@@ -179,8 +182,9 @@ void RHprof(double RH, int nlayers, double* relative_humidity_profile)
     }
 }
 
-void LNprof(double LeafN, double LAI, int nlayers, double kpLN, double* leafN_profile)
+void LNprof(double LeafN, double LAI, double kpLN, vector<double> leafN_profile)
 {
+    auto nlayers { leafN_profile.size() };
     double LI = LAI / nlayers;
     for (int i = 0; i < nlayers; ++i) {
         double CumLAI = LI * (i + 1);
@@ -1087,7 +1091,7 @@ rd_str rootDist(int n_layers, double rootDepth, double* depths, double rfl)
     double layerDepth = 0.0;
     double CumLayerDepth = 0.0;
     int CumRootDist = 1;
-    double rootDist[n_layers];
+    vector<double> rootDist(n_layers);
     double cumulative_a = 0.0;
 
     for (int i = 0; i < n_layers; ++i) {
@@ -1106,7 +1110,7 @@ rd_str rootDist(int n_layers, double rootDepth, double* depths, double rfl)
 
     for (int j = 0; j < n_layers; ++j) {
         if (j < CumRootDist) {
-            double a = poisson_density(j + 1, (double)CumRootDist * rfl);
+            double a = poisson_density(j + 1, CumRootDist * rfl);
             rootDist[j] = a;
             cumulative_a += a;
         } else {
--- a/src/module_library/AuxBioCro.cpp
+++ b/src/module_library/AuxBioCro.cpp
@@ -66,11 +66,11 @@ double poisson_density(int x, double lambda)
  */
 void WINDprof(double WindSpeed, double LAI, vector<double>& wind_speed_profile)
 {
-    auto nlayers { wind_speed_profile.size() };
+    auto nlayers = wind_speed_profile.size();
     constexpr double k = 0.7;
     double LI = LAI / nlayers;
 
-    for (int i = 0; i < nlayers; ++i) {
+    for (vector<double>::size_type i = 0; i < nlayers; ++i) {
         double CumLAI = LI * (i + 1);
         wind_speed_profile[i] = WindSpeed * exp(-k * (CumLAI - LI));
     }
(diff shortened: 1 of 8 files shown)


justinmcgrath:
vector.size() returns the number of elements, not the index of the largest element.

src/module_library/AuxBioCro.cpp
```diff
@@ -59,11 +62,11 @@ double poisson_density(int x, double lambda)
...
  *     `WindSpeed` is non-negative.
  *     `LAI` is non-negative
- *     `nlayers` is at least 1 and at most MAXLAY.
- *     `wind_speed_profile` is an array of at least size `nlayers`.
+ *     `wind_speed_profile` is an array of size at most MAXLAY.
  */
-void WINDprof(double WindSpeed, double LAI, int nlayers, double* wind_speed_profile)
+void WINDprof(double WindSpeed, double LAI, vector<double> wind_speed_profile)
 {
+    auto nlayers { wind_speed_profile.size() };
```
justinmcgrath, line 69: I'm trying to do a few other things, so I'm not looking too closely here, but isn't this wrong? -- vector.size() - 1 to loop through elements, or a different loop condition.

justinmcgrath: No, it's right. I'm not sure why this is producing different results.

gsrohde: Yeah, I don't know what's going on.  The automated tests are failing.  I'll try testing locally and see if I can figure it out.

justinmcgrath:
Ed has found that all of the new windspeed profile values are 0, so maybe that helps find the problem.

src/module_library/AuxBioCro.cpp
eloch216: I checked out the windspeed outputs, and they have all become 0 for every layer

gsrohde: Yes, I did notice that in the test results.  I'll look into it.

eloch216: I think the `wind_speed_profile` vector needs to be passed by reference, since the `WINDprof` function is expected to modify the profile

gsrohde: Ah, yes, of course.

justinmcgrath: For this, the function can accept values by reference, or it can accept as a pointer, then use "address of" on the vector. As a matter of habit, I accept by reference if a function doesn't modify the value, and accept as a pointer when it modified the value. This makes a clearer distinction between mutating and nonmutating functions, but I don't think it's a common enough convention to strongly advocate for it. Either one is fine.

gsrohde: LNprof too.  Once I changed these, the tests pass on my local machine, and so I expect they'll pass in the automated tests (knock on wood).

gsrohde: > For this, the function can accept values by reference, or it can accept as a pointer, then use "address of" on the vector. As a matter of habit, I accept by reference if a function doesn't modify the value, and accept as a pointer when it modifies the value. This makes a clearer distinction between mutating and nonmutating functions, but I don't think it's a common enough convention to strongly advocate for it. Either one is fine.

I use const references when the function won't modify the value and non-const references when they do.  I try to avoid pointers if at all possible and use "smart pointers" (unique_ptr or shared_ptr) if I can't avoid them altogether.

eloch216: Looks like Windows isn't happy. Justin and I figured out that the type of `i` should be `vector<double>::size_type` rather than `int` in the `for` loop in `WINDprof`.

gsrohde:
Well "windows-latest using R version 3.6.0" doesn't like some of the new code in `AuxBioCro.cpp`—namely, it doesn't like that I compare a `vector<double>::size_type` variable (`nlayers`) and an `int` in lines 73 and 189, and it doesn't like that I divide a double by `nlayers` in lines 71 and 188.

src/module_library/AuxBioCro.cpp
gsrohde: This wouldn't solve the problem of dividing `LAI / nlayers` in lines 71 and 188.

I thought of casting `leafN_profile.size()` and `wind_speed_profile.size()` to `int`s, but maybe it's better to use `vector<double>::size_type` for the type of `i` in the loop and then cast `nlayers` to a double in the divisions.

eloch216: Maybe `nlayers` could be set to a `vector<double>::size_type` rather than `auto`?

gsrohde: > Maybe `nlayers` could be set to a `vector<double>::size_type` rather than `auto`?

I think they should do exactly the same thing for any standards-compliant compiler.

gsrohde: Actually, I think I goofed: when using auto, I think you are supposed to use the `=` syntax for initialization.

justinmcgrath: auto will make the variable type an initializer_list, when it needs to be the type in the list.

eloch216: Doesn't the compiler error from Windows with R v3.6.0 indicate that `auto` is turning up a different type?

> module_library/AuxBioCro.cpp: In function 'void WINDprof(double, double, std::vector<double>&)':
> module_library/AuxBioCro.cpp:71:21: error: no match for 'operator/' (operand types are 'double' and 'std::initializer_list<long long unsigned int>')
>      double LI = LAI / nlayers;

gsrohde: Yes, didn't catch that.  (See my preceding comment too.)

gsrohde: > Doesn't the compiler error from Windows with R v3.6.0 indicate that `auto` is turning up a different type (`std::initializer_list<long long unsigned int>`)?

OK, I think I fixed the initialization problem and the type comparison problem in the "for" loop.  I'm going to assume `nlayers` will automatically get converted in the division now that it has a sensible type, but I'll add the `static_cast` call if it turns out to be necessary.

gsrohde: > Actually, I think I goofed: when using auto, I think you are supposed to use the `=` syntax for initialization.

For what it's worth (and I think it is worth something), I'll quote Stroustrup's extensive discussion of this:


> 6.3.6.2. auto and {}-lists
> 
> When we explicitly mention the type of an object we are
> initializing, we have two types to consider: the type of the
> object and the type of the initializer. For example:
> 
> ```
> char v1 = 12345;      //12345 is an int
> int v2 = 'c';         //'c' is a char
> T v3 = f();
> ```
> 
> By using the {}-initializer syntax for such definitions, we
> minimize the chances for unfortunate conversions:
> 
> ```
> char v1 {12345};      // error: narrowing
> int v2 {'c'};         // fine: implicit char->int conversion
> T v3 {f()};           // works if and only if the type of f() can be implicitly converted to a T
> ```
> 
> When we use auto, there is only one type involved, the type of
> the initializer, and we can safely use the = syntax:
> 
> ```
> auto v1 = 12345;       // v1 is an int
> auto v2 = 'c';         // v2 is a char
> auto v3 = f();         // v3 is of some appropriate type
> ```
> 
> In fact, it can be an advantage to use the = syntax with auto,
> because the {}-list syntax might surprise someone:
> 
> ```
> auto v1 {12345};        // v1 is a list of int
> auto v2 {'c'};          // v2 is a list of char
> auto v3 {f()};          // v3 is a list of some appropriate type
> ```
> 
> This is logical. Consider:
> 
> ```
> auto x0 {};          // error: cannot deduce a type
> auto x1 {1};         // list of int with one element
> auto x2 {1,2};       // list of int with two elements
> auto x3 {1,2,3};     // list of int with three elements
> ```
> 
> The type of a homogeneous list of elements of type T is taken to
> be of type initializer_list<T> (§3.2.1.3, §11.3.3). In
> particular, the type of x1 is not deduced to be int. Had it been,
> what would be the types of x2 and x3?
> 
> Consequently, I recommend using = rather than {} for objects specified auto whenever we don’t mean “list.”
> 
[Bjarne, Stroustrup. C++ Programming Language, The (pp. 164-165). Pearson Education. Kindle Edition.]

gsrohde: It looks like "macOS-latest using R version 3.6.0" failed as well, but for a different reason.  It's hard to fathom why the changes to the C++ code should cause a problem with building the vignettes.

gsrohde:
With my latest fixes, all the automated builds checked out okay.

eloch216 (approved):
Thanks, looks good!

justinmcgrath:
Thanks.

With Ed's pointer to MikTex, I also don't have any Latex warnings now, and as a bonus it automatically installs missing packages rather than failing. I'll send this to CRAN shortly.

PR #100, opened 2024-04-22
Reduce package size

eloch216:
Just following up on something from the email thread. I looked into changing the output step size for the soybean regression test. It looks like the results are pretty consistent across a range of step sizes. This can be checked with the following code:
```r
library(BioCro)
library(lattice)

test_step_size <- function(step_size) {
        initial_values,
        parameters,
        soybean_weather[['2002']],
        direct_modules,
        differential_modules,
        within(ode_solver, {output_step_size = step_size})
    )})
    
    res$output_step_size <- step_size
    res$final_pod <- res[nrow(res), 'Grain']
    res$nrow <- nrow(res)
    res
}

step_size_seq <- c(1, 2, 3, 6, 9, 12, 18, 24)

full_res <- do.call(rbind, lapply(step_size_seq, test_step_size))

xyplot(
    Leaf + Stem + Root + Grain ~ time,
    group = output_step_size,
    data = full_res,
    type = 'l',
    auto = TRUE
)

dev.new()
xyplot(
    final_pod ~ output_step_size,
    data = unique(full_res[, c('final_pod', 'output_step_size')]),
    type = 'p',
    pch = 16
)

dev.new()
xyplot(
    nrow ~ output_step_size,
    data = unique(full_res[, c('nrow', 'output_step_size')]),
    type = 'p',
    pch = 16
)
```

Here are the plots I see:

image

image

image

In this PR, I changed the output step to 6. This reduces the stored test result CSV to 1/6 of its original size. I want to see if the tests pass on all OSes. I suspect that differences in rounding on different OSes may become more pronounced with a larger output step size.

I think this will be another hotfix so I am requesting to merge into `main`. I haven't updated `NEWS` or the version yet in case we want to make other changes, or in case this fails on other OSes.

### DIFF
--- a/tests/testthat/crop_model_testing_helper_functions.R
+++ b/tests/testthat/crop_model_testing_helper_functions.R
@@ -100,11 +100,15 @@ SOYBEAN_IGNORE <- c(
     "sunlit_EPriestly_layer_9"
 )
 
+# Change the output step size for the soybean simulation
+soybean_test <- soybean
+soybean_test$ode_solver$output_step_size <- 6
+
 # Define the plants to test
 PLANT_TESTING_INFO <- list(
     specify_crop("miscanthus_x_giganteus", TRUE,  miscanthus_x_giganteus, WEATHER,                MISCANTHUS_X_GIGANTEUS_IGNORE), # INDEX = 1
     specify_crop("willow",                 TRUE,  willow,                 WEATHER,                WILLOW_IGNORE),                 # INDEX = 2
-    specify_crop("soybean",                TRUE,  soybean,                soybean_weather$'2002', SOYBEAN_IGNORE)                 # INDEX = 3
+    specify_crop("soybean",                TRUE,  soybean_test,           soybean_weather$'2002', SOYBEAN_IGNORE)                 # INDEX = 3
 )
 
 # Make a helping function that runs a simulation for one crop
@@ -164,6 +168,8 @@ update_stored_results <- function(test_info) {
 # Make a helping function that checks the result of a new simulation against the
 # stored data for one crop
 test_plant_model <- function(test_info) {
+    # Read the stored result from the data file
+    stored_result <- read.csv(test_info[['stored_result_file']])
 
     # Describe the current test
     description_validity <- paste(
@@ -217,7 +223,7 @@ test_plant_model <- function(test_info) {
         )
 
         test_that(description, {
-            expect_equal(nrow(new_result), nrow(test_info[['drivers']]))
+            expect_equal(nrow(new_result), nrow(stored_result))
         })
 
         # If the simulation finished, make additional checks
@@ -243,9 +249,6 @@ test_plant_model <- function(test_info) {
                     warning(msg)
                 }
             }
-
-            # Read the stored result from the data file
-            stored_result <- read.csv(test_info[['stored_result_file']])
 
             # Make sure all columns contain numeric data
             stored_result <- as.data.frame(sapply(stored_result, as.numeric))


eloch216:
Note that we cannot take the same approach for the miscanthus and willow models. They use the `thermal_time_senescence` and `thermal_time_and_frost_senescence` modules, which are only compatible with an Euler solver using a step size of 1.

We could do something more drastic, such as only saving a subset of the simulation output rows. This makes it more difficult to check that each call to `run_biocro` produces the correct number of rows in the output, though, which is an important test. I have also found the full stored results to be very useful for manually checking to see whether bug fixes or other "small" changes have a significant impact on the simulation output.

I am also a bit wary of putting the regression test data or weather data somewhere else. I think this will make it more difficult for us to write the tests, and/or for users to run basic simulations.

gsrohde:
This is _probably_ ok?  The tests still pass on my machine.  If you want a thorough check-over, I won't have time to do that until later this afternoon.

(The only thing a little jarring at first glance was using the name `soybean_test` for the stored result when all the other tests just use the crop name.  Maybe `soybean_sparse`?  No biggie though.)

justinmcgrath:
When I asked if they were actually concerned about VLAs and it wasn't an unintended check on Debian, Uwe Liggs sent the message below, asking to get the size below 5 MB, and pointing out the C++11 specification. He didn't mention the VLAs...

I am all but certain there is no way to get this below 5 MB even if we removed everything that isn't strictly necessary (test data, weather data, nonessential of documentation), but we can get it somewhat smaller.

```
We see:


   Found the following (possibly) invalid URLs:
     URL: https://ethos.bl.uk/OrderDetails.do?uin=uk.bl.ethos.268715
       From: inst/doc/thick_layer_absorption.html
       Status: 404
       Message: Not Found

   Size of tarball: 20844066 bytes

Pls reduce to lss than 5 MB.


Flavor: r-devel-linux-x86_64-debian-gcc, r-devel-windows-x86_64
Check: C++ specification, Result: NOTE
     Specified C++11: please drop specification unless essential


So your code does not support more recent versions of the C++ standard?
If kit does, please drop.


Flavor: r-devel-linux-x86_64-debian-gcc, r-devel-windows-x86_64
Check: top-level files, Result: NOTE
   Non-standard file/directory found at top level:
     'inc'

Flavor: r-devel-linux-x86_64-debian-gcc
Check: whether package can be installed, Result: WARNING
     module_library/AuxBioCro.cpp:1090:21: warning: variable length
arrays are a C99 feature [-Wvla-extension]
     module_library/CanAC.cpp:69:31: warning: variable length arrays are
a C99 feature [-Wvla-extension]
     module_library/CanAC.cpp:72:26: warning: variable length arrays are
a C99 feature [-Wvla-extension]
     module_library/c3CanAC.cpp:70:31: warning: variable length arrays
are a C99 feature [-Wvla-extension]
     module_library/c3CanAC.cpp:73:26: warning: variable length arrays
are a C99 feature [-Wvla-extension]
     module_library/multilayer_canopy_properties.cpp:125:31: warning:
variable length arrays are a C99 feature [-Wvla-extension]
     module_library/multilayer_canopy_properties.cpp:129:26: warning:
variable length arrays are a C99 feature [-Wvla-extension]
   See '/srv/hornik/tmp/CRAN/BioCro.Rcheck/00install.out' for details.
   Used C++ compiler: 'Debian clang version 17.0.6 (5)'


Please fix and resubmit.

Best,
Uwe Ligges
```

eloch216:
> (The only thing a little jarring at first glance was using the name `soybean_test` for the stored result when all the other tests just use the crop name. Maybe `soybean_sparse`? No biggie though.)

The stored result is still called `soybean_simulation.csv`. I just made an altered copy of the `soybean` model definition and renamed it to `soybean_test` because it is used for testing.

eloch216:
> Can we keep enough weather data to do tests and put the other weather data in a separate package?

I think we could make another repository for weather data. It could be called `BioCroWeather`. (Feel free to make other suggestions.) It could contain a simple R package with just data (no functions). I think the `DESCRIPTION` file of the main BioCro R package would be able to include it under `remotes` so it would be available to users without any additional steps. E.g. this package does something similar: https://github.com/poales/msuRACiFit/blob/master/DESCRIPTION. 

Right now, most of the examples and tests use `weather[['2005']]` or `soybean_weather[['2002']]`. But, the full set of weather data is used in the "Quantitative Comparison" vignette, so we would need to make sure it is all accessible.

I can work on a separate weather data repository but want to see if there are any comments about this organization scheme. I don't really like it because it makes things more complicated for us. But if it lets us get on CRAN, I suppose it's worthwhile.

eloch216:
I tried a different approach to reducing the stored regression test data. Here I just keep one time point for each day in each simulation. Now the test data is reduced to `1.03 MB` from `27.28 MB` (on `main`). The tarball is down to `9996 KB` with this change.

eloch216:
For saving space on vignettes, we could revisit the distinction between "vignettes" and "articles," where articles are included in the pkgdown site but not with the package itself: https://pkgdown.r-lib.org/articles/pkgdown.html#articles

By Googling "github vignettes/web_only", I found an example of someone using a web-only article / vignette: https://github.com/sahirbhatnagar/manhattanly/tree/master

justinmcgrath:
the signif function probably saves as much space as round but keeps more precision.

gsrohde:
> Could you check whether rounding the weather data would substantially impact the results? There are a lot of digits, which makes it hard to compress, but probably none of them need more than hundredths, and most need less than that.

Probably the same could be said for the results data (test_data).

eloch216:
> Let's see how far we can get without removing the weather data. The HTML documents are large because there is a lot of CSS for the styles, not because of the images, which was my first thought. I'm not sure how to fix that. I did see about post that prettydoc makes smaller files. https://stackoverflow.com/questions/44420766/knitr-generating-very-large-html-files
> 
> The pdfs are large because the pdfs figures are large. It's probably storing them as some vector graphics. I don't know if there's a way to change that format in the pdf, but you could save the figures as png then include those in the pdf. 200 ppi gives small files that still look pretty good. I can't tell what antialiasing does for it, so you could use antialias='none' to save more space.
> 
> If we could move some of these out of the package that would help even more.

I took a rather drastic measure here and moved all the vignettes into a "web only" folder. These vignettes will show up on the website, and they can be built locally with `tools::buildVignette` if the entire source repository is downloaded, but they won't actually be part of the R package. To compensate for this, I added a very small / short HTML vignette that directs readers to the main documentation website.

With this change, the tarball is down to `7113 KB`.

eloch216:
> > Could you check whether rounding the weather data would substantially impact the results? There are a lot of digits, which makes it hard to compress, but probably none of them need more than hundredths, and most need less than that.
> 
> Probably the same could be said for the results data (test_data).

I will look into this

gsrohde:
> We could do something more drastic, such as only saving a subset of the simulation output rows. This makes it more difficult to check that each call to `run_biocro` produces the correct number of rows in the output, though, which is an important test.

I'm not sure I'm completely understanding how these tests work, but couldn't we store and test the number of rows separately?

> I have also found the full stored results to be very useful for manually checking to see whether bug fixes or other "small" changes have a significant impact on the simulation output.

Maybe we could store "complete" results in test-data files that are ".Rignored" but also have smaller test-data files for distribution.  Then maybe we could implement some kind of flag could be turned on by developers when they want to use the complete data.

justinmcgrath:
> Probably the same could be said for the results data (test_data).

We could probably keep fewer digits in the results files, but I'd want to be conservative in this because part of their purpose is to see anything that makes a difference in results.

> Maybe we could store "complete" results in test-data files that are ".Rignored" but also have smaller test-data files for distribution. Then maybe we could implement some kind of flag could be turned on by developers when they want to use the complete data.

This seems like a good approach. Do we need any test data in the CRAN package? We could keep all of the digits and be sure to see small changes, but then CRAN, and users, don't get files they don't need.

eloch216:
I tried truncating the weather data digits using `signif` with `digits = 3`, and it gave us some significant savings. The tarball is now down to `5362 KB` when I build it locally.

justinmcgrath:
Can you round solar? Something like this `newweather = lapply(weather, function(x) within(x, {precip=signif(precip, 3); rh=signif(rh, 3); solar=round(solar); temp=signif(temp, 3); windspeed=signif(windspeed, 3)}))`

It's probably not necessary to precision less than 1 micromole, and maybe we can actually squeeze under 5 MB.

Does this truncation impact the results significantly?

eloch216:
I also just rounded the stored regression test data to 5 digits. We are down to `5182 KB` for the tarball.

justinmcgrath:
Can we choose the compression level when building the package? The default is 6, but gzip has a 9 that doesn't take much longer and is a bit smaller.

eloch216:
> Can we choose the compression level when building the package? The default is 6, but gzip has a 9 that doesn't take much longer and is a bit smaller.

Seems like that might not be an option:

```
C:\Users\eloch\Documents\GitHub>R CMD build --help

Using Ed's custom R settings on Mon Apr 22 17:03:51 2024
Usage: R CMD build [options] pkgdirs

Build R packages from package sources in the directories specified by
'pkgdirs'

Options:
  -h, --help            print short help message and exit
  -v, --version         print version info and exit

  --force               force removal of INDEX file
  --keep-empty-dirs     do not remove empty dirs
  --no-build-vignettes  do not (re)build package vignettes
  --no-manual           do not build the PDF manual even if \Sexprs are present
  --resave-data=        re-save data files as compactly as possible:
                        "no", "best", "gzip" (default)
  --resave-data         same as --resave-data=best
  --no-resave-data      same as --resave-data=no
  --compact-vignettes=  try to compact PDF files under inst/doc:
                        "no" (default), "qpdf", "gs", "gs+qpdf", "both"
  --compact-vignettes   same as --compact-vignettes=qpdf
  --compression=        type of compression to be used on tarball:
                        "gzip" (default), "none", "bzip2", "xz"
  --md5                 add MD5 sums
  --log                 log to file 'pkg-00build.log' when processing
                        the pkgdir with basename 'pkg'

Report bugs at <https://bugs.R-project.org>.
```

justinmcgrath:
I tried --resave-data, but I didn't see any effect. I think this is good though. When I run R CMD check, I no longer get a message about the installation size. It's just new package, 'inc' directory, C++11 and Gnu Make. I'll send this and we can see what they say.

gsrohde:
> I also just rounded the stored regression test data to 5 digits. We are down to `5182 KB` for the tarball.

Well I never did finish that course I once took in numerical analysis, but my impression is—depending on how the input factors into the result—if the input only has 3 s.d., the output probably doesn't _really_ have 5 s.d.

justinmcgrath:
> > I also just rounded the stored regression test data to 5 digits. We are down to `5182 KB` for the tarball.
> 
> Well I never did finish that course I once took in numerical analysis, but my impression is—depending on how the input factors into the result—if the input only has 3 s.d., the output probably doesn't _really_ have 5 s.d.

The point of the test results is to see if changes to the code change the results though, so it doesn't really matter what the real precision of the estimate is. I can see rounding a bit so that changes that are from machine error don't effect the results, but that would only be getting rid of a couple digits, not all but 5. At any rate, 5 is still a good number and this let's us move forward with this. If it ends up being a problem we can revisit it later.

justinmcgrath:
It's not small enough. Instead of keeping three sig dogs for solar, could you round to 0 digits? If that's not enough, then keep 2 signfogs instead of three for rh.

They won't allow "inc" at top level. They give no reason for this, and one could only speculate why anyone cares about such a meaningless thing.

Given the level of dumb shit they care about I can all but guarantee that they won't allow the long file names that result, so I think they need to be shortened.

justinmcgrath:
Even if rounding the data gets us under 5 MB, in the long term, we're going to need to move things out of the package. Otherwise we'll quickly run into limits if we want to add new modules.

eloch216:
> Even if rounding the data gets us under 5 MB, in the long term, we're going to need to move things out of the package. Otherwise we'll quickly run into limits if we want to add new modules.

Definitely. In the long run, we are already planning to split out the modules into a separate repository. When we do that, we can think about keeping the weather data with the modules, or moving it somewhere else. For right now, I think we should just do what we can to get the current repo on CRAN. These other operations will be a lot easier once that's done.

eloch216:
I rounded `solar` and `rh` further, and we are below 5 MB now! Moving the boost library will be more complicated since it will involve changes to two repositories, but I will start on it.

justinmcgrath:
I don't know that this would be easier, but I think the boost directory could be directly in src, as src/boost, saving the need to rename boost files. One downside is that if new libraries are added, they need to be added to the Makefile, which isn't too big of a problem, especially since that's unlikely or at least uncommon. It also seems to make a mess of the accessory files (NEWS.md, README.md, etc), which is more of a problem.

eloch216:
After moving the boost files to `src`, we now have a new `WARNING`:

```
* checking pragmas in C/C++ headers and code ... WARNING
File which contains pragma(s) suppressing important diagnostics
  'src/inc/boost/lexical_cast/try_lexical_convert.hpp'
Files which contain pragma(s) suppressing diagnostics:
  'src/inc/boost/atomic/detail/extra_fp_ops_generic.hpp'
  'src/inc/boost/get_pointer.hpp'
  'src/inc/boost/move/detail/std_ns_begin.hpp'
  'src/inc/boost/mpl/assert.hpp'
  'src/inc/boost/multi_index/detail/ignore_wstrict_aliasing.hpp'
  'src/inc/boost/smart_ptr/detail/shared_count.hpp'
  'src/inc/boost/smart_ptr/scoped_ptr.hpp'
  'src/inc/boost/smart_ptr/shared_ptr.hpp'
  'src/inc/boost/tuple/detail/tuple_basic.hpp'
  'src/inc/boost/type_traits/detail/has_prefix_operator.hpp'
  'src/inc/boost/type_traits/has_logical_not.hpp'
```

I guess we were shielded from this before because the boost files were not in `src`.

eloch216:
There are two paths that need to be shortened:

```
  BioCro/src/inc/boost/numeric/odeint/stepper/generation/generation_controlled_adams_bashforth_moulton.hpp
  BioCro/src/inc/boost/numeric/odeint/stepper/generation/generation_runge_kutta_cash_karp54_classic.hpp
```

As far as hassles go, this one isn't too bad since we don't need to rename the files very often. I will make a branch / PR for this over on `biocro/inc`. We might also need to make changes there related to the pragma warning.

justinmcgrath:
The issue with port seems like a problem with R's code, in particular something with 3.6.0. I don't care to figure out what the issue is, so we can use a later version.

I don't know what to do about Boost pragmas. It's not our code and we're not going to change it. The degree to which these people care about things that have nothing to do with them is insane. We can try shortening the names and for this warning say it's not our code. Then I assume they'll say no, and we'll never get this on CRAN.

justinmcgrath:
Maybe a sed script could remove or comment out the pragmas easily.

eloch216:
It looks like the BH package just comments out any pragmas that conflict with CRAN policy: https://github.com/eddelbuettel/bh/commit/313da3cb04227d96daafa2d73668b30fa3fd10fc

I also took a look at the latest version of boost (1.85), and least some of these pragmas have been removed. (The ones in `'src/inc/boost/lexical_cast/try_lexical_convert.hpp'`)

gsrohde:
> It looks like the BH package just comments out any pragmas that conflict with CRAN policy: [eddelbuettel/bh@313da3c](https://github.com/eddelbuettel/bh/commit/313da3cb04227d96daafa2d73668b30fa3fd10fc)
> 
> I also took a look at the latest version of boost (1.85), and least some of these pragmas have been removed. (The ones in `'src/inc/boost/lexical_cast/try_lexical_convert.hpp'`)

Yes, I was a little worried that commenting out the pragmas would make other warnings show up.  The fact that BH and now Boost is removing them makes me think that maybe doing so won't cause problems.  But have you checked?

eloch216:
> > It looks like the BH package just comments out any pragmas that conflict with CRAN policy: [eddelbuettel/bh@313da3c](https://github.com/eddelbuettel/bh/commit/313da3cb04227d96daafa2d73668b30fa3fd10fc)
> > I also took a look at the latest version of boost (1.85), and least some of these pragmas have been removed. (The ones in `'src/inc/boost/lexical_cast/try_lexical_convert.hpp'`)
> 
> Yes, I was a little worried that commenting out the pragmas would make other warnings show up. The fact that BH and now Boost is removing them makes me think that maybe doing so won't cause problems. But have you checked?

I started checking. I commented out the "diagnostic ignored" pragmas in the first three files listed in the warning. Those files have disappeared from the pragma warnings, but now another one has popped up (which had been ignored before):
```
  ../src/inc/boost/get_pointer.hpp:48:40: warning: 'template<class> class std::auto_ptr' is deprecated: use 'std::unique_ptr' instead [-Wdeprecated-declarations]
```

So I guess there may be additional problems when all the pragmas are commented out.

gsrohde:
I don't know much about algorithm the Boost script we use uses to select files from Boost, but by adding the `-H` option to the C++ compile command, I've found that nearly half the files we include never get used in the compilation.  Just to check that I wasn't missing anything, I tried removing all the files that didn't get mentioned.  (This reduces the number of files under `inc` from 3756 to 1943, and the uncompressed directory tree size from 39M to 15M.)  Then I reran `R CMD INSTALL` with the `--preclean` option and reran the tests as well, and all went smoothly.

justinmcgrath:
> I don't know much about algorithm the Boost script we use uses to select files from Boost, but by adding the `-H` option to the C++ compile command, I've found that nearly half the files we include never get used in the compilation. Just to check that I wasn't missing anything, I tried removing all the files that didn't get mentioned. (This reduces the number of files under `inc` from 3756 to 1943, and the uncompressed directory tree size from 39M to 15M.) Then I reran `R CMD INSTALL` with the `--preclean` option and reran the tests as well, and all went smoothly.

That's a nice finding. I had assumed that bcp was giving the minimal set of files. Boost compresses very well, but that still seems like enough that we could probably add back everything we removed and still have space to spare.

eloch216:
Looks like we only have the weird failure on Mac OS with R v3.6.0. I tried looking into this. I'm not sure if this is helpful, but here's what I found:

First, I visited https://api.r-hub.io/rversions/. There, I learned two queries: https://api.r-hub.io/rversions/available/macos/arm64 and https://api.r-hub.io/rversions/available/macos/. From these, it seems that R version 3.6.0 is not available for Mac OS with `arm64` architecture, but it is available with the default architecture (`x86_64`). I don't know if the workflow recently changed the architecture for MacOS, or if R version 3.6.0 suddenly became unavailable.

gsrohde:
> I don't know if the workflow recently changed the architecture for MacOS, or if R version 3.6.0 suddenly became unavailable.

I noticed the following in the commit logs for the `r-lib/actions/setup-r` action:
```
commit 30cc889da92691a41efe29876fe70d797039033d
Author: Gábor Csárdi <csardi.gabor@gmail.com>
Date:   Thu Apr 25 12:31:45 2024 +0200

    check-full: no need to test on R 3.6 any more
    
    It is now not supported.
```

I tried pinning the version of this action to a specific (older) commit (older than the commit that removed 3.6 support) instead of just the v2 branch, but I still get this error:
```
Error: Error: Failed to resolve R version 3.6.0 at 
    https://api.r-hub.io/rversions/resolve/3.6.0/mac/arm64.
```
It seems like there should be an option to choose `x86_64` over `arm64`, but I didn't see one, and I doubt it's worth the effort to come up with our own implementation of this.

eloch216:
> > I don't know if the workflow recently changed the architecture for MacOS, or if R version 3.6.0 suddenly became unavailable.
> 
> I noticed the following in the commit logs for the `r-lib/actions/setup-r` action:
> 
> ```
> commit 30cc889da92691a41efe29876fe70d797039033d
> Author: Gábor Csárdi <csardi.gabor@gmail.com>
> Date:   Thu Apr 25 12:31:45 2024 +0200
> 
>     check-full: no need to test on R 3.6 any more
>     
>     It is now not supported.
> ```

Interesting. I was wondering what "it is not supported" means. I.e., who is it not supported by? It looks like CRAN itself no longer provides Mac OS X binaries for R < 4.0.0: https://cran.r-project.org/bin/macosx/base/. In contrast, Windows versions are available all the way back to R v1.0.0: https://cran.r-project.org/bin/windows/base/old/.

Maybe we should switch the "old" Mac OS check to R v4.0.0 instead of 3.6.0. The support for 3.6.0 is mostly for Justin, who does not use Mac. I don't like having this red "R CMD check failing" icon on our page for something that has nothing to do with our code.

justinmcgrath:
Support the the Ubuntu LTS I use is going to end soon (some few month, I don't remember the number) and I will update that OS. In the meantime, there is no need to ensure that this works for 3.6. I don't think major changes will be made at this point, and if somehow support for 3.6 is lost, I just won't update BioCro on that computer.

gsrohde:
> In the meantime, there is no need to ensure that this works for 3.6.

I'm unclear what the plan is.  Are we changing the minimum-tested R version to 4.0.0 for all three OS's or just macOS?  And do you want me to make the changes to the workflow, or are one of you planning to do it?  And if me, should it be in this branch (or a branch off of this branch)?

justinmcgrath:
> > In the meantime, there is no need to ensure that this works for 3.6.
> 
> I'm unclear what the plan is. Are we changing the minimum-tested R version to 4.0.0 for all three OS's or just macOS? And do you want me to make the changes to the workflow, or are one of you planning to do it? And if me, should it be in this branch (or a branch off of this branch)?

I am fine with whatever changes get the test to pass and leave us with something that is not a pain to maintain. I don't know how to edit GitHub Actions, so it won't be me.

It seems like this will apply to everything now, regardless of what's going on with CRAN. In that case, we want this in develop. It's not going to overlap with anything here, so I think branch from then merge back into develop. Then we can bring that into here.

gsrohde:
> I am fine with whatever changes get the test to pass and leave us with something that is not a pain to maintain. I don't know how to edit GitHub Actions, so it won't be me.

OK, I made a PR to update the minimum R version checked on macOS to 4.0.0.  That should make the checks pass, and if they do, I'll request reviews.  If I misunderstood and I should update the minimum to 4.0.0 for other OS's as well, that should be easy to do.

gsrohde:
Oops—looks like 4.1.0 is required for arm64 (https://api.r-hub.io/rversions/available/macos/arm64).

justinmcgrath:
Thoughts about pointing to src/inc/LICENSE_1_0.txt or LICENSE.note in the comment? Benjamin mentioned a Copyright field pointing to a file, so this would be similar to that.

The file LICENSE.note still says "inc/LICENSE_1_0.txt".

eloch216:
> Thoughts about pointing to src/inc/LICENSE_1_0.txt or LICENSE.note in the comment? Benjamin mentioned a Copyright field pointing to a file, so this would be similar to that.
> 
> The file LICENSE.note still says "inc/LICENSE_1_0.txt".

Here is some info about bundling other packages with an R package: https://r-pkgs.org/license.html#sec-how-to-include

It doesn't mention including license info in the authors list, so I am thinking we shouldn't do that.

I can update `LICENSE.note` with the new location.

justinmcgrath:
Ok. This looks good. I'll resubmit this.

justinmcgrath:
Does everything work if the Makefile gives the C++11 specification? Because of the NOTE that's produced, every time I send this it fails the automatic checks and I need to respond to it, explaining again why it's there. If we can remove it from DESCRIPTION and add it to the Makefile, it would spare some hassle. We would do this after this gets on CRAN, but I wanted to ask about it here.

justinmcgrath:
This is on CRAN now. I can install it from Linux. I think it will take a day or two for the Windows binaries to be available. I think it's safe to merge this to main and call it a success.

justinmcgrath:
> (This has essentially become a release branch, although it didn't quite start that way.)

Yeah, thankfully what we've worked out is flexible enough to deal with that. Next time we starting working on something for CRAN, let's try to think of it as a release from the start.

eloch216:
I added a new CRAN badge to the main README, and new (simple) installation instructions. I guess `install.packages('BioCro')` doesn't quite work yet, but it will soon enough.

Issue #94, opened 2024-04-16
Compilation warnings


gsrohde:
Here are some warnings I get when I add the `-Wall` and `-Wextra` to the compilation command:

```
In file included from R_get_all_ode_solvers.cpp:6:
In file included from ./framework/ode_solver_library/ode_solver_factory.h:7:
In file included from ./framework/ode_solver_library/../ode_solver.h:5:
In file included from ../inc/boost/numeric/odeint.hpp:79:
In file included from ../inc/boost/numeric/odeint/stepper/generation.hpp:32:
../inc/boost/numeric/odeint/stepper/generation/generation_controlled_adams_bashforth_moulton.hpp:43:99: warning: unused
      parameter 'stepper' [-Wunused-parameter]
    controller_type operator()( value_type abs_error , value_type rel_error , const stepper_type &stepper )
```

I think we can ignore this type of error coming from Boost, and I won't list any more of them.

There are only two non–Boost-related warnings:

```
clang++ -mmacosx-version-min=10.13 -std=gnu++11 -MT module_library/rue_leaf_photosynthesis.o -MMD -MP -MF .deps/module_library/rue_leaf_photosynthesis.Td -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I../inc -DR_NO_REMAP -Wall -Wextra  -I/usr/local/include   -fPIC  -Wall -g -O2  -c -o module_library/rue_leaf_photosynthesis.o module_library/rue_leaf_photosynthesis.cpp
module_library/rue_leaf_photosynthesis.cpp:83:5: warning: missing field 'iterations' initializer
      [-Wmissing-field-initializers]
    };
```

I'm not sure how the iteration-related code works, but maybe this should be explicitly set to zero here.  In any case, this is probably the only warning that should possibly be dealt with.

```
clang++ -mmacosx-version-min=10.13 -std=gnu++11 -MT module_library/c3photo.o -MMD -MP -MF .deps/module_library/c3photo.Td -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I../inc -DR_NO_REMAP -Wall -Wextra  -I/usr/local/include   -fPIC  -Wall -g -O2  -c -o module_library/c3photo.o module_library/c3photo.cpp
module_library/c3photo.cpp:30:18: warning: unused parameter 'AP' [-Wunused-parameter]
    double const AP,                           // Pa (TEMPORARILY UNUSED)
```

I guess this is a known anomaly.

[...]

PR #101, opened 2024-04-23
Explicitly set iterations to 0

eloch216:
This PR addresses issue https://github.com/biocro/biocro/issues/94.

The radiation use efficiency (RUE) leaf model does not need to be solved iteratively, so there is no number of iterations related to these photosynthesis calculations. Now this output is explicitly set to 0 to avoid a compilation warning.

Regarding the other warning about an unused input argument: The pressure is not currently used, but I think it is likely we will re-implement it in the future, so I didn't want to go through the process of removing it and then adding it back in. Modifying the input arguments to `c3photoC` is tedious because this function is called in so many places throughout BioCro.


eloch216:
Looks like another compiler warning has popped up on Mac OS using the release version of R:

```
  module_library/thermal_time_and_frost_senescence.h:282:13: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
```

Looking at the code, there is definitely something weird going on here. That line calculates something, but doesn't assign the value to anything. I think it is calculating the value of `frost_leaf_death_rate`. I will try looking at the history of this file to see if I can figure out what it should be. Fixing this error might change the willow model output, since it uses this module.

eloch216:
I think I may have introduced this error back on September 5, 2019 during this commit: https://github.com/biocro/biocro/commit/12f55730d56e9ef051915a55e760652aec8d3550

The original code in `thermal_time_and_frost_senescence_calculator.hpp` was:
```
double frost_leaf_death_rate = 0.0;

	if(TTc >= seneLeaf) {	// Leaf senescence has started
		bool A = lat >= 0.0;		// In Northern hemisphere
		bool B = doy_dbl >= 180.0;	// In second half of the year
		if((A && B) || ((!A) && (!B))) {	// Winter in either hemisphere
			if(temp > Tfrostlow) {
				// frost_leaf_death_rate changes linearly from 100 to 0 as temp changes from Tfrostlow to Tfrosthigh
				frost_leaf_death_rate = 100 * (temp - Tfrosthigh) / (Tfrostlow - Tfrosthigh);
				frost_leaf_death_rate = (frost_leaf_death_rate > 100.0) ? 100.0 : frost_leaf_death_rate;	// Is this right? I don't think the death rate can exceed 100, but it CAN become negative. (EBL)
			}
			else frost_leaf_death_rate = 0.0;	// No frost death rate when the temperature is too low?
		}
	}
```

But the version added to `thermal_time_and_frost_senescence` was:
```
// Calculate the leaf death rate due to frost
	double frost_leaf_death_rate = 0.0;
	if(TTc >= seneLeaf) {	// Leaf senescence has started
		bool A = lat >= 0.0;		// In Northern hemisphere
		bool B = doy_dbl >= 180.0;	// In second half of the year
		if((A && B) || ((!A) && (!B))) {	// Winter in either hemisphere
			// frost_leaf_death_rate changes linearly from 100 to 0 as temp changes from Tfrostlow to Tfrosthigh and is limited to [0,100]
			std::max(0.0, std::min(100.0, 100.0 * (Tfrosthigh - temp) / (Tfrosthigh - Tfrostlow)));
		}
	}
```

eloch216:
After fixing the code, there was a small change in the willow simulation output. It was enough to trigger test failures, but had no real impact on the biomass calculations (see below).

image

src/module_library/thermal_time_and_frost_senescence.h
```diff
@@ -271,18 +271,26 @@ void thermal_time_and_frost_senescence::do_operation() const
...
+    // Tfrostlow, the leaf death rate reaches its maximum (100%). For
+    // temperatures between Tfrosthigh and Tfrostlow, the death rate changes
+    // linearly with temperature.
+    bool const sene_started = TTc >= seneLeaf;
+    bool const in_north = lat >= 0.0;
+    bool const late_year = time >= 180.0;
+    bool const in_winter = (in_north && late_year) || (!in_north && !late_year);
+
+    double const base_frost_death_rate =
+        100.0 * std::max(0.0, std::min(1.0, (Tfrosthigh - temp) / (Tfrosthigh - Tfrostlow)));
```
gsrohde, line 290: I was going to suggest moving the `100.0 *` to apply only to the second argument of `std::max`, thereby possibly saving one multiplication operation.  On the other hand, if multiplication by 100 represents, say, conversion to percent, the the way you have it is semantically clearer.

gsrohde (commented):
You probably need to update NEWS again regarding your 3rd commit, but other than that, it all seems OK.

src/module_library/thermal_time_and_frost_senescence.h
eloch216: I believe that's what it represents. This module is a bit weird.

PR #102, opened 2024-04-25
Change minimum R version on macOS checked by the R-check workflow to 4.2.0

gsrohde:


justinmcgrath:
One of the checks doesn't pass - Mac with 4.1 because deSolve can't be built.

These checks all used to pass with this same code. Are these failing because the testing tool has changed?

eloch216:
This is what I had in mind -- keep 3.6.0 for Windows and Ubuntu since those checks are passing without issue, and only switch to a newer minimum version for Mac.

`deSolve` claims it only requires R >= 3.3.0 (https://cran.r-project.org/web/packages/deSolve/index.html) so this is a strange error. It could be a fluke that will go away if we run the test again later.

eloch216:
We could also try R 4.1.3 (the last patch update for 4.1) instead of 4.1.0 in case this issue is related to an R bug that was fixed later

eloch216 (approved):

eloch216:
It's weird that `deSolve` can't be built for R below 4.2.0 on Mac, but I can live with this.

gsrohde:
> It's weird that `deSolve` can't be built for R below 4.2.0 on Mac, but I can live with this.

The little I read, or understood of what I read, suggested to me that 4.1 should be sufficient, so I don't know what the issue is (other than I think it has something to do with ARM/Apple Silicon support), but I didn't think it worth delving into.

I myself am stuck with 10-year-old (Intel) Macs that won't upgrade beyond macOS 10.15, but even so, I can upgrade to R 4.2.3, though not beyond that.  With luck, that will be sufficient for the foreseeable future.  It would be nice, though, if the workflows tested non-ARM versions of macOS as well, but maybe not worth the trouble to try to implement that.

eloch216:
> It would be nice, though, if the workflows tested non-ARM versions of macOS as well, but maybe not worth the trouble to try to implement that.

Agreed -- it would be nice, but we probably shouldn't spend time implementing it.

PR #105, opened 2024-04-30
Move C++11 specification from DESCRIPTION file to Makevars files.

gsrohde:
A simplified version of PR #86.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,7 @@ Suggests:
     lattice,
     deSolve
 VignetteBuilder: knitr
-SystemRequirements: C++11, GNU make
+SystemRequirements: GNU make
 License: MIT + file LICENSE
 LazyData: true
 ByteCompile: TRUE
--- a/src/Makevars
+++ b/src/Makevars
@@ -1,3 +1,5 @@
+CXX_STD = CXX11
+
 PKG_CPPFLAGS+=-I../src/inc -DR_NO_REMAP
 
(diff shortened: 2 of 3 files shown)


gsrohde:
Even though I removed the C++11 requirement from DESCRIPTION, the specification in the Makevars files triggers the NOTE for the Release and Devel versions of R.

eloch216:
> Even though I removed the C++11 requirement from DESCRIPTION, the specification in the Makevars files triggers the NOTE for the Release and Devel versions of R.
> 
> I'm not sure if _any_ standards specification would trigger the NOTE or whether it's just the fact that it's a decade+-old specification. I could try specifying the C++14 to see if the NOTE still appears. My guess? It will.

I tried setting C++14 in the `DESCRIPTION` on another branch (https://github.com/biocro/biocro/pull/81/files) and it still triggered the note: https://github.com/biocro/biocro/actions/runs/7847478921/job/21416504609

justinmcgrath:
CXX_STD looks like an R specific thing, and appears to be identical to setting it in DESCRIPTION. It doesn't seem like this is going to work. I think let's close this pull request and the other similar one (or two) and make a new issue about moving to C++17. If any standard is specified, there will be a NOTE. It's a bad policy, but we can see what it would take to get there. If it's not realistic to achieve, I'll just keep emailing every time we submit the package.

gsrohde:
> I tried setting C++14 in the `DESCRIPTION` on another branch ... and it still triggered the note

I probably knew that from when we were looking at this issue before, but I've been away from it for awhile.

PR #106, opened 2024-05-06
Update canopy model

eloch216:
Now that BioCro is on CRAN, we can work on some model improvements. This PR makes a few long-overdue changes to the canopy model. The most important are:

- I eliminated the "average incident PPFD" and "average absorbed shortwave" quantities. Here, "average" refers to an average between the sunlit and shaded leaves in a canopy layer. The "average absorbed shortwave" had been used as an input to the evapo-transpiration / energy balance calculations, where it plays a major role in determining the leaf temperature. However, it doesn't make sense to use an "average" value for both sunlit and shaded leaves, since each leaf type receives very different levels of light, and hence absorbs very different levels of shortwave light energy. This had been discussed previously: https://github.com/ebimodeling/biocro-dev/issues/137. Note that the equation was in fact not present in the Campbell & Norman textbook.

- I replaced `thick_layer_absorption` with `thin_layer_absorption` for calculating the absorbed shortwave energy in each canopy layer and leaf class. The thick layer equation is not appropriate for a thin leaf layer.

I also added a new vignette that explains some of the rationale and key features of our multilayer canopy model. It includes relevant equations from the Campball & Norman textbook which we use as the source of our model. In fact, I wrote the vignette first and then made changes to the code to ensure it matches the cited model.

Following these changes, there are some differences in the model outputs. It looks like biomass has slightly decreased for C3 crops, while it has slightly increased for Miscanthus. I would guess this is mostly due to changes in leaf temperature. The predicted temperature for sunlit leaves has increased quite a bit. I think this is helpful for Miscanthus but not for willow or soybean.

2024-05-06 crop model outputs

I also attached two PDF versions of the vignette so people can see it without needing to build it. One was built using the CRAN version of BioCro, and one with this new version. There are clear changes in the sunlit / shaded fractions, the absorbed shortwave energy, the leaf-level assimilation rates, and leaf temperatures. 

The end of the vignette calculate the total PPFD absorbed by the canopy. I added this partially in response to some of the discussion in a previous issue (https://github.com/ebimodeling/biocro-dev/issues/241), where it had seemed that an impossible amount of light had been absorbed by the canopy.

[2024-05-06 canopy_photosynthesis CRAN.pdf](https://github.com/biocro/biocro/files/15223699/2024-05-06.canopy_photosynthesis.CRAN.pdf)
[2024-05-06 canopy_photosynthesis new.pdf](https://github.com/biocro/biocro/files/15223700/2024-05-06.canopy_photosynthesis.new.pdf)


eloch216:
BTW, there are more changes that need to be made to the canopy model, especially regarding transpiration and leaf temperature. I am planning to proceed in stages, so there will be more PRs in the future. I think it will be easier to do it this way, rather than making all the changes at once.

src/module_library/c3CanAC.cpp
```diff
@@ -141,9 +141,10 @@ canopy_photosynthesis_outputs c3CanAC(
         // energy balance to get a better temperature estimate using that value
         // of stomatal conductance. Get the final estimate of stomatal
         // conductance using the new value of the leaf temperature.
-        double iabs_diff = light_profile.shaded_absorbed_ppfd[current_layer];  // micromole / m^2 /s
-        double pLeafshade = light_profile.shaded_fraction[current_layer];      // dimensionless
-        double Leafshade = LAIc * pLeafshade;                                  // dimensionless
+        double iabs_diff = light_profile.shaded_absorbed_ppfd[current_layer];    // micromole / m^2 /s
+        double j_diff = light_profile.sunlit_absorbed_shortwave[current_layer];  // J / m^2 / s
```
justinmcgrath, line 145: I have not dug into the model itself, so I don' t know what's going on, but is this supposed to be `sunlit_absorbed_shortwave`?

eloch216: Whoops, this is a copy-paste mistake. It should be shaded.

eloch216:
It looks like there are test failures on Ubuntu related to soil evaporation: https://github.com/biocro/biocro/actions/runs/8972595095/job/24640926611?pr=106. This is surprising. I will try looking into it.

There is also an issue with a vignette: https://github.com/biocro/biocro/actions/runs/8972595097/job/24640910461. This kind of problem is difficult to detect because we set `error = TRUE` in most of our vignettes. This means that vignette errors are not picked up by `R CMD check`. (Also see https://github.com/ebimodeling/biocro-dev/issues/480). I am going to change the vignette error handling and then fix the issue with the Ball-Berry vignette. I think this will be helpful in the future.

justinmcgrath:
I think using the different estimates for sunlit and shaded leaves make sense. You comment doesn't mention changing how the fraction of sunlit leaves was calculated. Why was that change made? What makes a layer thin vs thick?

eloch216:
> Why was how the sunlit fraction calculated changed? I didn't see anything about it in your comment. 

There was an explanation but it was buried in a commit description because I forgot to mention it here: "Previously we had used Equation 15.21 to find the sunlit LAI in the layer, and subsequently divided by the total LAI in the layer to get the fraction. It is simpler and more appropriate to use Equation 15.22 for the fraction of sunlit leaves in an infinitely thin layer."

> What makes a layer thin vs thick?

This is discussed at length in the vignette about thick layer absorption. Basically, for a thin layer, `Q_abs = Q_in * (1 - T - R)`, where `T` and `R` are the reflectance and transmittance. If you have infinitely many such layers, then you have a thick layer with `Q_abs = Q_in * (1 - R - T) / (1 - T)`. We had been using the equation for an infinitely thick layer, which is not appropriate for an infinitesimally small piece of leaf area in a canopy.

justinmcgrath:
> This is discussed at length in the vignette about thick layer absorption. Basically, for a thin layer, `Q_abs = Q_in * (1 - T - R)`, where `T` and `R` are the reflectance and transmittance. If you have infinitely many such layers, then you have a thick layer with `Q_abs = Q_in * (1 - R - T) / (1 - T)`. We had been using the equation for an infinitely thick layer, which is not appropriate for an infinitesimally small piece of leaf area in a canopy.

Shouldn't it be integrated over the thickenss? I assume that is why the thick layer equation was being used. Although maybe that isn't exactly correct, using an infinitesimal layer thickness doesn't seem right either. I don't know if the integral for a single layer is something we can calculate though.

eloch216:
> > This is discussed at length in the vignette about thick layer absorption. Basically, for a thin layer, `Q_abs = Q_in * (1 - T - R)`, where `T` and `R` are the reflectance and transmittance. If you have infinitely many such layers, then you have a thick layer with `Q_abs = Q_in * (1 - R - T) / (1 - T)`. We had been using the equation for an infinitely thick layer, which is not appropriate for an infinitesimally small piece of leaf area in a canopy.
> 
> Shouldn't it be integrated over the thickenss? I assume that is why the thick layer equation was being used. Although maybe that isn't exactly correct, using an infinitesimal layer thickness doesn't seem right either. I don't know if the integral for a single layer is something we can calculate though.

My short answer is that I don't think the absorbed PPFD should be integrated over the layer thickness.

In my view, the canopy assimilation rate is defined as an integral. Then, a multilayer canopy model is essentially a numerical technique for actually evaluating the integral. I tried to explain this in the vignette (see screenshot below). Historically, I don't think BioCro has been very clear about what exactly it means by a "multilayer canopy model." I was hoping to lay down some rules in this vignette. If people disagree, we can change the vignette and try doing something else in the code.

image

I think Equation 2.1 is equivalent to the description of the multi-layer model in de Prury & Farquahar (1997): "Canopy  photosynthesis rate was calculated by the summation of the product of rate of leaf photosynthesis per unit leaf area by the leaf area in each class (Eqn A9)." (https://doi.org/10.1111/j.1365-3040.1997.00094.x) I don't know why they only wrote this in words instead of providing an equation.

justinmcgrath:
I'm not seeing how that means that PPFD shouldn't be integrated for the layer. A(l) implicitly has Q as a parameter, so it would be something like A(l, Q(l)). (The "l"s are a lowercase L, but "l" and "I" look the same in this font).

I've never learned much about numerical analysis, so I don't know how it would turn out, but I would think that the Reimann sum of A(I, Q(I)) would need account for Q across the layer as well, or does that work itself when summing A?

eloch216:
> I'm not seeing how that means that PPFD shouldn't be integrated for the layer. A(l) implicitly has Q as a parameter, so it would be something like A(l, Q(l)). (The "l"s are a lowercase L, but "l" and "I" look the same in this font).
> 
> I've never learned much about numerical analysis, so I don't know how it would turn out, but I would think that the Reimann sum of A(I, Q(I)) would need account for Q across the layer as well, or does that work itself when summing A?

The Riemann sum does account for Q across the layer, but in a very simple way. We are treating Q and the other inputs as if they were constant across each layer. Specifically, we assume that the value of Q and A at the layer's midpoint are representative of the layer as a whole. Graphically, we are replacing a smooth function with a set of rectangles. (See picture below, illustrating how it works for one of the leaf types.)

We _could_ come up with a more complicated way to choose a more representative point than the midpoint. But the added complexity would probably not be worth it, and we could instead increase accuracy by just using more layers in the approximation. The best way to increase accuracy would be to use Gaussian quadrature rather than a Riemann sum to approximate the integral.

In response to integrating Q across the layer, I would ask: What would we do with that value once we had it? We have a photosynthesis model that's defined for individual leaves, but not one that predicts A for layers of leaves. If we already had a way to get A from the total light absorbed by a layer of leaves, then we would already have our canopy model.

eloch216:
Is there an issue with Equation 1.2 (expressing canopy assimilation as an integral over leaf-level assimilation)? I was looking around for a source to cite for this equation (or Equation 1.1) but cannot seem to find one.

justinmcgrath:
That does look like a more straightforward way to handle it. I suspect that as N approaches infinity both methods would approach the integral though, so personally I have open questions that don't need to be addressed but maybe someone knows the answers. Does integrating the inputs converge, and does one method converge faster? There was the weird problem of absorbing more light than was incident on the canopy, so does one or both of these have the same sort of issues? Intuitively, it doesn't seem possible for the thin layer version to absorb more light than is incident.

eloch216:
> I suspect that as N approaches infinity both methods would approach the integral though

I'm not sure what are the two methods you're referring to. Maybe it would be easier to talk about this at the next BioCro meeting.

BTW, I am pretty puzzled by the regression test failure. Failures only occur for Miscanthus on Ubuntu, and specifically only for the `soil_evaporation_rate`. This quantity is calculated by the `soil_evaporation` module, which was not modified in this PR. My guess is this is due to roundoff errors, since I don't really understand how else it could happen.

I updated the `soil_evaporation` module test case to make it more realistic. If this passes on Ubuntu, then I will feel pretty sure there are just roundoff errors. In that case we can probably just ignore `soil_evaporation_rate` in the Miscanthus regression tests.

uphone87:
> The best way to increase accuracy would be to use Gaussian quadrature rather than a Riemann sum to approximate the integral.

Just curious, why would Gaussian quadrature be better than Riemann sum for this problem?

uphone87 (approved):
Thanks for revisiting this. I think it's interesting and necessary to explore the foundation of the integral of canopy photosynthesis.

vignettes/web_only/canopy_photosynthesis.Rmd
```diff
@@ -0,0 +1,638 @@
...
+notation is convenient because it allows us to define the net CO$_2$
+assimilation rate per unit ground area ($A_{n,canopy}$) using a one-dimensional
+integral over $\ell$:
+
+\begin{equation}
+    A_{n,canopy} = \int_{\ell = 0}^{L} A_{n,leaf}(\ell) d \ell,
+    (\#eq:AnCanopy)
+\end{equation}
+
+where $d \ell$ is the LAI of an infinitesimally small piece of leaf and
```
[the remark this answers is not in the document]
eloch216: It's tricky to explain what `$d \ell$` really means. How about this:

where $A_{n,leaf}(\ell)$ is the net CO$_2$ assimilation rate (per leaf area) of
an infinitesimally small piece of leaf located at cLAI $\ell$, and $d \ell$ is
the cLAI associated with that piece. ($d \ell$ can be thought of as $dS / S$,
where $dS$ is the area of the piece of leaf and $S$ is the total leaf area of
the canopy.)

eloch216:
> > The best way to increase accuracy would be to use Gaussian quadrature rather than a Riemann sum to approximate the integral.
> 
> Just curious, why would Gaussian quadrature be better than Riemann sum for this problem?

In my understanding, it is more efficient than a Riemann sum for integrating any function that can be approximated by a polynomial. I have not really gone into all the details. But I think it's a similar situation with solving sets of ODEs, where the Euler method is simple but less efficient than Runge-Kutta or adaptive Runge-Kutta methods. Likewise, a Riemann sum is simple but less efficient that Gaussian quadrature or adaptive quadrature. Besides this, there is a paper specifically about using Gaussian quadrature to calculate canopy assimilation rates: https://www.sciencedirect.com/science/article/abs/pii/0168192386900638

gsrohde:
> There is also an issue with a vignette: https://github.com/biocro/biocro/actions/runs/8972595097/job/24640910461. This kind of problem is difficult to detect because we set `error = TRUE` in most of our vignettes. This means that vignette errors are not picked up by `R CMD check`. (Also see [ebimodeling/biocro-dev#480](https://github.com/ebimodeling/biocro-dev/issues/480)). I am going to change the vignette error handling and then fix the issue with the Ball-Berry vignette. I think this will be helpful in the future.

I'm afraid I let issue [ebimodeling/biocro-dev#480](https://github.com/ebimodeling/biocro-dev/issues/480) fall by the wayside.  There was a logical reason, I'm sure, for setting `error = TRUE` in these vignettes.  I just don't remember what it was right now, and I don't know whether it is still relevant.  I'll try to see what I can come up with, and either elaborate on that issue or close it.  In the meantime, I guess it's OK to use the FALSE settings.

gsrohde (commented):
I know I'm a little late to the game here, but I couldn't not comment on some of the changes you made, especially since at least one of them was prompted by an earlier comment of mine.

eloch216:
> I know I'm a little late to the game here, but I couldn't not comment on some of the changes you made, especially since at least one of them was prompted by an earlier comment of mine.

I can make these changes in PR #109

PR #107, opened 2024-05-16
Separate par and nir

eloch216:
Some more changes to the canopy model. The major ones are:

  - Used the same absorptivity value for direct and diffuse light

  - Used separate leaf transmittance and reflectance values for PAR and NIR
    radiation within the canopy, rather than always assuming that light in the
    two bands are absorbed and scattered equally; in general, this caused a
    reduction in the absorbed shortwave energy for all leaves

  - Started calculating absorptivity as `1 - R - T`, where `R` and `T` are the
    leaf reflectance and transmittance coefficients, respectively. This ensures
    that the constraint `A + R + T` is always satisfied.


NEWS.md
```diff
@@ -51,6 +51,17 @@ be directly added to this file to describe the related changes.
...
+  - Used the same absorptivity value for direct and diffuse light
+
+  - Used separate leaf transmittance and reflectance values for PAR and NIR
+    radiation within the canopy, rather than always assuming that light in the
+    two bands are absorbed and scattered equally; in general, this caused a
+    reduction in the absorbed shortwave energy for all leaves
+
+  - Started calculating absorptivity as `1 - R - T`, where `R` and `T` are the
+    leaf reflectance and transmittance coefficients, respectively. This ensures
+    that the constraint `A + R + T` is always satisfied.
```
uphone87, line 63: do you mean 'A+R+T=1'?

data/soybean.R
uphone87, line 1: For soybean, I've been using 0.15 for reflectance + transmittance of PAR in the BioCro-ePhotosynthsis simulations. Sightly different from 0.13 you use here, but I think it's negligible.

NEWS.md
eloch216: Whoops, yes I do mean that!

data/soybean.R
eloch216: I used these values since I thought it would be more straightforward to use the numbers as they are written in the source cited. But I agree there won't be much of an impact on canopy photosynthesis due to this difference. I can change them to match your calculations if you want.

uphone87: Ah yes. Thanks. I remember the literature cited. I think I just rounded a bit for no particular reason 😄 . I wanted to have reluctance to take 0.1 and transmittance to be 0.05. However, these values definitely have some variations, as shown in some other sources. Would changing them to 0.1 and 0.5 harder to justify? I would still refer to the same source you cited. Otherwise, I will use 0.13 to be consistent with you in the future.

uphone87: BTW. Do you know how do they come up with a single number to represent the reflectance or transmittance of PAR? Often time, they are curves against bandwidth [ like Fig. 1 in here](https://acsess.onlinelibrary.wiley.com/doi/epdf/10.2134/agronj1971.00021962006300010038x). So do they need to integrate from 400-700nm to represent PAR?

eloch216: I think the single number should be determined from the absorption spectrum $\alpha(\lambda)$ and the spectral power density $M(\lambda)$ of the incident radiation as $\int_{400 nm}^{700 nm} \alpha(\lambda) \cdot M(\lambda) \cdot d \lambda$. I guess they're usually calculated assuming solar radiation. If you just integrate $\alpha(\lambda)$, that's equivalent to assuming white light ($M(\lambda) = 1$). Maybe people do this too.

Does https://doi.org/10.2134/agronj1971.00021962006300010038x give single values for PAR reflectance and transmission? Or are you estimating 0.1 and 0.05 from the graphs?

eloch216: I think we could cite another source if we want to use a rounded number, and just say it's approximate based on multiple sources. I can do that so the model here is the same as yours.

uphone87: > I think we could cite another source if we want to use a rounded number, and just say it's approximate based on multiple sources. I can do that so the model here is the same as yours.

Thanks. This would be good, saving me from re-running the ACi-AQ optimization for LD11.

> Does https://doi.org/10.2134/agronj1971.00021962006300010038x give single values for PAR reflectance and transmission? Or are you estimating 0.1 and 0.05 from the graphs?

They did not give single values. I rounded because I believe these values should have some variations. Also, Fig. 3 in [this paper](https://doi.org/10.2134/agronj1991.00021962008300030026x)) clearly shows that there are changes over time as soybean leaf grows. The average transmittance seems a bit over 0.05. We could cite these three papers.

PR #109, opened 2024-05-30
Update energy balance and transpiration

eloch216:
This is a pretty big PR that makes a few long-overdue changes to BioCro's canopy model. These changes are related to the "EvapoTrans" functions, which use energy balance considerations to calculate leaf temperatures and transpiration rates throughout the canopy. Below I will give an overview of the major issues that existed before, and how this PR addresses them.

1. In the past, we have had two separate functions for C3 and C4 leaves: `c3EvapoTrans` and `EvapoTrans2`. This doesn't make sense because energy balance calculations do not depend on the photosynthesis type. Rather, they are related to the absorbed light energy and the thermodynamic properties of air and water vapor. So, one major change here is to consolidate these into a single function, now called `leaf_energy_balance`.

2. A key part of these calculations is the estimation of boundary layer conductances. The `c3EvapoTrans` function used a model from Thornley & Johnson (1990), which represents conductance across the canopy boundary layer. The `EvapoTrans2` function used a model from Nikolov et al. (1995), which represents conductance across the leaf boundary layer. Since our canopy model considers transpiration at the leaf level, water vapor must actually pass across both of these boundary layers. So, the new `leaf_energy_balance` function now uses both of these models, and considers the overall boundary layer conductance as $1 /g_{b,total} = 1 / g_{b,leaf} + 1 / g_{b,canopy}$.

3. The leaf boundary layer conductance model from Nikolov et al. (1995) must be solved iteratively; this is mentioned very clearly in the text. Yet, BioCro's implementation of the model did not include an iterative loop, so its conductances were not being calculated properly. It also used the wrong value for one of the constants in the model. These mistakes have been fixed.

4. Both `c3EvapoTrans` and `EvapoTrans2` imposed limits on the temperature difference $\Delta T = T_{leaf} - T_{air}$ such that $\| \Delta T \|$ must be less than 5 or 10 degrees C. This seems overly restrictive, and I was hoping to remove this constraint entirely. However, I found during testing that the iterative loop in `leaf_energy_balance` still sometimes fails to converge, usually in situations where stomatal conductance is very small (and hence assimilation is small too). So I kept a constraint but set it to 30 degrees C. So, most of the time, it has no impact on the calculated leaf temperature and transpiration rate, and only kicks in to prevent truly outrageous values.

5. Our rationale for these calculations was not explained very well before, so I added a vignette about them. I also made a "wrapper" module for the `leaf_energy_balance` function called `BioCro:leaf_evapotranspiration`. This is useful for testing the code and illustrating the behavior of this function. It is used in an example at the end of the vignette.

There are also a few smaller changes:

6. I standardized the units of stomatal conductance so it is always given in `mol / m^2 / s`. There were some places where it had units of `mmol / m^2 / s` and this created some potential for mistakes if the conversions were not made properly. It's difficult to fully standardize the units of boundary layer conductance, since sometimes it is more convenient to express it as a "mass conductance" (units of `m / s`) or a "molecular conductance" (units of `mol / m^2 / s`). Nevertheless, these conductances have been standardized to some extent by making sure there are never any prefixes ahead of these units.

7. We assume that the absorbed longwave radiation from the sky can be calculated using the Stefan-Boltzmann law with emissivity of 1 and T = air temperature. Apparently this is not always true, and the emissivity in particular changes with cloud cover. I think it is fine to continue using this simple approach, but I've learned that some climate models (like DSSAT) can predict longwave radiation, and some weather stations can measure it. So I've made `absorbed_longwave` an input to the canopy and leaf-level models. Now it would be possible to take another approach besides the Stefan-Boltzmann law.

There is still room for improvement. For example, we could stop using the Penman-Monteith approach and directly solve the energy balance equation numerically. I tried doing this and it didn't work well with the fixed point method, so I think this will have to wait for the Newton-Raphson stuff to become available. Another issue is that we could probably improve the calculations of wind speed within the canopy. There is a section in the Campbell & Norman (1998) textbook about doing that. But, I don't have any intention of doing this stuff right now. The changes in this PR are already very helpful.

One way to see what's different is to compare the Ball-Berry vignette on this branch against the version on `main`. The `main` version is available online here: https://biocro.github.io/BioCro-documentation/latest/pkgdown/articles/web_only/ball_berry_model.html. The image below highlights a few key differences. One is that we are no longer hitting the arbitrary leaf temperature limit. Another is that the canopy profiles of leaf temperature and leaf surface humidity are much more smooth. Finally, the sunlit leaves are now consistently at a higher temperature than shaded leaves. These differences are the sum result of several key changes in this PR and two others (#106 and #107 ). Following these changes, it does seem that canopy assimilation rates have generally increased compared to `main`. Once this PR is approved and merged, we can re-parameterize the soybean model.

2024-05-30 B-B comparison


uphone87:
> There is still room for improvement. For example, we could stop using the Penman-Monteith approach and directly solve the energy balance equation numerically.

Why would this be an improvement? Isn't P-M model more widely used for ET estimation?

uphone87:
Okay. I think I understand your point. The iterative process for solving the P-M is not guaranteed to converge, which is similar to the c3photoC iteration. Using Newton-Raphson could potentially solve this issue. It would be an improvement.

src/module_library/leaf_energy_balance.cpp
```diff
@@ -0,0 +1,223 @@
...
+            leaf_temperature - air_temperature,
+            p_w_air,
+            gsw,
+            leaf_width,
+            wind_speed,
+            air_pressure);  // m / s
+
+        // Get the boundary layer conductance and total conductance to water
+        // vapor
+        gbw = 1.0 / (1.0 / gbw_leaf + 1.0 / gbw_canopy);  // m / s
```
uphone87, line 171: Why do the `gbw_leaf` and `gbw_canopy` need to be inverted for addition?

uphone87:
The middle panel shows that leaf temperature is lower in the top than the bottom. What condition was used for this simulation? In general I feel that leaf temperature with upper canopy leaves often being warmer than lower canopy leaves. This is because upper canopy leaves are exposed to more solar radiation and evaporative demand, and they may have limited stomatal conductance, which can reduce transpirational cooling. The lower ones are usually cooler due to shading.

justinmcgrath:
> So I kept a constraint but set it to 30 degrees C.

A difference of 10 degrees C would be a lot. If it's producing such unrealistic values, then presumably something is wrong with the model or the way it's solved. That seems worth making a note of to revisit later, if there's not a plan to address it here.

justinmcgrath:
The sunlit leaves are warmer at the bottom than at the top. The humidity at the leaf surface is pretty similar in all layers, so is this because of a smaller canopy boundary conductance in the lower layers?

src/module_library/leaf_energy_balance.cpp
justinmcgrath: This is analogous to resistances and conductances in series for electrical circuits. Resistances in series are summed. Since conductance is the inverse of resistance, the equivalent sort of calculations is the inverse of the sum of inverses.

`r_t = r_1 + r_2`. If `c_1 = 1 / r_1` and `c_2 = 1 / r_2`, then `r_t = 1 / c_1 + 1 / c_2`, and thus `c_t = 1 / r_t = 1 / ( 1 / c_1 + 1 / c_2)`.

justinmcgrath:
> The middle panel shows that leaf temperature is lower in the top than the bottom. What condition was used for this simulation? In general I feel that leaf temperature with upper canopy leaves often being warmer than lower canopy leaves. This is because upper canopy leaves are exposed to more solar radiation and evaporative demand, and they may have limited stomatal conductance, which can reduce transpirational cooling. The lower ones are usually cooler due to shading.

The shaded ones do have lower temperature, but I had a question about this as well (my last comment above). I had thought it was from a larger boundary layer conductance in the lower layer, but if that were the cause of this, I'd expect humidity at the leaf surface to increase in the shaded leaves, yet the opposite happens. My next guess would be longwave radiation from the layers above, but I don't think BioCro accounts for that. I recall that transpiration doesn't impact humidity, so maybe it's an artifact of that; that is, perhaps the higher temperature lower in the canopy is real, but because it doesn't account for water flux, the lower humidity at the leaf surface is wrong.

eloch216:
> > There is still room for improvement. For example, we could stop using the Penman-Monteith approach and directly solve the energy balance equation numerically.
> 
> Why would this be an improvement? Isn't P-M model more widely used for ET estimation?

The P-M is an approximation to the energy balance equation where the saturation water vapor pressure has been linearized, allowing an algebraic solution for leaf temperature or transpiration. This makes it very useful, especially in pedagogical contexts. But, one of the assumptions inherent in this approach to energy balance is that the stomatal and boundary layer conductances do not depend on leaf temperature. This assumption is violated in BioCro, so we are forced to take a numerical approach to finding leaf temperature anyway, which (in my opinion) removes the advantage of the P-M equation. Since we have to take an iterative approach regardless, I think it would ultimately make more sense to use the actual saturation water vapor pressure rather than the linearized version. These issues are explained with a bit more detail in the new "Energy Balance" vignette.

eloch216:
> The middle panel shows that leaf temperature is lower in the top than the bottom. What condition was used for this simulation? In general I feel that leaf temperature with upper canopy leaves often being warmer than lower canopy leaves. This is because upper canopy leaves are exposed to more solar radiation and evaporative demand, and they may have limited stomatal conductance, which can reduce transpirational cooling. The lower ones are usually cooler due to shading.

This is for a bright day -- the exact conditions can be found in the vignette. The fraction of sunlit and shaded leaves changes in each layer. In the top layer, there are essentially no shaded leaves, while in the lower layers, there are fewer and fewer sunlit leaves. I could add a weighted average temperature to the figure. It would be close to the sunlit temperature at the top, and become closer to the shaded temperature deeper into the canopy. I think this average temperature would agree with your expectations.

> The sunlit leaves are warmer at the bottom than at the top. The humidity at the leaf surface is pretty similar in all layers, so is this because of a smaller canopy boundary conductance in the lower layers?

In our model, wind speed decreases lower in the canopy. Leaf boundary layer conductance depends on wind speed, so it also decreases lower in the canopy. Smaller boundary layer conductance tends to increase leaf temperatures. I think this is why leaf temperatures tend to increase deeper in the canopy for both sunlit and shaded leaves.

If you build the Ball-Berry vignette, you can see two other figures showing stomatal conductance and boundary layer conductance through the canopy. Both decrease lower in the canopy, and shaded leaves have much smaller stomatal conductance than sunlit leaves.

> The shaded ones do have lower temperature, but I had a question about this as well (my last comment above). I had thought it was from a larger boundary layer conductance in the lower layer, but if that were the cause of this, I'd expect humidity at the leaf surface to increase in the shaded leaves, yet the opposite happens.

Relative humidity is tricky. The transpiration rate is driven by gradients in water vapor concentration, not relative humidity gradients. And humidity depends on temperature, so a leaf with more water vapor at its surface could still have a lower humidity if its temperature is higher. I could add some more plots showing the actual transpiration rates, and maybe humidity just outside the leaf boundary layer (which would still be within the canopy). My guess is that these outputs are all consistent with each other according to the equations of our models, but it's good to double check, especially if they seem counter-intuitive.

> I recall that transpiration doesn't impact humidity, so maybe it's an artifact of that; that is, perhaps the higher temperature lower in the canopy is real --possible if lower A results in lower gs -- but because it doesn't account for water flux, the lower humidity at the leaf surface is wrong.

I'm not totally sure what you mean by "transpiration doesn't impact humidity," but I could add another calculation as a consistency check: it's possible to calculate the transpiration rate as `g_s * (w_i - w_s)`, `g_b * (w_s - w_a)`, and `g_t * (w_i - w_a)`. We are at steady-state by assumptions so these should all give the same value. Here the `w` values are water vapor concentrations, either mole fractions (`mol / mol`) or densities (`kg / m^3`) depending on how the conductance is specified (either `mol / m^2 / s` or `m / s`).

justinmcgrath:
> I'm not totally sure what you mean by "transpiration doesn't impact humidity,"

As I recall, the humidity profile within the canopy is a function of humidity at the top of the canopy, given by the weather data, and layer within the canopy. Maybe there's something about wind speed too, I don't recall. But I don't think it's affected by transpiration rate. The profile is intended to look like the humidity distribution within a canopy, but it's not based on the transpiration rate within the canopy. Maybe I misremember or didn't understand it though.

A mechanistic model would have fluxes between the different layers and between each layer and the leaves within the layer. I can see how by forcing the humidity profile to match some shape, you could have temperature profiles that don't correspond to what you'd expect from humidity profiles. However, I'm not sure that is what is happening here, and maybe as you pointed out it's the difference between vapor pressure and relative humidity.

justinmcgrath:
In any case, I don't think this is something to worry about this in PR. I was just trying to understand it. This looks like an outright improvement regarding the design. I haven't had time to look at the changes in detail.

eloch216:
> > I'm not totally sure what you mean by "transpiration doesn't impact humidity,"
> 
> As I recall, the humidity profile within the canopy is a function of humidity at the top of the canopy, given by the weather data, and layer within the canopy. Maybe there's something about wind speed too, I don't recall. But I don't think it's affected by transpiration rate. The profile is intended to look like the humidity distribution within a canopy, but it's not based on the transpiration rate within the canopy. Maybe I misremember or didn't understand it though.
> 
> A mechanistic model would have fluxes between the different layers and between each layer and the leaves within the layer. I can see how by forcing the humidity profile to match some shape, you could have temperature profiles that don't correspond to what you'd expect from humidity profiles. However, I'm not sure that is what is happening here, and maybe as you pointed out it's the difference between vapor pressure and relative humidity.

Oh, I understand now. We stopped using that artificial humidity profile as part of this PR: https://github.com/ebimodeling/biocro-dev/pull/499

These surface humidity values are calculated using `w_s = w_i - E / g_sw` to determine the surface water vapor concentration, and then using the leaf temperature to find the corresponding humidity value. So they are based on the transpiration rate.

However, each leaf type (sunlit or shaded) and canopy layer are considered to have independent gas exchange with the bulk air. So the leaf-leaf interactions you mentioned are not included.

justinmcgrath:
> Oh, I understand now. We stopped using that artificial humidity profile as part of this PR: [ebimodeling/biocro-dev#499](https://github.com/ebimodeling/biocro-dev/pull/499)

I forgot that was changed, but I think the basic point still stands if I understand it. Rather than a profile where bulk air humidity is different in each layer, it's a profile where it is the same in each layer. In either case, transpiration isn't affecting humidity of the bulk air, just humidity at the leaf surface.

I assume that that results in drier air at the leaf surface in lower layers compared to a model where transpiration increases the bulk air humidity. However maybe there's something I'm missing something and this is actually accounting for the increased humidity in lower layers. Maybe differences in boundary layer conductance in the different layers accounts for that?

eloch216:
There is also an "interesting" difference in the canopy model convergence (with respect to the number of layers used) at different LAI. On this branch, the canopy assimilation rate doesn't change very much for `nlayer > 10`, but on `main`, assimilation still hasn't converged for `nlayer = 50` when `lai` is greater than 4. There also seems to be an overall increase in canopy assimilation rate here for the same conditions.

2024-06-04 nlayers

The code I used for this was:

```r
library(BioCro)
library(lattice)
library(RColorBrewer)

# Note: absorbed_longwave and gbw_canopy are only used on evapotrans branch
rc <- module_response_curve(
  'BioCro:c3_canopy',
  within(soybean$parameters, {
    StomataWS = 1.0
    cosine_zenith_angle = 1.0
    rh = 0.7
    temp = 30
    vmax = vmax1
    windspeed = 2
    solar = 1500
    lnb0 = 0
    lnb1 = 0
    absorbed_longwave = 5.67e-8 * 1 * (air_temperature + 273.15)^4
    gbw_canopy = 0.019 # estimated using evaluate_module('BioCro:canopy_gbw_thornley', list(canopy_height = 1, windspeed = 2, min_gbw_canopy = 0.005, windspeed_height = 5))
  }),
  expand.grid(
    lai = seq(from = 1, to = 10, by = 1),
    nlayers = seq(1, 50)
  )
)

cols <- brewer.pal(11, 'RdYlBu')[c(1:5,7:11)]

xyplot(
    canopy_assimilation_rate ~ nlayers,
    group = lai,
    data = rc,
    type = 'b',
    auto = TRUE,
    grid = TRUE,
    par.settings = list(
        superpose.line = list(col = cols),
        superpose.symbol = list(col = cols, pch = 16)
    )
)
```

justinmcgrath:
10 seems like a good number.

The two versions also converge at substantially different values. Things like this make me wish we had better means of validation.

The new version also appears to overshoot the converged value before coming back down, or possibly it oscillates while approaching the converged value (LAIs of 6, 7, 8 and 10). That seems interesting, but I don't know what it means. Maybe it's something similar to what happens with the fixed point approach when solving the PS equations.

justinmcgrath:
And just another thing I notice, but in the main branch leaves are usually cooler than the air, and on the new branch they're usually warmer. Again, I don't know which is right, or what this means really, but clearly these models disagree on some fundamental things.

eloch216:
> 10 seems like a good number.

I'll change it to 10 then.

> The new version also appears to overshoot the converged value before coming back down, or possibly it oscillates while approaching the converged value (LAIs of 6, 7, 8 and 10). That seems interesting, but I don't know what it means. Maybe it's something similar to what happens with the fixed point approach when solving the PS equations.

Agreed, it's interesting but I don't know what it means. I think it's different than what happens with the fixed-point iteration method, because the results at one value of `nlayers` don't have any influence on the result at `nlayers + 1`. The calculations at each value of `nlayers` are independent from each other.

> And just another thing I noticed, but in the main branch leaves are usually cooler than the air, and on the new branch they're usually warmer. Again, I don't know which is right, or what this means really, but clearly these models disagree on some fundamental things.

I remember trying to look into this before, and finding it surprisingly difficult to locate any measurements of the temperature differences between leaves and the surrounding air, or between sunlit and shaded leaves. I'm hoping we can use Carl's thermal camera this summer to at least get a few rough measurements that would allow us to answer broad questions like "which type of leaf has a higher temperature?" and "are leaves hotter than the air or vice-versa?".

If the new model is producing unrealistic results, my guess is that the issue is related to parameterization and not the model itself. I think the new version does a better job of following established model equations from the textbooks we use as sources.

> The two versions also converge at substantially different values. Things like this make me wish we had better means of validation.

I made an attempt at rough validation. I found two older papers that measured soybean canopy assimilation with clear-top chambers at several times throughout a growing season. I believe the measurements were made under bright conditions, so I'm treating them as maximal values of canopy assimilation rates on those days. My approach is to run the soybean model for one year on this branch and on `main`, and compare the largest predicted values of canopy assimilation to the ones measured in the papers. Of course this is complicated by the fact that nobody reports these things in the same units. I have converted everything to `micromol [CO2] / m^2 / s` to make it easier to compare them.

The figure below shows our predicted values:
 * On `main`, the maximum rates are around `0.03 - 0.04 Mg [glucose] / ha / hr`, which corresponds to `28 - 37 micromol [CO2] / m^2 / s`.
 * On `evapotrans`, the maximum rates are around `0.04 - 0.05 Mg [glucose] / ha / hr`, which corresponds to `37 - 46 micromol [CO2] / m^2 / s`.

[Wells et al. (1982)](https://onlinelibrary.wiley.com/doi/abs/10.2135/cropsci1982.0011183X002200040044x): The highest observed rates are reported as `4 - 6 g [CO2] / m^2 / hr`, which corresponds to `25 - 38 micromol [CO2] / m^2 / s`.

[Wells (1991)](https://onlinelibrary.wiley.com/doi/abs/10.2135/cropsci1991.0011183X003100030044x): The highest observed rates in Figure 2 are around `1.2 - 1.6 mg [CO2] / m^2 / s`, which corresponds to `27 - 36 micromol [CO2] / m^2 / s`.

From this it looks like canopy assimilation might be overestimated on the `evapotrans` branch. Alternatively, it could be a real difference in the cultivars, since these papers are quite old. I did find a paper from Carl that shows gross primary productivity (GPP) estimated from eddy covariance measurements: [Wu et al. (2023)](https://www.sciencedirect.com/science/article/pii/S016819232300223X). It's a bit difficult to interpret, but I think the middle column of Figure 9 shows that most values of soybean GPP are below `40 micromol / m^2 / s`. I'm not very familiar with this stuff so I don't know if these values are directly comparable to net canopy assimilation rates. If they are, then it seems the highest rates have been pretty stable for decades and it's probably not a cultivar issue.

As mentioned just above, I think the overestimation (if it exists) is probably due to a parameterization issue rather than an issue with the model itself.

2024-06-05 CanopyAssim

Here is the code I used for these figures:

```r
library(BioCro)
library(lattice)

    initial_values,
    parameters,
    soybean_weather[['2002']],
    direct_modules,
    differential_modules,
    ode_solver
)})

daily_max <- tapply(res$canopy_assimilation_rate, res$doy, max)
doy <- as.numeric(names(daily_max))

xyplot(
    daily_max ~ doy,
    type = 'b',
    pch = 16,
    ylab = 'Daily maximum canopy assimilation rate (Mg / ha / hr)',
)

```

And here are the conversion factors I used for the units:

```
Mg [glucose] / ha / hr:

1 Mg [glucose] / ha / hr * (1e6 g / Mg) * (1 mol [glucose] / 180 g [glucose]) * (6 mol [CO2] / mol [glucose]) * (1e6 micromol / mol) * (1e-4 ha / m^2) * (1 hr / 3600 s)
   = 1e6 * 6 * 1e6 * 1e-4 / (3600 * 180) micromol [CO2] / m^2 / s
   = 1e5 / 108 micromol [CO2] / m^2 / s
```

```
mg [CO2] / m^2 / s:

1 mg [CO2] / m^2 / s * (1e-3 g / mg) * (1 mol [CO2] / 44 g [CO2]) * (1e6 micromol / mol)
   = 1e-3 * 1e6 / 44 micromol [CO2] / m^2 / s
   = 1e3 / 44 micromol [CO2] / m^2 / s
```

```
g [CO2] / m^2 / hr:

1 g [CO2] / m^2 / hr * (1 mol [CO2] / 44 g [CO2]) * (1e6 micromol / mol) * (1 hr / 3600 s) micromol / m^2 / s
   = 1e6 / (44 * 3600) micromol / m^2 / s
   = 1e4 / 1584 micromol / m^2 / s
```

justinmcgrath:
> I'm hoping we can use Carl's thermal camera this summer

We also have a gun type IR thermometer (sort of like this https://www.fluke.com/en-us/product/temperature-measurement/ir-thermometers/fluke-568). It's not as cool, but it's very fast.

eloch216:
> [Wells et al. (1982)](https://onlinelibrary.wiley.com/doi/abs/10.2135/cropsci1982.0011183X002200040044x): The highest observed rates are reported as `4 - 6 g [CO2] / m^2 / hr`, which corresponds to `25 - 38 micromol [CO2] / m^2 / s`.

> [Wells (1991)](https://onlinelibrary.wiley.com/doi/abs/10.2135/cropsci1991.0011183X003100030044x): The highest observed rates in Figure 2 are around `1.2 - 1.6 mg [CO2] / m^2 / s`, which corresponds to `27 - 36 micromol [CO2] / m^2 / s`.

> On `evapotrans`, the maximum rates are around `0.04 - 0.05 Mg [glucose] / ha / hr`, which corresponds to `37 - 46 micromol [CO2] / m^2 / s`.

> From this it looks like canopy assimilation might be overestimated on the `evapotrans` branch. Alternatively, it could be a real difference in the cultivars, since these papers are quite old.

I realized I was using the default Soybean-BioCro value of atmospheric [CO2], which is 372.59 ppm. The papers I referenced above report measurements that were made in 1975 and 1988, when atmospheric [CO2] was 331.13 and 351.16 ppm, respectively. Using these lower values of `Catm` brings our predicted values down to around `0.035 - 0.045 Mg [glucose] / ha / hr`, which corresponds to `32 - 42 micromol [CO2] / m^2 / s`. Still a little bit higher than the reported values but not a huge difference, and perhaps this really could be due to cultivar differences.

eloch216:
Another point is that calculating the daily maximum might not be the correct comparison against the measured data. We can also look at the daily mean near midday (12 PM - 2 PM). These values are even closer to the measured ones.

I am starting to think that the new assimilation rates are not too high -- at least, they're not excessively high in an obvious way, which is probably the best we can get right now considering all the uncertainties in measuring canopy assimilation and how to best compare it to the model calculations.

image

The code I used:

```r
library(BioCro)
library(lattice)

    initial_values,
    within(parameters, {Catm = 331}),
    soybean_weather[['2002']],
    direct_modules,
    differential_modules,
    ode_solver
)})

daily_bright_mean <- as.numeric(by(res, res$doy, function(x) {
    mean(x[x$hour >= 12 & x$hour <= 14, 'canopy_assimilation_rate'])
}))

xyplot(
    daily_bright_mean ~ doy,
    type = 'b',
    pch = 16,
    ylab = 'Daily mean daytime canopy assimilation rate (Mg / ha / hr)',
)
```

justinmcgrath (approved):

justinmcgrath:
Sorry, I think at least 80% of the time I forget to actually submit the review.

eloch216:
I will ignore the test failures on R 3.6.0 for now and merge this.

PR #118, opened 2024-08-19
modify growth_respiration_fraction based on assim's sign

yufenghe22:
The growth respiration fraction is applied to reduce the canopy assimilation. Although it is set to be 0 at default, I'm proposing this minor fix for future use. 
When canopy assimilation is negative (e.g., respiration at night), this fraction would lower the respiration. 
Based on my understanding, I think when canopy assimilation is negative, the growth_respiration_fraction should be added instead of being subtracted.

### DIFF
--- a/src/module_library/multilayer_canopy_integrator.h
+++ b/src/module_library/multilayer_canopy_integrator.h
@@ -184,7 +184,11 @@ void multilayer_canopy_integrator::run() const
     // Note: this was originally only done for the C3 canopy
     // Note: it seems like this should not be necessary since the assimilation
     //   model includes respiration
-    canopy_assimilation_rate *= (1.0 - growth_respiration_fraction);
+    if(canopy_assimilation_rate>0){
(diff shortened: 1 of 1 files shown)


leavesandlemmas:
I think this is a good idea. It could also be implemented by using the absolute value of `canopy_assimilation_rate`.

For example, something like this pseudocode:
```
growth_respiration = growth_respiration_fraction * abs(canopy_assimilation_rate)
net_assimilation_rate = canopy_assimilation_rate  - growth_respiration
``` 
Or maybe using the sign function:
```
f  = growth_respiration_fraction * sign(canopy_assimilation_rate)
net_assimilation_rate = canopy_assimilation_rate * (1   - x )
```

eloch216:
I liked Scott's suggestion to use `abs`, rather than treating `An > 0` and `An <= 0` as separate cases. It seems a little more streamlined. I'll go ahead and merge this.

PR #119, opened 2024-08-21
Implementation of the von-Caemmerer model of C4 photosynthesis. 

leavesandlemmas:


src/module_library/von_caemmerer_c4_assimilation.cpp
```diff
@@ -0,0 +1,159 @@
+
+#include "von_caemmerer_c4_assimilation.h"
+
+double ElectronTransportModel::operator()(double light_photon_flux_density){
```
eloch216, line 4: In BioCro we try to avoid CamelCase identifiers: https://biocro.github.io/BioCro-documentation/latest/bookdown/contributing-to-biocro.html#formatting-code

eloch216:
Two other small comments for now:

1. Can you run `clang-format` on the three new files ("von_caemmerer_c4_assimilation.h", "von_caemmerer_c4_assimilation.cpp", and "von_caemmerer_c4_biocro.h")? 

2. The test case for `BioCro:von_caemmerer_c4_biocro` is now failing because the outputs don't match the stored results. I guess this must be related to the "bug fix" from one of the recent commits? If the change in outputs is expected, then this test case should be updated. This can probably be done using `update_csv_cases`.

eloch216:
The non-dimensionalization in the C++ code made it difficult to directly compare it with the model equations as written in Susanne's paper and textbook. So, I wrote a script that compares calculations from this module against ones from PhotoGEA. They seem to be in perfect agreement for the A-Ci and A-Q curves calculated in the script.

[aci_comparison_differences.pdf](https://github.com/user-attachments/files/16788187/aci_comparison_differences.pdf)
[aq_comparison_differences.pdf](https://github.com/user-attachments/files/16788189/aq_comparison_differences.pdf)

[compare_c4_biocro_photogea.txt](https://github.com/user-attachments/files/16788194/compare_c4_biocro_photogea.txt)

src/module_library/von_caemmerer_c4_assimilation.h
```diff
@@ -0,0 +1,142 @@
...
+#define VON_CAEMMERER_C4_H
+
+// standard library includes
+#include <cmath>
+#include <algorithm>
+#include <array>
+#include <limits>
+
+// includes from biocro
+#include "../framework/quadratic_root.h"
```
eloch216, line 12: It's good practice to only use `include` directives where they are required. In this case, `quadratic_root_minus` from `quadratic_root.h` is used in `von_caemmerer_c4_assimilation.cpp`, so the `include` statement should go in that file. Otherwise the `.cpp` file is relying on the `.h` file to include `quadratic_root.h` for it.

This particular case is not very troublesome, but it's possible to end up with complicated chains of `include` statements. Then if you remove an `include` from one from a file that no longer needs it, suddenly there are compilation errors with other unrelated files. So it's better to just put them only where they are needed.

src/module_library/von_caemmerer_c4_assimilation.cpp
```diff
@@ -0,0 +1,159 @@
...
+    double la = 1.0 - f1 * f5;  // dimensionless
+
+    // Equation 4.43 (here we use `lb` rather than `b` as in Equation 4.43)
+    double lb = -(Ajm + Ajbs + param.bundle_sheath_conductance * oxygen_mesophyll * f5 + param.co2_compensation_point_per_o2 * f1 * f6);  // micromol / m^2 / s
+
+    // Equation 4.45 (here we use `lc` rather than `c` as in Equation 4.44)
+    double lc = Ajm * Ajbs - param.co2_compensation_point_per_o2 * param.bundle_sheath_conductance * oxygen_mesophyll * f6;  // (micromol / m^2 / s)^2
+
+    // Equation 4.41 for the light-limited assimilation rate
+    return quadratic_root_minus(1, lb/la, lc/la);;  // micromol / m^2 / s
```
eloch216, line 104: There are two semicolons at the end of this line.

src/module_library/von_caemmerer_c4_assimilation.cpp
```diff
@@ -0,0 +1,159 @@
+
+#include "von_caemmerer_c4_assimilation.h"
+
+double ElectronTransportModel::operator()(double light_photon_flux_density){
+    double x = light_nondimensionalize(light_photon_flux_density);
+    if (std::isinf(x)){
```
eloch216, line 6: `std::isinf` is from the `<cmath>` library, so this file should have a `#include <numeric> // for std::isinf` at the top. This may be another case of an `include` in the header file that should be in the `.cpp` file. It would be good to double check all of the `includes` in the header file and move them to the `.cpp` file if they are only used in the `.cpp` file.

leavesandlemmas:
> The non-dimensionalization in the C++ code made it difficult to directly compare it with the model equations as written in Susanne's paper and textbook. So, I wrote a script that compares calculations from this module against ones from PhotoGEA. They seem to be in perfect agreement for the A-Ci and A-Q curves calculated in the script.

The first draft  didn't originally have the non-dimensionalization... it should have basically the same as PhotoGEA. However, when I was trying to find a bug ( which turned out to be a discrepancy in the value of Jmax and not a bug at all), I thought maybe there was an error in the algebra for getting the quadratic. I non-dimensionalized everything to ease rederiving the coefficients of the quadratic.  I ended up leaving it as is.

I feel like there ought to be a good way to separate the non-dimensionalization from solving the quadratic so that fewer parameters have to be pushed around, but I haven't thought of a better way yet.

src/module_library/von_caemmerer_c4_assimilation.cpp
leavesandlemmas: Why `#include <numeric>` and not `#include <cmath>`  if `std::isinf` is in `cmath`?

eloch216:
Your coding style is a lot different from mine and I'm not used to it yet, so it's a bit slow for me to go over your code. I think I will eventually get better at reading and following it.

I suspect that some of the class member functions could be replaced by free functions, following the advice here: https://www.aristeia.com/Papers/CUJ_Feb_2000.pdf. One of the suggestions is that if the function can be implemented using only the class's public interface, then it should be a free function rather than a member function. For example, the `ElectronTransportModel::atp_production_electron_transport_ratio` function only accesses public members of `ElectronTransportModel`, so it could be a free function instead. I have personally found this advice to be useful, and I think the code is a bit easier to follow when I use this rule.

Right now I have very mixed feelings about the non-dimensionalization. It complicates the process of comparing these equations to the reference publications, and I'm not quite sure what the benefit is. If you want to stick with this, I think it will need more documentation to explain how and why the normalizations have occurred. There are a few places where more informative variable names may also be helpful. E.g. it's not very clear what the `x0`, `x1`, `y0`, and `y1` variables represent.

src/module_library/von_caemmerer_c4_assimilation.h
```diff
@@ -0,0 +1,142 @@
...
+        double p(){return val[2];}
+
+};
+
+/**
+ * @brief A object for the Michaelis Menten maximum rate parameters:
+ * the maximum rubisco carboxylation rate (Vcmax), the maximum PEP carboxylation rate,
+ * the maximum PEP regeneration rate.
+ */
+class Vmax_type {
```
eloch216, line 41: I think `Vmax_type` and `K_type` could both be `struct` rather than `class`. Then you wouldn't need to write these trivial getter functions. E.g. `Vmax_type` would become 

```
struct Vmax_type {
    double rubisco_carboxylation;
    double pep_carboxylation;
    double pep_regeneration;
};
```

In BioCro, we usually use `struct` when we want to indicate that related parameters should be grouped together. There are plenty of examples, especially in `AuxBioCro.h`.

leavesandlemmas:
> Your coding style is a lot different from mine and I'm not used to it yet, so it's a bit slow for me to go over your code. I think I will eventually get better at reading and following it.
> 
> I suspect that some of the class member functions could be replaced by free functions, following the advice here: https://www.aristeia.com/Papers/CUJ_Feb_2000.pdf. One of the suggestions is that if the function can be implemented using only the class's public interface, then it should be a free function rather than a member function. For example, the `ElectronTransportModel::atp_production_electron_transport_ratio` function only accesses public members of `ElectronTransportModel`, so it could be a free function instead. I have personally found this advice to be useful, and I think the code is a bit easier to follow when I use this rule.

Maybe there's a better way to set things up. What I want to do is something like have a function that takes parameters as arguments, then returns a function that takes variables as arguments. SOmething like:

```python

def g(param):
    def func(x):
       return param[0] * x/(param[1] + x)
    return func

f(x, param) == g(param)(x)

  
```  
I saw an example of this in c++ using function objects, so that seemed to be the best way to get the same functionality. 

> Right now I have very mixed feelings about the non-dimensionalization. It complicates the process of comparing these equations to the reference publications, and I'm not quite sure what the benefit is. If you want to stick with this, I think it will need more documentation to explain how and why the normalizations have occurred. There are a few places where more informative variable names may also be helpful. E.g. it's not very clear what the `x0`, `x1`, `y0`, and `y1` variables represent.

Definitely the variables could have better names. I am not sure if keeping it is worth it.  In principle, some non-dimensionalization makes the model's code potentially less sensitive to using different units. But it does need better documentation.

src/module_library/von_caemmerer_c4_assimilation.cpp
eloch216: That was a typo 😂 If you look again you will see the comment has been edited.

eloch216:
> What I want to do is something like have a function that takes parameters as arguments, then returns a function that takes variables as arguments.

Just want to double check that I understand what you mean. Is the following code from `von_caemmerer_c4_biocro.h` an example of this?

```
    // bundle parameters into parameter objects
    Vmax_type V({Vcmax, Vpmax, Vpr});
    K_type K({Kc, Ko, Kp});
    VC_C4_Param_type param = {alpha_psii, gbs, ao, gamma_star, RL, Rm_frac};
    ElectronTransportModel electron_transport{Jmax, f_spectral, f_cyc, absorptance, 4.0, theta, x_etr};
    VC_C4_Model vc_c4_assim{electron_transport, K, param, V};
    // call vc_c4_assim for assimilation rates
    VC_C4_output out = vc_c4_assim.get_assimilation(Qin, Cm, POm);
```

So here `Qin`, `Cm`, and `POm` are "variables" and the other model inputs (`Vcmax`, `alpha_psii`, `Jmax`, etc) are "parameters?"

leavesandlemmas:
> So here `Qin`, `Cm`, and `POm` are "variables" and the other model inputs (`Vcmax`, `alpha_psii`, `Jmax`, etc) are "parameters?"

Yeah, that's the idea.  I mean I know that it might not really matter to the computer. I was trying to find ways to keep the number of arguments of each function smaller. And I think its harder to understand the model in code form if it's not clear what's variable and what's constant parameters. I read somewhere that function objects are useful because they are objects, so you can pass them to functions/methods as arguments (and return them), and because the function object can store state. Also, the Boost.Odeint integrator gives an example where you write the integrand as a function object so the parameters are data members and the operator() method computes the time-derivatives given state values.

 I was making some choices here in anticipation of solving the stomatal conductance and leaf balance. Basically any numerical method will require the biochemical assimilation to be a function of temperature and intercellular CO2, with everything else held constant.

leavesandlemmas:
> I see. So the "final" function (`VC_C4_Model::get_assimilation`) only has inputs that can potentially change during a loop when we're combining this model with the Ball-Berry model and/or energy balance.

Yes, or I think I was going to use the operator() method rather than this method. 
 

> As an alternative design choice, an assimilation function can be written without any distinction between "constants" and "variables," as was done for the FvCB model. Then when it is called in a loop, partial application could be used to clearly specify which inputs are held constant and which are varying during the loop. (We didn't do this in `c3photoC`, but it _could_ be done, and might be helpful.) In this approach, the loop writer is responsible for making their intentions clear, while the assimilation function is flexible and can be used with many potential loops. I would argue that this approach increases encanpsulation, because the writer of the assimilation function doesn't need to know how it will potentially be used in a loop.

What does encapsulation mean exactly?    

Also surely, there must be a way to define functions (in the mathematical sense)  which  allow flexibility in defining partial applications / currying but don't require an argument lists which are arbitrarily long. I mean one can always make any more complicated. IDK, mathematically a function ` f : X -> hom(Y,Z)`   , taking values in X and returning an element in the hom-set of Y and Y,  can be converted to a function on the cartesian product `g: X ×Y -> Z`  by basically setting `g(x,y) = f(x)(y)` so it would then be possible to have a partial application `h(x) = g(x,y) = f(x)(y) `  In c++ code, this is like 

```c++
template<typename T>
struct F { 
  T x;
  T operator()(T y){ return /* some expression involving x and y */; }
 };
 
template<typename T>
struct H { 
  T y;
  T operator()(T x){ 
    F<T> f(x); 
    return f(x)(y); 
    }
 };
```

I have a feeling this is not considered good c++ style however.

justinmcgrath:
> I suspect that some of the class member functions could be replaced by free functions, following the advice here

From a very broad perspective, there's no reason for member functions to exist at all. They're confusing, and therefore I only use them when limitations of the language require it or make it very cumbersome to avoid. We haven't come across any situation where a module needs member functions other than those defined in the Module class. Maybe something will require it some day, but without looking at this model in detail, I'm confident that this isn't a place they're needed.

src/module_library/von_caemmerer_c4_assimilation.h
justinmcgrath: state_map is the object type that most often should be used to pass things around. There's some instances of structs remaining  from previous code, but structs are inflexible and they aren't going to cooperate well with state_maps. Unless there's a specific reason to use a struct, I would use a state_map.

justinmcgrath:
> Maybe there's a better way to set things up. What I want to do is something like have a function that takes parameters as arguments, then returns a function that takes variables as arguments.

I think you want what C++ people call lambdas. You can also use function pointers but the syntax is hard to read. This is a place where classes could make sense. I haven't looked at this model in detail, but offhand I don't see how anything a Module does would need to work like this.

justinmcgrath:
> Also surely, there must be a way to define functions (in the mathematical sense) which allow flexibility in defining partial applications / currying but don't require an argument lists which are arbitrarily long.

This is why we use state_maps.

src/module_library/von_caemmerer_c4_assimilation.h
eloch216: > state_map is the object type that most often should be used to pass things around. There's some instances of structs remaining from previous code, but structs are inflexible and they aren't going to cooperate well with state_maps. Unless there's a specific reason to use a struct, I would use a state_map.

I will slightly disagree here, because there is still a pretty common scenario where a `struct` is appropriate (in my opinion). They are useful as return values of free functions that are used by multiple modules or other free functions. Details and an example are below.

It's not very straightforward to run a module from within a function, such as a free function or the `do_operation()` method of another module. So in cases where a model component needs to be used in multiple places in the C++ code, it's easier to write it as a free function.

(Side note: To run a module, the function would need to first define a state map with appropriate keys, then instantiate the module, run the module, and then retrieve the outputs from the state map. It can be done but it's a bit awkward. There is probably a way to improve this process but right now that's how it must be done. Sometimes it is worthwhile, as in the `multilayer_canopy_photosynthesis` template class, but not usually.)

An example of such a reusable code component is the FvCB model. We want it to be available on its own as a module (the `BioCro:FvCB` module), but we also need to use it in the coupled FvCB + Ball-Berry code (`c3photoC()`). So, the model is implemented via a free function called `FvCB_assim()`. The `BioCro:FvCB` module is a wrapper for this free function, and `c3photoC` also calls it.

It's typical that these free functions will need to return multiple numeric values. In that case a `struct` is used for the return value. This can also be seen in the FvCB example, where the return value of the `FvCB_assim()` function is a `struct` with values of net assimilation, Rubisco-limited assimilation, RuBP-regeneration-limited assimilation, etc.

A `state_map` would not be a good type for the input argument or return value of such a free function. If an input argument was a `state_map`, the free function would need to search through the map keys to find the values of specific elements; this is slow. If its return value was a `state_map`, then some other function would likely need to search through the map keys.

justinmcgrath: That sounds like a reasonable use. I didn't look too closely at this code and just assumed it was inheriting and extending a module.

leavesandlemmas:
I think this pull request should be closed and replaced with a rewrite.

justinmcgrath:
Whether to make things or build from the old one can go either way. Even for a rewrite, having the conversation that led there in one place can be nice. But at the same time, it's easy enough to include a link in the new place. It doesn't make much difference either way, so use whichever you prefer.

leavesandlemmas:
> Just be careful when you're deciding which model inputs are variables or constants. Much of this is situation-dependent. E.g. if I am calculating a CO2 response curve, then the incident light is constant. But it would be variable in a light response curve. Another one is that Vcmax depends on the Rubisco activation state, so it may be variable (changing with light and CO2) in a more complex approach to modeling photosynthesis.

> Because of this, these divisions can be arbitrary and add unnecessary complexity to computer code. In BioCro, it is very intentional that modules just have "inputs" and "outputs," without any other distinctions between types of inputs such as constant vs. variable. This is discussed in the paper.

What about having higher-order modules, modules with some inputs which are also modules?

PR #120, opened 2024-08-22
Modify maintenance respiration and re-optimize the Soybean-BioCro

yufenghe22:
This PR re-optimized the Soybean-BioCro with the 2002-2006 datasets by making the following changes:

1. Renamed the `mrc1` and `mrc2` to `grc_stem` and `grc_root`. The "mrc" refers to maintenance respiration coefficient. However, they are used to scale the assimilation. From the following literatures, I learned that they should be called growth respiration. Therefore, the renaming was made to reflect this definition.
 - Apsim: (https://apsimdev.apsim.info/ApsimX/Documents/AgPastureScience.pdf)
 - Thornley, J. H. M. "Growth, maintenance and respiration: a re-interpretation." Annals of Botany 41.6 (1977): 1191-1203.).

2. The real maintenance respiration is often modelled by a fraction of dry biomasses. Therefore, I added a new module called  `maintenance_respiration`. It removes a fraction of an organ by a constant parameter called `mrc_*` (e.g., `mrc_leaf`) and also by a Q10-based temperature response. 

3. A comparison between the model and observation with the newly optimized parameters can be found [here](https://github.com/Matthews-Research-Group/soil-tipping-bucket/tree/main/run/optimization/figs). The changed parameters include partitioning coefficients and growth and maintenance respiration coefficients. The script to repeat it can be found [here](https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/main/run/optimization/plot_compare_with_obs.R). To repeat, you'll also need the Data files and some functions in the **optimization** folder.

@leighmatth


leavesandlemmas:
The changes look good to me. Separating maintenance respiration and growth respiration means that respiration can be a significant, even when assimilation is zero. 

One future goal I have is to collect the respiration terms into a single whole-plant respiration rate so that the model produces an estimate of net carbon exchange; then the model could be compared against eddy covariance flux data for soybean fields.

leavesandlemmas (approved):
Seems fine

src/module_library/maintenance_respiration.h
```diff
@@ -0,0 +1,128 @@
...
+    double* Root_op;
+    double* Rhizome_op;
+    double* Grain_op;
+    double* Shell_op;
+
+    // Implement the pure virtual function do_operation():
+    void do_operation() const override final;
+};
+
+string_vector maintenance_respiration::get_inputs()
```
eloch216, line 80: The `get_inputs()` function is missing some of the module inputs: `"mrc_leaf"`, `"mrc_stem"`, `"mrc_root"`, and `"mrc_grain"`.

Because of this, the module is not fully functional. For example, typing `module_info('BioCro:maintenance_respiration')` produces this error:

```
Error: could not create the module
Additional details:
Thrown by get_input: the quantity 'mrc_leaf' was not defined in the state_map.
```

I found this issue by running the package tests. In general, it is a good idea to make sure all tests pass before submitting a PR. There are several ways to run the tests. I usually do it from R via `test_dir('biocro/tests/testthat')` (since I am usually working one directory above the `biocro` source directory). Other options are described here: https://biocro.github.io/BioCro-documentation/latest/bookdown/running-the-testthat-tests.html

src/module_library/BioCro.h
```diff
@@ -5,6 +5,7 @@
 #include "AuxBioCro.h"
 
 double resp(double comp, double mrc, double temp);
+double temperature_response(double temp);
```
eloch216, line 8: It's confusing when the base file names of the `.h` and `.cpp` files associated with a function are not the same, and I've been trying to standardize these things over time. The `temperature_response` function is implemented in `AuxBioCro.cpp`, so it should be defined in `AuxBioCro.h`, not `BioCro.h`.

Related to this, the function name may be too generic. There are many potential temperature response functions. I would prefer something like `Q10_temperature_response` instead.

src/module_library/AuxBioCro.cpp
```diff
@@ -732,16 +732,41 @@ soilML_str soilML(
...
+ *  respiration instead of maintenance respiration, as defined in these papers:
+ *  - Apsim: (https://apsimdev.apsim.info/ApsimX/Documents/AgPastureScience.pdf)
+ *  - Thornley, J. H. M. "Growth, maintenance and respiration: a re-interpretation."
+ *    Annals of Botany 41.6 (1977): 1191-1203.
  */
-double resp(double base_rate, double mrc, double temp)
+double resp(double base_rate, double grc, double temp)
 {
-    double ans = base_rate * (1 - (mrc * pow(2, (temp / 10.0))));
+    double ans = base_rate * (1 - (grc * pow(2, (temp / 10.0))));
```
eloch216, line 744: Since we now have function for Q10 temperature response, it would be nice to use it here. There are probably a few other Q10s throughout BioCro. We don't have to standardize them all right now, but this one would be a good start.

eloch216:
This looks good overall. I think it will be helpful to distinguish between growth and maintenance respiration. Can you add a note about these changes to `NEWS.md`?

Similar to Scott O, I think it would be very useful if we can separately keep track of how much carbon is lost to respiratory processes (including photorespiration), and how much gross carbon is assimilated. These can be added in another PR, though.

It looks like something is wrong with our `R CMD check` GitHub action, because all the operating systems and versions failed with an error such as:

```
Error: Bad request - quarto-dev/quarto-actions/setup@577558713f28cb3a0e7a735f80dce913d89b77d2 is not allowed to be used in biocro/biocro. Actions in this workflow must be: within a repository owned by biocro, created by GitHub, or matching the following: JoshuaTheMiller/conditional-build-matrix@81b51eb8d89e07b86404934b5fecde1cea1163a5, r-lib/actions/*@v2.
```

Hopefully there is an easy fix for this. I can look into it. In the meantime, I manually ran `R CMD check` on my computer and discovered a few issues that need to be addressed.

1. The `BioCro:maintenance_respiration` module needs a test case file in the `tests/module_test_cases` directory. This can be created with the `initialize_csv` function.

2. The test case file for the `BioCro:no_leaf_resp_neg_assim_partitioning_growth_calculator` needs to be updated to reflect the new names of its input quantities.

3. The saved soybean simulation output needs to be updated so the soybean regression test passes. There are instructions for doing this in `tests/testthat/crop_model_testing_helper_functions.R`.

src/module_library/maintenance_respiration.h
```diff
@@ -0,0 +1,128 @@
+#ifndef MAINTENANCE_RESPIRATION_H
```
eloch216, line 1: Can you run `clang-format` on this file? Some of the formatting is not standard (mostly related to blank spaces)

yufenghe22:
> The other is that the `Grain` mass now has a slight decrease at the end of the year, I guess due to maintenance respiration. I think this is probably reasonable.

Yes. This can be controlled solely by the `mrc_grain`. The current `mrc_grain` is 1e-4, close to no reduction. I can change it to 1e-5 to make sure there is no decrease at all at the end.

src/module_library/maintenance_respiration.h
yufenghe22: Thanks for bring this up! It slipped my mind. At some point, the testing was in my routine of making changes.

yufenghe22:
> One is that the new `Stem` mass has a sharp peak that was not present before. I think the old result may have been more realistic, but I don't think we need to worry about this.

Is this year 2002? 2005 and 2006 are less sharper.

src/module_library/BioCro.h
leavesandlemmas: The Q10 temperature response .

Here's my attempt for a more reusable function. I think the exponential version would be more stable for optimization, since apparently  for floats `pow(x,y)` calculates  `exp(y *  log(x))`  and this way, it isn't have to do `log(x)`. 

```c++
double temperature_response_exponential(
  double temperature, double reference_temperature, double q10_coef
 ){
    double temp_diff = temperature -reference_temperature;
    double u = q10_coef*temp_diff; // q10_coef = log(2)/10
    return exp(  u);
}
```

src/module_library/AuxBioCro.cpp
leavesandlemmas, line 744: Here's my rewrite using the `temperature_response_exponential` function. IDK if this is a good/correct use of `constexpr double variable = ...` the value of `log(2)/10` would only be needed at compile time here.

```c++
double resp(double base_rate, double grc, double temp){
  constexpr double coef = log(2)/10; 
  double gr = temperature_response_exponential(temp, 0, coef); 
  return base_rate *  ( 1- gr );
} 
```

yufenghe22: The Q10 function normalizes `temp` by a `Tref` (usually 25 C). This function does not do that. This was holding me from changing it to a Q10 function.

yufenghe22: I guess I can make the `Tref` an input to the Q10 function. In this case, it will be 0. This does not need to re-optimize the `grc` parameters.

leavesandlemmas: It isn't necessary to reoptimize even if we change `Tref`.  We can calculate what parameter value is equivalent to the original at a different temperature: 

```r
grc_25C = grc_0C * 2^(25/10) 
```

yufenghe22: Yes. This makes sense. I will however just add the `Tref` as an input for readability.

eloch216:
Thanks, the new changes look good to me!

> > One is that the new `Stem` mass has a sharp peak that was not present before. I think the old result may have been more realistic, but I don't think we need to worry about this.
> 
> Is this year 2002? 2005 and 2006 are less sharper.

Yes, it was year 2002. I tried running the model for 2005 and 2006 and the stem mass still has a peak, though. It's not a big deal anyway.

eloch216 (approved):

src/module_library/BioCro.h
eloch216: > I think the exponential version would be more stable for optimization, since apparently for floats `pow(x,y)` calculates `exp(y * log(x))` and this way, it isn't have to do `log(x)`.

I usually don't worry about this kind of stuff. The performance difference is likely so small it's not worth spending effort on it.

PR #122, opened 2024-08-28
Change r cmd check

eloch216:
There have been some recent errors with the `R CMD check` workflow such as the following:

```
Error: Bad request - quarto-dev/quarto-actions/setup@577558713f28cb3a0e7a735f80dce913d89b77d2 is not allowed to be used in biocro/biocro. Actions in this workflow must be: within a repository owned by biocro, created by GitHub, or matching the following: JoshuaTheMiller/conditional-build-matrix@81b51eb8d89e07b86404934b5fecde1cea1163a5, r-lib/actions/*@v2.
```

I'm not quite sure what to make of it, but I looked for examples using `JoshuaTheMiller/conditional-build-matrix` and noticed a few small differences compared to our code. This is just a shot in the dark to see if it fixes the problem.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -70,7 +70,7 @@ jobs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v4.1.1
+        uses: actions/checkout@v4
         with:
           submodules: true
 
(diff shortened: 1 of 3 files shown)


eloch216:
Well that seems to have only made things worse!

eloch216:
Update: I fixed the issue by changing a setting in this repository. I went to Settings -> Actions and added `quarto-dev/quarto-actions/setup@577558713f28cb3a0e7a735f80dce913d89b77d2` to the allowed actions. We didn't need this before, but I guess maybe one of the other actions started using `quarto-dev`. Leaving this note in case something similar happens again in the future.

PR #125, opened 2024-09-27
Cowpea stuff

eloch216:
This PR is for a few changes I made while working on Cowpea-BioCro. The most important changes are:

- I wanted to use different values for a few of the hard-coded temperature response parameters, so I un-hard-coded them. Here I'm referring to things like Arrhenius activation energies. Because of this change, there are now more parameters in each C3 crop model, but the model behavior has not changed. (Note: no changes were made to the stored test data for soybean or willow, and the tests still pass.)

- I wanted to use a bilinear thermal time model and the logistic partitioning growth model. In order to do this, I needed to add a new module for calculating development index from thermal time: `BioCro:development_index_from_thermal_time`.

- I thought I needed to use a logistic equation for calculating the specific leaf area from thermal time. So, I split out the SLA calculations from the `BioCro:parameter_calculator` module,  and made two modules for calculating SLA: `BioCro:sla_linear` is the same linear model we have been using, while `BioCro:sla_logistic` is a new logistic option. Ultimately I did not actually need the logistic version for Cowpea BioCro, but I still think this is a useful change.

I also made some organizational changes, and added a module for calculating temperature-dependent values of C3 photosynthetic parameters. A more complete list of changes can be found in `NEWS.md`.

If you look through the diff, you will also see some changes to lists of function input arguments or module inputs/outputs. These are mostly related to alphabetizing; I am slowly alphabetizing these lists when I encounter them because I think it makes things more organized. Finally, there are a few small changes to fix `Doxygen` warning messages. These are unrelated to anything else in the PR but I figured I should fix them since I noticed them.


src/module_library/c3_parameters.h
yufenghe22, line 1: This looks much better instead of hard-coding them!

yufenghe22 (approved):
It doesn't seem that any of these are directly related to `cowpea-stuff`.  They make `all-crop-stuff` look cleaner.

eloch216:
I'll merge this during the afternoon of Friday, October 18 if I don't hear anything else by then.

PR #127, opened 2024-11-24
Fix weather data

yufenghe22:
This PR fixes two issues in the cmi weather data:

1. The year column was always shown as `2000` due to incorrect round-up.
2. The year 2022 and 2023 were identical.

Note: the cmi weather data on the `main` branch also needs this fix.


yufenghe22:
Running `testthat.R` on my laptop came back with no fails. Any idea why the github checks failed?

justinmcgrath:
The changes look good to me.

It's failing due to some issue with documentation building. Although, I do like the automated workflows, it would be nice if we didn't rely on random people's software. Building the documentation isn't as critical as having BioCro function correctly, but fixing broken dependencies is always frustrating.

justinmcgrath:
I added quarto-dev/quarto-actions/setup@87b35bb88b36317fa36b5189e9553b4164a5c5a3 to the list of allowed actions, which I think should fix the errors, and I started the tests again.

yufenghe22:
I think we had a similar issue [last time](https://github.com/biocro/biocro/pull/120#issuecomment-2315721490). It was resolved in that PR, but I don't know what/who the fix was. Maybe @eloch216 fixed it.

justinmcgrath:
The biocro repository is set to allow only specific actions. Somewhere we have a workflow that includes that one, but it hasn't been added to the list of allowed actions. We don't call it directly, so something we're calling is getting changed, and then that requires us to update our repository. I added that to the list of allowed actions here https://github.com/biocro/biocro/settings/actions. This isn't a big deal, but it would be nice to find a way to specify a specific version so that we don't need to do this, or to find a way that doesn't require this.

I accidentally only ran one test last time. I think I've got them all running now. The one I ran passed, so I'm hopeful it will all work.

yufenghe22:
Thanks! All passed now. I'll merge if there's no further comments by the end of the day.
(BTW, I don’t have permissions to view the BioCro repository’s settings. That’s fine tho—I’m okay staying out of the Github's technical details).

eloch216:
I am now realizing that the `soybean_weather` data is not fully described. I think it was originally created by calling this script for 2002, 2004, 2005, and 2006: https://github.com/cropsinsilico/Soybean-BioCro/blob/main/WeatherDataProcessing/get_weather_files.R

Then it was apparently rounded, but not with `inst/extdata/rounding_weather_values.R`, because that script only modifies the `weather` data.

I can work on getting the new soybean weather data so the new test will pass.

eloch216:
I added a new script for processing the raw soybean weather data, updated the documentation for the soybean weather data, and updated the saved soybean regression test data. I think all the tests should pass now.

eloch216 (approved):

justinmcgrath (approved):

eloch216:
BTW - I think this is my first ever successful use of "test driven development." I made a test, it failed, and I changed the code to make it pass.

justinmcgrath:
> BTW - I think this is my first ever successful use of "test driven development." I made a test, it failed, and I changed the code to make it pass.

Yeah, it turned out that "to make sure this doesn't happen again" was really "to make sure this hasn't happened before".

PR #130, opened 2024-12-10
Module writer - EBL changes

eloch216:
Here I made some changes to the module writer, mostly to standardize the code style to match other BioCro R functions. The biggest changes are:
- The header files now follow clang-format guidelines as much as possible. (The main exception is that unit comments in `get_inputs` and `get_outputs` will not be vertically aligned.)
- There are some more tests to ensure the header format hasn't changed.

PR #131, opened 2024-12-10
Replace `tensor_string_vector` by an example

eloch216:
Scott O. pointed out that the `tensor_string_vector` can be reproduced almost entirely by a call to `kronecker` (https://github.com/biocro/biocro/pull/130#discussion_r1878654532).

I am thinking that we could probably just include an example of how to do this, rather than have a function in the namespace for doing it.


justinmcgrath:
I don't think it makes sense to have a swap argument to this function. One would just swap the arguments. So removing it makes sense to me. It is also the same as `as.vector(outer(A,B, paste, '_'))`. It doesn't really matter what gets used in the example, although I assume more people are familiar with an outer product than a kronecker product.

PR #133, opened 2024-12-13
Use upload-artifact v4.4.3

eloch216:
This PR attempts to address issue https://github.com/biocro/biocro/issues/132

We only use `upload-artifact` in one spot (in the documentation workflow). The workflow had been using version 3.1.3 of the action, so I changed it to the latest version (4.4.3).

I also followed the instructions in the "Multiple uploads to the same named Artifact" section here: https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md. I think that section seems to apply for our situation.

### DIFF
--- a/.github/actions/build_pkgdown_site/action.yml
+++ b/.github/actions/build_pkgdown_site/action.yml
@@ -14,6 +14,12 @@ runs:
     - name: Install Pandoc
       uses: r-lib/actions/setup-pandoc@v2
 
+    - name: Install R
+      uses: r-lib/actions/setup-r@v2
+      with:
+        r-version: "release"
+        use-public-rspm: true
+
     - name: Install TinyTex
       run: |
         install.packages('tinytex')
--- a/.github/actions/render_bookdown_book/action.yml
+++ b/.github/actions/render_bookdown_book/action.yml
@@ -17,12 +17,18 @@ runs:
       if: steps.pandoc-test.outcome == 'failure'
       uses: r-lib/actions/setup-pandoc@v2
 
-    - name: 3. Install bookdown
+    - name: 3. Install R
+      uses: r-lib/actions/setup-r@v2
+      with:
+        r-version: "release"
+        use-public-rspm: true
+
+    - name: 4. Install bookdown
       run: |
         install.packages(c("bookdown"))
       shell: Rscript {0}
 
-    - name: 4. Run bookdown
+    - name: 5. Run bookdown
       working-directory: ${{ inputs.package_root }}/documentation/script
       run: |
         Rscript run_bookdown.R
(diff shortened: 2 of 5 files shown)


eloch216:
There seems to be something wrong with the `R CMD check` workflow today. It keeps failing with a weird message, which occurs before `R CMD check` even runs: `Error: Rscript: command not found`. I'm also seeing this error on another PR: https://github.com/biocro/biocro/pull/131.

justinmcgrath:
I suspect the failure is from the test machines changing to Ubuntu 24.04. In some places it specifies 20.04, but for R CMD check it uses ubuntu-latest. My guess is it's trying to install R 3.6 and not finding it.

eloch216:
> I suspect the failure is from the test machines changing to Ubuntu 24.04. In some places it specifies 20.04, but for R CMD check it uses ubuntu-latest. My guess is it's trying to install R 3.6 and not finding it.

Good point, I think it is related to the change to Ubuntu 24.04. I found this page with some info about the change: https://github.com/actions/runner-images/issues/10636

Specifically, it looks like R is not included in the Ubuntu 24.04 image.

The code that keeps causing an error can be found here: https://github.com/biocro/biocro/blob/31a414542d64513ba0c5425ee2b72fe651670fe5/.github/workflows/R-CMD-check-backend.yaml#L75-L134

This job contains R code that is being run on Ubuntu, but now R is not installed on the latest version of Ubuntu, so the code fails.

I am sure other people are encountering this too, since lots of people use R code in these workflow definitions.

justinmcgrath:
Ah, so this isn't a problem with an old version of R, but an issues of not having R at all? Previously I thought the script installed R when it ran.

eloch216:
> Ah, so this isn't a problem with an old version of R, but an issues of not having R at all? Previously I thought the script installed R when it ran.

There are quite a lot of "scripts" involved in the workflows. When we perform an action like "Check the package on operating system X using R version Y," then R is explicitly installed so we can use the correct version.

This is a different part of the workflow, where the arguments to `R CMD check` are assembled. This is done using R code, but it is a separate step from the part I mentioned above. It's all very basic stuff, like concatenating strings, so we never specified or installed a version of R before. But now R is not included in the Ubuntu image, so the code fails.

I have a guess for how to fix it. I will give it a try. It will also be nice to hear Scott R's feedback on it.

eloch216:
@gsrohde Could you look over these workflow changes? I think I have addressed the errors and warning messages, but I am not as familiar with the workflows as you are.

eloch216:
I will go ahead and merge this since it seems to solve the issues with `R CMD check`. It will be nice to have this in `develop` so we can use it in current / future PRs, where checks are failing because of the ubuntu image issue.

PR #139, opened 2025-01-06
Reversion to a simple time check. 

leavesandlemmas:


eloch216:
It looks like the `time-check-simple` branch builds on the `time-is-consistent-v2` branch, so it also includes the changes to the `time` units.

I think the changes in the "initial commit" commit should be added to a branch based on `develop`, and then the PR should request to merge into `develop`. Then the PR would only include the new check that `time` is ordered and evenly-spaced.

leavesandlemmas:
> It looks like the `time-check-simple` branch builds on the `time-is-consistent-v2` branch, so it also includes the changes to the `time` units.
> 
> I think the changes in the "initial commit" commit should be added to a branch based on `develop`, and then the PR should request to merge into `develop`. Then the PR would only include the new check that `time` is ordered and evenly-spaced.

I don't know why it didn't copy from the develop. I thought i had started a new branch from develop

eloch216:
I see that you changed the target of this PR to `develop`, but unfortunately that doesn't fully solve the issue because the `time-check-simple` branch also includes all the commits on the `time-is-consistent-v2` branch.

Issue #129, opened 2024-12-10
`run_biocro` behaves strangely when provided repeated time points. 


leavesandlemmas:
Working with Justin yesterday, we discovered that `run_biocro` will happily accept a drivers dataframe where the times in the `time` variable are not unique (or sorted). It seems the integration method implicitly assumes the drivers will only provide a sorted sequence of unique times.  So either this needs to be checked somewhere inside `run_biocro` and then raise an exception (or correct the time driver); or the behavior should change so that `run_biocro` returns the solution at the requested times, even when those times are repeated or out of order. 

The code below reproduces the error for the harmonic oscillator.  Using different time drivers, `run_biocro`  always returns the correct solution for `times_sequential` where the times are sorted, unqiue, and their timestep matches the timestep of the `timestep` parameter.  The true solution 

```r
library(BioCro)

times_repeated = c(0, rep(1:5, each = 3)) # repeats times 
times_sequential = seq(0, 10, 1) # gives correct behavior
times_permuted = sample(times_sequential, length(times_sequential))  # applies random permutation to times
times_scaled = times_sequential * 10  # rescales time


# simple biocro model
harmonic_model = list(
  initial_values = list(position = 1, velocity = 0),
  parameters = list(spring_constant = 1, mass = 1, timestep=1),
  drivers = data.frame(time = times_repeated),
  direct_module_names = list(),
  differential_module_names = list("BioCro:harmonic_oscillator"),
  ode_solver = soybean$ode_solver
)

result = do.call(run_biocro, harmonic_model)

# true solution 
harmonic_sol <- function(t, x0, y0, mass, spring_constant){
  freq = sqrt(spring_constant/mass)
  z0 = complex(real = x0, imaginary =  y0)
  z = exp(-1i * freq *  t) * z0
  z
}

# compute true solution
z = harmonic_sol(
  harmonic_model$drivers$time,
  harmonic_model$initial_values$position,
  harmonic_model$initial_values$velocity,
  harmonic_model$parameters$mass,
  harmonic_model$parameters$spring_constant
  )
result$position_true <- Re(z)
result$velocity_true <- Im(z)

# compute solution with times as integer multiples of timestep parameter
z = harmonic_sol(
  seq(0,length(harmonic_model$drivers$time) - 1,1) * harmonic_model$parameters$timestep,
  harmonic_model$initial_values$position,
  harmonic_model$initial_values$velocity,
  harmonic_model$parameters$mass,
  harmonic_model$parameters$spring_constant
[...]

PR #140, opened 2025-01-07
Check that time in drivers is evenly spaced.

leavesandlemmas:
As per #137  this pull request only creates a check for time to be evenly spaced, removing most of the other changes of earlier pull request. This check doesn't use `timestep` and it doesn't check if time is spaced by the `timestep`.


eloch216:
It looks like there are some test failures. I haven't fully investigated, but I think they are related to `validate_dynamical_system_inputs`. Like `run_biocro`, it also calls `add_time_to_weather_data`. Now this needs to be done before calling `check_run_biocro_inputs`.

BTW, I have found the following workflow to be helpful when making changes to BioCro:
1. Build the package and run the tests locally. The tests don't take much time to run so this is a good way to get fast feedback.
2. When all the tests are passing, run `R CMD check` locally. This takes longer than just running the tests, but it will also check the examples and vignettes, which sometimes turns up more issues.
3. When `R CMD check` is passing locally, make a PR, but don't assign any reviewers. Making the PR will cause the workflow tests to run, which includes `R CMD check` on several operating systems. Sometimes an issue will appear that didn't occur locally on your own operating system / R version.
4. When the online checks have passed, then assign reviewers to begin the review process.

Doing this will ensure that there are no basic issues when reviewers begin to look at the code. In the long run, it saves time.

gsrohde (commented):
A few comments.

R/input_checking_functions.R
```diff
@@ -272,3 +272,32 @@ check_boolean <- function(args_to_check) {
...
+    time_is_null <- is.null(drivers[['time']])
+    if(time_is_null){
+        error_message <- "No `time` variable found in the `drivers` dataframe."
+        return(error_message)
+    }
+
+    time <- drivers[['time']]
+    error_message <- character()
+    if (length(time) < 3){
+        # automatic pass because 3+ rows are needed to check the spacing.
```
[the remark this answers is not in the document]
leavesandlemmas: > In fact, I don't even see a good reason not to allow a one-row driver (perhaps we just want to run direct modules on the data for a single row!) except that it complicates the logic of the lines below (not to mention that the associated system no longer deserves to be called "dynamic").

I don't think I understand. The check passes for drivers which are only 1  or 2 rows. There aren't enough rows to check the spacing, but I can make the check pass if there aren't any differential modules.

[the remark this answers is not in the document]
eloch216: I don't think this new check would break that example, because it will always pass for drivers with fewer than 3 rows. Besides, `R CMD check` will fail if that example stops working, so we will know about it.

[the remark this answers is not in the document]
gsrohde: I rather hastily concluded that the branch returning an `error_message` object meant that the test failed, and I also missed the comment clearly stating otherwise.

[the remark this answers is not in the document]
gsrohde: > Besides, `R CMD check` will fail if that example stops working, so we will know about it.

Yes, you're right.  Somehow I was thinking that vignettes that were output as PDFs didn't get checked.  I've been away from all this for awhile.

gsrohde:
> I have found the following workflow to be helpful when making changes to BioCro

That does seem helpful, and I'm tempted to say this should be put in the documentation somewhere.  But there's always the perennial question of whether the documentation gets used enough to justify the effort and the related question of how reliable and up to date the documentation is (related, because if users suspect it isn't reliable or up to date, they are less likely to pay attention to it).  Checked examples at least provide some assurance that some portions of the documentation are up to date.  But nothing in the developer's manual gets checked.

(Two of the primary motivations I have found for writing documentation are to help me think through issues by having to clearly explain them, and to remind myself of how to do certain things that I would always be having to reconstruct or re–look up if I didn't write them down somewhere.)

eloch216:
> > I have found the following workflow to be helpful when making changes to BioCro
> 
> That does seem helpful, and I'm tempted to say this should be put in the documentation somewhere. But there's always the perennial question of whether the documentation gets used enough to justify the effort and the related question of how reliable and up to date the documentation is 

I think it's a good idea to include this in the documentation. It's not something that will need to change very often, and I think it may be helpful when "onboarding" new BioCro contributors.

I will add an "Etiquette for Pull Requests" section to the Contribution Guidelines with this list and a few other points. Stay tuned.

justinmcgrath:
> It looks like there are some test failures. I haven't fully investigated, but I think they are related to `validate_dynamical_system_inputs`. Like `run_biocro`, it also calls `add_time_to_weather_data`. Now this needs to be done before calling `check_run_biocro_inputs`.
> 
> BTW, I have found the following workflow to be helpful when making changes to BioCro:
> 
>     1. Build the package and run the tests locally. The tests don't take much time to run so this is a good way to get fast feedback.
> 
>     2. When all the tests are passing, run `R CMD check` locally. This takes longer than just running the tests, but it will also check the examples and vignettes, which sometimes turns up more issues.
> 
>     3. When `R CMD check` is passing locally, make a PR, but don't assign any reviewers. Making the PR will cause the workflow tests to run, which includes `R CMD check` on several operating systems. Sometimes an issue will appear that didn't occur locally on your own operating system / R version.
> 
>     4. When the online checks have passed, then assign reviewers to begin the review process.
> 
> 
> Doing this will ensure that there are no basic issues when reviewers begin to look at the code. In the long run, it saves time.
> 
> Sometimes the online tests will fail due to an issue with the testing setup itself. In that case you don't need to wait for them to pass.

This would be nice to have in the guidelines for reference.

R/input_checking_functions.R
```diff
@@ -272,3 +272,35 @@ check_boolean <- function(args_to_check) {
...
+# inexact floating point arithmetic.
+is_time_sequential <- function(drivers, differential_modules, rtol = sqrt(.Machine$double.eps)){
+    if (length(differential_modules) == 0) {
+        return (TRUE)
+    }
+
+    time_is_null <- is.null(drivers[['time']])
+    if (time_is_null) {
+        stop("No `time` variable found in the `drivers` dataframe.")
+        return (FALSE)
```
justinmcgrath, line 287: stop() ends execution, so the return statement won't ever be reached.  Overall, this is not the approach that I would use. I don't know what to call the idea, but here you're interacting directly with the user in a location where that isn't expected. One way to think of it, is that there are two paths for information to flow through the user - whatever called chain leads to this function, and this function directly.

More than one path makes things difficult to follow. I spotted this because I couldn't figure out what was going on in run_biocro with `time_is_sequential`. It's minor here, and with only one case it wouldn't be a big deal, but this is the sort of design that if propagated makes code very hard to follow.

Instead, return FALSE, and handle the error in run_biocro, with the error_messages. It looks like there's already a function to check inputs, so putting the time check there seems appropriate.

90% of this has been added since I've been able to closely follow the code, so if the details of what I've said don't make sense, please ignore them, but hopefully the overall advice is helpful.

justinmcgrath: To add a bit more, exceptions (like in C++) deliberately do this, but they can be caught at higher levels so there's still control over what happens. I feel like they can make sense (although some people adamantly dislike exceptions). The use here isn't like exceptions though.

justinmcgrath: It looks like previously this did use the call chain and this change is recent. If you describe what wasn't working with that approach, someone could probably help.

leavesandlemmas: > Instead, return FALSE, and handle the error in run_biocro, with the error_messages. It looks like there's already a function to check inputs, so putting the time check there seems appropriate.

I tried this, but it doesn't work without redesigning. Implemented that way, the tests fail because one of the tests checks that the drivers are a dataframe, but if the drivers aren't a dataframe, the time_is_sequential check hits a "subscript out of bounds" error.

justinmcgrath: My first thought is that check_run_biocro_inputs could return immediately after failing the data frame check. That isn't ideal because it would be nice to get all of the errors at once.

The time check could be included with the data frame check. That let's all check continue, but logically it is a bit odd. Regardless, it seems like the better option to me.

Maybe someone else has better suggestions.

justinmcgrath: As a comment unrelated to work here, it would be clearer if `send_error_messages` were named `stop_and_send_error_messages`.

eloch216:
A new file has been added: `tests/testthat/testthat-problems.rds`. It seems like this is a byproduct of `testthat` failures, although I haven't seen it before. Nevertheless, it should probably be added to `.gitignore` and removed from the repository.

.gitignore
```diff
@@ -71,3 +71,6 @@ tags
 docs/
 index.md
 about_pkgdown_documentation.md
+
+# Files generated by testthat
+.rds
```
eloch216, line 76: It might be better to make this more specific. An `rds` file is a binary file containing a single R object, as created by `saveRDS`. It's possible that the package could contain legitimate `rds` files that should be tracked by `git`, such as in the `data` directory or elsewhere in the `tests` directory (representing saved output from a function).

eloch216:
I think it would be helpful to add some tests based on Issue https://github.com/biocro/biocro/issues/129 to make sure these types of mistakes are always caught. Here is a suggestion. These tests could be included in a new file: `tests/testthat/test.check_time_is_sequential.R`.

```
# Set up harmonic oscillator inputs
times_sequential <- seq(0, 10, 1)            # gives correct behavior
times_repeated   <- c(0, rep(1:5, each = 3)) # repeats times
times_disordered <- c(0, 2, 1, 7, 10, 8, 3)  # not sequential or evenly spaced
times_reversed   <- rev(times_sequential)    # evenly spaced but not increasing

testing_model <- list(
    initial_values = list(position = 1, velocity = 0),
    parameters = list(spring_constant = 1, mass = 1, timestep=1),
    direct_module_names = list(),
    differential_module_names = list("BioCro:harmonic_oscillator"),
    ode_solver = soybean$ode_solver
)

test_that('Time errors are caught', {
    # Sequential times should work
    expect_silent(
        with(testing_model, {run_biocro(
            initial_values,
            parameters,
            data.frame(time = times_sequential),
            direct_module_names,
            differential_module_names,
            ode_solver
        )})
    )

    # Repeated times should cause an error
    expect_error(
        with(testing_model, {run_biocro(
            initial_values,
            parameters,
            data.frame(time = times_repeated),
            direct_module_names,
            differential_module_names,
            ode_solver
        )}),
        'The `time` variable is not evenly spaced or sequential.'
    )

    # Disordered times should cause an error
    expect_error(
        with(testing_model, {run_biocro(
            initial_values,
            parameters,
            data.frame(time = times_disordered),
            direct_module_names,
            differential_module_names,
            ode_solver
        )}),
        'The `time` variable is not evenly spaced or sequential.'
    )

    # Reversed times should not cause an error
    expect_silent(
        with(testing_model, {run_biocro(
            initial_values,
            parameters,
            data.frame(time = times_reversed),
            direct_module_names,
            differential_module_names,
            ode_solver
        )})
    )
})
```

R/input_checking_functions.R
```diff
@@ -272,3 +272,34 @@ check_boolean <- function(args_to_check) {
     }
     return(error_message)
 }
+
+# Checks that the `time` variable is ordered, strictly increasing, and evenly spaced with the `timestep`.
+# The check is that for all n, dt = t[n] - t[n-1] up to error tolerance due to
+# inexact floating point arithmetic.
+check_time_is_sequential <- function(drivers, differential_modules, rtol = sqrt(.Machine$double.eps)){
```
eloch216, line 279: The input arguments could be moved to separate lines to avoid exceeding 80 characters.

R/input_checking_functions.R
```diff
@@ -272,3 +272,34 @@ check_boolean <- function(args_to_check) {
...
+    if (length(differential_modules) == 0) {
+        return (character())
```
eloch216, line 281: In R, `return` is a function, so we usually use it like `return(X)` (without a space after the word "return"). This is different from C++ where `return` is a keyword.

There are a few other places where there's an unnecessary space after `return`.

R/input_checking_functions.R
```diff
@@ -272,3 +272,34 @@ check_boolean <- function(args_to_check) {
...
+
+    no_time_variable <- !('time' %in% names(drivers))
+    if (no_time_variable) {
+        return ("No `time` variable found in the `drivers` dataframe.")
+    }
+
+    time <- drivers[['time']]
+    if (length(time) < 3) {
+        # automatic pass because >2 rows are needed to check the spacing.
+        return (TRUE)
```
eloch216, line 292: Elsewhere in this file the "no error" value is `character()`. It would be good to stick to that convention here.

R/input_checking_functions.R
leavesandlemmas: I did it this way to mirror the spaces in `if (x) {}`; it seems like the same reasons would apply.

leavesandlemmas:
> I think it would be helpful to add some tests based on Issue #129 to make sure these types of mistakes are always caught. Here is a suggestion. These tests could be included in a new file: `tests/testthat/test.check_time_is_sequential.R`.
> 
> ```r
> # Set up harmonic oscillator inputs
> times_sequential <- seq(0, 10, 1)            # gives correct behavior
> times_repeated   <- c(0, rep(1:5, each = 3)) # repeats times
> times_disordered <- c(0, 2, 1, 7, 10, 8, 3)  # not sequential or evenly spaced
> times_reversed   <- rev(times_sequential)    # evenly spaced but not increasing
> 
> testing_model <- list(
>     initial_values = list(position = 1, velocity = 0),
>     parameters = list(spring_constant = 1, mass = 1, timestep=1),
>     direct_module_names = list(),
>     differential_module_names = list("BioCro:harmonic_oscillator"),
>     ode_solver = soybean$ode_solver
> )
> 
> test_that('Time errors are caught', {
>     # Sequential times should work
>     expect_silent(
>         with(testing_model, {run_biocro(
>             initial_values,
>             parameters,
>             data.frame(time = times_sequential),
>             direct_module_names,
>             differential_module_names,
>             ode_solver
>         )})
>     )
> 
>     # Repeated times should cause an error
>     expect_error(
>         with(testing_model, {run_biocro(
>             initial_values,
>             parameters,
>             data.frame(time = times_repeated),
>             direct_module_names,
>             differential_module_names,
>             ode_solver
>         )}),
>         'The `time` variable is not evenly spaced or sequential.'
>     )
> 
>     # Disordered times should cause an error
>     expect_error(
>         with(testing_model, {run_biocro(
>             initial_values,
>             parameters,
>             data.frame(time = times_disordered),
>             direct_module_names,
>             differential_module_names,
>             ode_solver
>         )}),
>         'The `time` variable is not evenly spaced or sequential.'
>     )
> 
>     # Reversed times should not cause an error
>     expect_silent(
>         with(testing_model, {run_biocro(
>             initial_values,
>             parameters,
>             data.frame(time = times_reversed),
>             direct_module_names,
>             differential_module_names,
>             ode_solver
>         )})
>     )
> })
> ```
@eloch216  do I need to modify any other files to add the test ? Or just add a new file to the `testthat` directory?

R/input_checking_functions.R
eloch216: This could also be a good thing to include in the tests. (Checking that a driver with 2 rows passes.)

eloch216:
> @eloch216 do I need to modify any other files to add the test ? Or just add a new file to the `testthat` directory?

Just add a new file. By default, the `testthat` package will run all the files in `tests/testthat`.

R/input_checking_functions.R
eloch216: I can see your point. But the major goal is for all the code to be styled the same way. No other instances of `return` in the `R` directory have a space after them. On the other hand, all instances of `if` do have a space after them. (There is one exception in `check_names`, which is a mistake.)

eloch216: Update: I think `if` is not a function. You can type `return` in R to see the content of the `return` function. But that doesn't seem to work for `if`. Hence the different spacing syntax.

leavesandlemmas: > I can see your point. But the major goal is for all the code to be styled the same way. No other instances of `return` in the `R` directory have a space after them. On the other hand, all instances of `if` do have a space after them. (There is one exception in `check_names`, which is a mistake.)

I've updated the spacing in the syntax. 

> Update: I think `if` is not a function. You can type `return` in R to see the content of the `return` function. But that doesn't seem to work for `if`. Hence the different spacing syntax.

FWIW I get the same thing for both, just have to use back ticks for `if` (and `for` or operators like `+`) :

```r
`if` 
# .Primitive("if")
return 
# .Primitive("return")

is.function(`if`)
# TRUE
```

eloch216: This is completely pointless, but I learned that it's possible to redefine `return` so it calls itself, generating an infinite recursion error:

```R
return <- function(x) {return(x)}
add <- function(a, b) {return(a + b)}
add(1, 2)
```

eloch216: Since this has come up, I added these considerations to the style guide in PR https://github.com/biocro/biocro/pull/141

R/input_checking_functions.R
leavesandlemmas: THese are fixed and I added a test.

leavesandlemmas:
I don't have any other planned changes on this PR.

eloch216:
> I don't have any other planned changes on this PR.

In that case, I will just go ahead and commit my final pedantic changes

eloch216 (approved):

eloch216:
Oh, something I just remembered: It would be a good to add a quick note to `NEWS.md` about these changes. Just something like "`run_biocro` and `validate_dynamical_system_inputs` now check to make sure `time` values are sequential and evenly-spaced."

R/input_checking_functions.R
justinmcgrath: To add to this pointless thread, if return is a function, what does it return?

justinmcgrath (approved):
Looks good other than the conflicts in NEWS.md.

R/input_checking_functions.R
leavesandlemmas: > To add to this pointless thread, if return is a function, what does it return?

Like mathematically as a function? Or in R?  It seems like R makes a distinction between specials, builtins, and closures. Normal functions are closures, but return is a special. But I have no idea what that means under the hood. Mathematically, it would seem to me that `return` maps from a function's local scope to the scope in which the function is called, and returns the value of whatever is passed to `return`. 

On unrelated note, TIL that these are equivalent:
```r
f <-  \(x)  x + 1
f  <- function(x) x + 1
```

gsrohde: > IDK if you can call `if` a function in R.

For what it's worth, the _Advanced R_ book says "Every operation is a function call" (including control flow operators like `for`, `if`, and `while`).  See http://adv-r.had.co.nz/Functions.html#all-calls.

gsrohde: > On unrelated note, TIL that these are equivalent:
> 
> ```r
> f <-  \(x)  x + 1
> f  <- function(x) x + 1
> ```

Well, I'm going to be very pedantic here, but they aren't quite equivalent: `f` seems to remember which syntax was used to define it.  With the first definition,
```
> f
\(x)  x + 1
```
With the second,
```
> f
function(x)  x + 1
```

I would have expected the definition syntax in the printed definition to be normalized to one or the other, but it isn't.

Also surprising to me is that the new `\(` syntax doesn't seem to have made it into the CRAN R manuals, even though it's been around for a few years.  And in the one place I did find to read about it—an article on medium.com—it didn't exactly get rave reviews in the comments section.

gsrohde: > This is completely pointless, but I learned that it's possible to redefine `return` so it calls itself, generating an infinite recursion error:
> 
> ```r
> return <- function(x) {return(x)}
> add <- function(a, b) {return(a + b)}
> add(1, 2)
> ```

Maybe pointless to do, but not pointless to know about.  I can imagine that such redefinitions could be a vector for some rather insidious hacking.  Consider
```
`if` <- function(a, b) { <erase a bunch of valuable files> }
```

eloch216:
I think this can be merged now.

For a next step, I would suggest changing `time` (as added to the drivers via `add_time_to_weather_data` and as used by a few modules such as `solar_position_michalsky`) to units of hours, instead of days.  It could probably be defined as fractional hours since the start of year, as we currently do when using days. This will make our models a bit easier to understand, since the `timestep` and `time` quantities will use the same time units as the module derivatives. This change should probably be a PR on its own.

When that PR is done and merged, then it would be possible to check that the values of `time` are actually spaced by `timestep`. That would be another PR, which should be fairly small.

Between this PR and the two ones suggested above, I think this would accomplish most of what PR https://github.com/biocro/biocro/pull/137 set out to do, but in smaller chunks that are easier to evaluate.

Issue #142, opened 2025-01-13
Changing the units of `time` to hours.


leavesandlemmas:
Regarding the recent PR on checking if `time` is evenly spaced, @eloch216  suggested:

> For a next step, I would suggest changing `time` (as added to the drivers via `add_time_to_weather_data` and as used by a few modules such as `solar_position_michalsky`) to units of hours, instead of days.  It could probably be defined as fractional hours since the start of year, as we currently do when using days. This will make our models a bit easier to understand, since the `timestep` and `time` quantities will use the same time units as the module derivatives. This change should probably be a PR on its own.

> When that PR is done and merged, then it would be possible to check that the values of `time` are actually spaced by `timestep`. That would be another PR, which should be fairly small.

> Between this PR and the two ones suggested above, I think this would accomplish most of what PR https://github.com/biocro/biocro/pull/137 set out to do, but in smaller chunks that are easier to evaluate. These PRs won't change the user interface, so most old scripts should continue to work, but this will set us up to make some more complicated changes to the framework itself later.

_Originally posted by @eloch216 in https://github.com/biocro/biocro/issues/140#issuecomment-2587616860_

So here is a list of all modules that take `time` as a variable.
Direct Modules:
+ `fake_solar` (input)
+ `grimm_soybean_flowering_calculator` (input)
+ `oscillator_clock_calculator` (input)
+ `solar_position_michalsky` (input)
+ `soybean_development_rate_calculator` (input)
+ `thermal_time_development_rate_calculator` (input) 

Differential modules: 
+ `thermal_time_and_frost_senescence` (input)
+ `thermal_time_beta` (input)
+ `thermal_time_bilinear` (input)
+ `thermal_time_linear` (input)
+ `thermal_time_linear_extended` (input)
+ `thermal_time_trilinear` (input)

The thermal time modules also take a `sowing_time` parameter in days and they just check whether ` time > sowing_time `. 

I can change all of these modules to take `time`  measured in hours (and `sowing_time` would implicitly be in hours). However, I am wondering if it would be better to have these modules take `doy` rather than `time`. 

Also, the `solar_position_michalsky` module actually takes `time` and extracts `hour` and `doy` from it, which seems a little silly since `doy` and `hour` are invariably provided to `run_biocro`. 

Alternatively, the `time` variable could be computed from / passed with the drivers , and then a direct module computes other types of time-variables.

gsrohde:
> * thermal_time_development_rate_calculator

This is actually a direct module.

[...]

PR #143, opened 2025-01-14
142 changing the units of time to hours

leavesandlemmas:
PR based on #142


leavesandlemmas:
I changed the definition of `time` slightly. It's now the amount of time since the beginning of the year in hours. So `time =  60.5` is Jan 3, 12:30.  Thus, we have `time = 24*(doy - 1) + hour` with the `-1` because Jan 1 is first day of the year but less than 24 hours have passed since the year began. 

I think this definition should be tied to `year` at some point so that simulations in southern hemisphere   or simulations that run multiple years  have a continuous variable to index them. For instance if `time = 8783` at Dec 31, 2024 23:00, then it should be `time = 8784` at midnight, and `8785` on Jan 1, 1:00. But that's not currently how `time` is computed. While most crops only grow one season (< 1 y), if BioCro is adapted for simulations of the soil biochemistry or for trees (like orchards), then multiple years would be required. Effectively `time` would be like the julian date but in hours instead of days.

leavesandlemmas:
> It looks like the number formatting has changed a bit the test results. Although the information looks the same, it looks like changes that increase file size, like 0.0 instead of 0. CRAN has a small package size limit that we just squeeze under. If this has made these files larger, then I'd prefer the previous format. If you're not sure how the happened, then we can worry about this the next time we send an update to CRAN.


The saved test data had to change because the old `time` variable is an expected output for many of the tests. So the tests are incompatible with changing `time` from days to hours (and shifting to zero index). I also had to change the harmonic oscillator's test because its code expects `time` to be in units of days rather hours (or rather it expects that `time` 's unit is equal to 24 of `timestep`'s unit).  

An alternative route would be to keep `time` in days and convert the relevant differential modules from rates per hour to rates per day, so that `timestep` can have the same units at `time`.

eloch216:
> It looks like the number formatting has changed a bit the test results. Although the information looks the same, it looks like changes that increase file size, like 0.0 instead of 0.

@leavesandlemmas How did you update the stored test results? If you use the `update_stored_results` function defined in `tests/testthat/crop_model_testing_helper_functions.R`, the results should be automatically rounded. If you are already using that function and the file size has nevertheless increased, then we can instead deal with this when getting ready to submit to CRAN (if file size is a problem).

eloch216:
FYI, my usual workflow for updating the test results is:

1. Set the working directory of an R session to `tests/testthat`.

2. Source `crop_model_testing_helper_functions.R`.

3. Compare the new and old results using the suggested code in lines 144-145:
```
compare_crop_output <- compare_crop(PLANT_TESTING_INFO[[INDEX]], c('time', 'Leaf', 'Stem', 'Root', 'Grain'))
lattice::xyplot(Leaf + Stem + Root + Grain ~ time, group = version, data = compare_crop_output, type = 'l', auto = TRUE, grid = TRUE)
```

where `INDEX` can be set to 1, 2, or 3 for miscanthus, willow, and soybean. It is very important to make this manual check to confirm that the results have not changed in unexpected ways. For this PR I would expect to see no change in the biomass values.

4. If the new results look okay, then update the stored results with `update_stored_results(PLANT_TESTING_INFO[[INDEX]])`.

While this is on my mind, at some point I would like to move the contents of `crop_model_testing_helper_functions.R` to the `BioCro` package namespace. These functions will be useful to anyone with an independent module library, so they can test their models. (For example, so the nitrogen uptake and distribution model can have regression tests.) They should be renamed to be more general, since they are not really exclusive to crop models.

leavesandlemmas:
> > It looks like the number formatting has changed a bit the test results. Although the information looks the same, it looks like changes that increase file size, like 0.0 instead of 0.
> 
> @leavesandlemmas How did you update the stored test results? If you use the `update_stored_results` function defined in `tests/testthat/crop_model_testing_helper_functions.R`, the results should be automatically rounded. If you are already using that function and the file size has nevertheless increased, then we can instead deal with this when getting ready to submit to CRAN (if file size is a problem).

I edited the files using a text editor, except the simulation outputs. For those, I using a python script to modify the `time` column, using the pandas library for the heavy lifting of parsing the csv (although its not hard to write a function using the standard library only).

```python
import pandas as pd
def rewrite_time(filename ):
    dat = pd.read_csv(filename)
    dat['time'] = 24*(dat['doy'] - 1) + dat['hour']
    dat.to_csv(filename, index=False)

```

justinmcgrath:
> The saved test data had to change because the old `time` variable is an expected output for many of the tests.

I expect some things to change, but most of the changes were different formats of the same number: 0 -> 0.0; 8e-04 -> 0.0008.

justinmcgrath:
If this takes any longer than 20 minutes to figure out, don't worry about it. It's very likely we'll need to do something about the package size the next time we submit not matter what is done here. If it's easy, I think it's worth doing now. If it's hard, it will be more efficient to address it as part of a larger problem.

leavesandlemmas:
> > The saved test data had to change because the old `time` variable is an expected output for many of the tests.
> 
> I expect some things to change, but most of the changes were different formats of the same number: 0 -> 0.0; 8e-04 -> 0.0008.

Oh, that's probably a difference in parsing of the CSV. I think we could solve this issue using compression. For example, we already compress the weather data into a binary file (`.rdata`). It looks like it would halve the size of the `soybean_simulation.csv`

justinmcgrath:
> I think we could solve this issue using compression.

I recall that we have a reason for not doing that. Hopefully someone else remembers. If we it is a possible solution, it's that sort of thing that would be good for a different PR as part of preparing the package for CRAN.

leavesandlemmas:
> > It looks like the number formatting has changed a bit the test results. Although the information looks the same, it looks like changes that increase file size, like 0.0 instead of 0.
> 
> @leavesandlemmas How did you update the stored test results? If you use the `update_stored_results` function defined in `tests/testthat/crop_model_testing_helper_functions.R`, the results should be automatically rounded. If you are already using that function and the file size has nevertheless increased, then we can instead deal with this when getting ready to submit to CRAN (if file size is a problem).

So the file size has still increased, but the increase is a bit smaller after using `update_stored_results` 

> > I think we could solve this issue using compression.
> 
> I recall that we have a reason for not doing that. I recall that it made it difficult to see changes. Hopefully someone else remembers. If we it is a possible solution, it's that sort of thing that would be good for a different PR as part of preparing the package for CRAN.

It's true that `.rdata` isn't a text format so you have to have R's functions for parsing the binary file. So it's less good for long term data storage where the data might be read by other programs in the future. But it seems to me that the test_data would be safe to compress. 

Some other kind of compression could be used. It would make a big difference. UTF-8 text encoding means that every character symbol needs at least 1 byte (assuming only ASCII characters), but since most of the file are floats, you need a byte for every significant digit (plus one for the comma / newline, and one for the decimal point). So saving 12 digits per float plus the comma and decimal point is 14 bytes per entry. But a 64-bit float only needs 8 bytes (or about a 33%  reduction in size). And it would be possible to reduce the precision to 32 bit floats, only 4 bytes (for a 29% reduction). 

As another idea, the weather data could split off into another package. The weather data is pretty big even with the compression to `.rdata` . That would remove 2 MB of storage from BioCro. For testing BioCro, maybe a module could compute test weather data from a given time.

tests/testthat/test.HarmonicOscillationModeling.R
```diff
@@ -132,7 +132,7 @@ run_trial <- function(initial_position, initial_velocity, mass, spring_constant,
     result <- run_biocro(initial_values, parameters, drivers, direct_modules, differential_modules, ode_solver)
 
     ## add useful columns to the resulting data frame:
-    result$time <- result$time * 24 # time is in hours
+    # result$time <- result$time * 24 # time is in hours
```
gsrohde, line 135: Did you mean to ultimately remove this, not just comment it out?  It seems like you may as well.

leavesandlemmas: Yes,  just forgot to remove it.

src/module_library/fake_solar.h
```diff
@@ -69,8 +69,9 @@ void fake_solar::do_operation() const
     // Calculate the radiation in a way that's good for testing different day lengths, and the switch to dark conditions
 
     // Extract out the DOY and hour
-    int doy = (int)time;
-    double hour = 24.0 * (time - doy);
+    int d =  std::floor(time / 24);
```
eloch216, line 72: Since this module now uses `std::floor`, which is part of the `cmath` header library, this file should have `#include <cmath> // for std::floor` up at the top, just before the includes for `module.h` and `state_map.h`.

src/module_library/thermal_time_and_frost_senescence.h
```diff
@@ -283,7 +283,8 @@ void thermal_time_and_frost_senescence::do_operation() const
     // linearly with temperature.
     bool const sene_started = TTc >= seneLeaf;
     bool const in_north = lat >= 0.0;
-    bool const late_year = time >= 180.0;
+    int const doy = 1 + std::floor(time / 24);
```
eloch216, line 286: Since this module now uses `std::floor`, which is part of the `cmath` header library, this file should have `#include <cmath> // for std::floor` up at the top, just before the includes for `module.h` and `state_map.h`.

eloch216:
Overall this looks great.

The only snag is that I suspect some of our documentation may mention a requirement to include a `time` column expressed in days. Any such documentation needs to be updated, but may be difficult to find. 

I will try looking around to see if I can find any documentation that needs to be updated.

eloch216:
I found quite a few places where the documentation should be updated:
- `man/add_time_to_weather_data.Rd`
- `man/run_biocro.Rd`
- `man/system_derivatives.Rd`
- `vignettes/web_only/an_introduction_to_biocro/test-heating_degree_days_with_run_biocro.R`
- `vignettes/web_only/a_practical_guide_to_biocro.Rnw`
- `vignettes/web_only/BioCro-II_Paper--Section-1.1-example.Rmd`
- `vignettes/web_only/dynamical_systems.Rmd`
- `vignettes/web_only/quantitative_model_comparison.Rnw`

For `system_derivatives.Rd`, I am not 100% if the description of Example 3 is still correct.

I didn't actually find anything in `dynamical_systems.Rmd` that I am 100% sure needs to be changed, which was surprising. But I think the the error message mentioned in "A minimal system" should at least be checked again to make sure this is still true. And it would be good to double check that nothing else needs to change.

Time in days is also mentioned in `vignettes/web_only/an_introduction_to_biocro.R`, but this vignette is out of date in many respects so it may not be worth addressing this one. For example, it refers to a `doy_dbl` column in the drivers, which was renamed to `time` a long time ago. This document needs a big overhaul and this is just one of many things that need to change.

Besides these files, `NEWS.md` should also be updated to include the change from days to hours.

justinmcgrath:
I also like the CSVs and if we end up needing space, I would favor other approaches to gaining it. They're less than 1 MB now, so we'd get little anyway. If it comes to it, then something other than Rdata for compression would likely be more flexible.

leavesandlemmas:
> One of my hopes for the regression tests is that they would be able to run in C++ without the R interface, even if this is only true in principle rather than in practice. This would require that all the relevant info -- the model definition, the weather data, and the saved results -- are saved in simple formats that can be read in C++. Right now this is definitely not the case, at the very least because the weather data is only available in binary format.

The HDF5 file format might be a good candidate. It was developed to store climate simulations, I believe. I think it was originally written as a C library (so c++ compatible) and there are bindings for python and R. There's probably also some file format that's used for databases.  Certain scalars are duplicated for every row because the output is a dataframe, but maybe a non dataframe file format might allow for less storage.

Alternatively, the regression tests could use a fake weather simulation, which is simply computed from a timestamp. That would mean that they wouldn't need to load any weather data.

gsrohde:
> One of my hopes for the regression tests is that they would be able to run in C++ without the R interface, even if this is only true in principle rather than in practice. This would require that all the relevant info -- the model definition, the weather data, and the saved results -- are saved in simple formats that can be read in C++.

For what it's worth, I wrote some regression tests in C++ for a BioCro-related project I was working on last year (https://github.com/gsrohde/DAWN), though I should mention that these were mainly focused on testing the framework I was writing than on testing BioCro modules.  Also, input data, including weather data, was in XML, which is even bulkier that CSV, though perhaps C++ has a compression library that would ameliorate this problem for XML data (or for that matter, for CSV data).  I mention this at least partly in case you are or will be shopping for C++ test frameworks.  I used GoogleTest and was fairly happy with it.  I may be biased by having invested not a little time in learning how to use it.

In general, it would seem on principle better to be testing the C++ code in C++ and reserve the R tests for R code, but I haven't really tried to think this through.  Moreover, even if a C++ testing framework were to be used to test the C++ portion of the code, it would still be convenient to be able to have those tests run and reported via the R testing framework.

leavesandlemmas:
I think the most elegant and future proof solution is to add some functions which convert between a timestamp (POSIX objects) and the `time` variable. That way, timestamps or day of year could used in the vignettes and this is will be converted to `time`. 

That way, if we change the origin/epoch for `time`, then we would only have to modify these functions rather than every use of `time`.

vignettes/web_only/BioCro-II_Paper--Section-1.1-example.Rmd
```diff
@@ -222,9 +223,9 @@ xyplot(Leaf ~ time / 3600, data = result, xlim = c(0, 24), xlab = "time (in hour
 ## Second modification: Changing the time specification from seconds to days
 
 The usual assumption in BioCro is that the `time` quantity represents
-the time, in days, from the beginning of the year, not the number of
-seconds that have elapsed from the start of the simulation.  (More
-precisely, the integral part of the `time` value should be the day of
+the time, in hours, that has elapsed from the beginning of the year,
+not the number of seconds that have elapsed from the start of the simulation.
+(More precisely, the integral part of the `time` value should be the day of
```
justinmcgrath, line 228: Hasn't this changed to the integral part is hours? I don't know whether now it starts at 1 or 0.

vignettes/web_only/BioCro-II_Paper--Section-1.1-example.Rmd
```diff
@@ -263,15 +264,17 @@ result <- run_biocro(
...
+result$time <- result$time /86400
+
 cable(result[1:4,]) # beginning
 cable(result[seq(21591, 21661, 10),]) # around dawn
 cable(result[43199:43203,]) # around mid day
 cable(result[seq(64741, 64811, 10),]) # around dusk
 cable(result[86398:86401,]) # end
 
-xyplot(Q ~ 24 * time, data = result, xlim = c(0, 24), xlab = "hour", ylab = "Q (mol / m^2 / s)")
+xyplot(Q ~  24 * time , data = result, xlab = "hour", ylab = "Q (mol / m^2 / s)")
```
justinmcgrath, line 275: Extra spaces have been added.

vignettes/web_only/an_introduction_to_biocro/test-heating_degree_days_with_run_biocro.R
```diff
@@ -4,12 +4,13 @@ direct_modules <- c()
 differential_modules <- c("heating_degree_days")
 
 ## Given an initial value for the number of heating degree days, a
-## vector of equally-spaced time values (as fractional day-of-year
-## values), a corresponding vector of temperatures, a threshold
-## temperature value, a timestep value, and the specification of a
-## solver, run the system and return the final heating-degree-days
-## value.
+## vector of equally-spaced time values, a corresponding vector of
+## temperatures, a thresholdtemperature value, a timestep value, and the
```
justinmcgrath, line 8: Missing space.

vignettes/web_only/an_introduction_to_biocro/test-heating_degree_days_with_run_biocro.R
```diff
@@ -4,12 +4,13 @@ direct_modules <- c()
...
+## specification of a solver, run the system and return the final heatin
```
justinmcgrath, line 9: Line break midword.

R/run_biocro.R
```diff
@@ -205,10 +206,6 @@ run_biocro <- function(
         verbose
     ))
 
-    # Make sure doy and hour are properly defined
-    result$doy = floor(result$time)
-    result$hour = 24.0*(result$time - result$doy)
```
eloch216, line 210: Deleting this lines can cause some undesired behavior. Try running the following code on this branch and on `develop`:

```
library(BioCro)
library(lattice)

short_weather <- soybean_weather$'2002'[seq_len(36), ]

result <- with(soybean, {run_biocro(
  initial_values,
  parameters,
  short_weather,
  direct_modules,
  differential_modules,
  within(ode_solver, {output_step_size = 0.5})
)})

unique(result$doy)

unique(result$hour)

xyplot(Leaf ~ hour | factor(doy), data = result, type = 'b', pch = 16)
```

Results are below. On this branch, there is an unexpected non-integer value of `doy`.

plotting_issue

I can make a PR with a suggested alternative, but just wanted to demonstrate the issue here.

justinmcgrath: Do you know where the 152.5 comes from? I assume it's something with the solver since you set the step size to 0.5 to get this to happen. But it's particularly weird, because I would have thought it would say doy=152.5 and hour=12, which would be consistent with each other if calculating doy with fractions of day. However, in that row, none of doy, hour and time agree with each other. It makes me think there's a problem with something else.

| doy   | hour | time   |
|-------|------|--------|
| 152.0 | 22.0 | 3646.0 |
| 152.0 | 22.5 | 3646.5 |
| 152.0 | 23.0 | 3647.0 |
| 152.5 | 11.5 | 3647.5 |   # None of these agree with each other.
| 153.0 |  0.0 | 3648.0 |
| 153.0 |  0.5 | 3648.5 |

leavesandlemmas: > Do you know where the 152.5 comes from?

It's linearly interpolated from 152 and 153 at `time = 3647` and `3648`. Same for the `hour` variable.  The c++ code treats DOY and hour like temperature as Ed mentioned before. Also, I think they are consistent. It's just that 3647.5 is 30 minutes before midnight on DOY 153. So it's 152.9792 for fractional DOY, 

```r
fractional_doy <- 1 + 3647.5 / 24 
fractional_doy
fractional_doy_v2 <- 152.5 + 11.5/24
fractional_doy_v2
```

I thought this would be easiest to see using the datetime object, but sadly it automatically converts to daylight savings time, And 3647.5 is 30 min after midnight on DST. 

```r
as.POSIXct('2002-01-01') + as.difftime(3647.5, units='hours')
```

So should I add the correction back or do we want to consider other alternatives ?

eloch216: This was addressed by PR https://github.com/biocro/biocro/pull/144

man/soybean.Rd
```diff
@@ -140,7 +140,7 @@
     \item The \code{sowing_time} input to the
     \code{soybean_development_rate_calculator} module is set to 0 because
     Soybean-BioCro uses the weather data to set the sowing time. In other words,
-    the weather data is truncated so it begins at the sowing date.
+    the weather data is truncated so it begins at the beggining of the simulation.
```
eloch216, line 143: "beggining" -> "beginning"

Though I'm not sure if I see why this wording needed to change

eloch216:
I still want to take a closer look at the documentation (as mentioned above), but I may not be able to get to it for a few days. Feel free to merge before then if you want. Overall I am happy with the switch from days to hours.

leavesandlemmas:
@eloch216  Have you had a chance to look through the documentation ?  If it helps, I wrote a documentation page for the `time` variable's definition so maybe other parts of the documentation can just point that  (in case of future changes).

If I don't hear anything by end of tomorrow, I will plan to merge this into develop over the weekend or next monday.

eloch216:
> @eloch216 Have you had a chance to look through the documentation ? If it helps, I wrote a documentation page for the `time` variable's definition so maybe other parts of the documentation can just point that (in case of future changes).
> 
> If I don't hear anything by end of tomorrow, I will plan to merge this into develop over the weekend or next monday.

I haven't been able to look yet -- I've been a bit swamped with other stuff (PRU seminar, PhotoGEA reviewer feedback, traveling back to Illinois from Canada). I probably won't be able to really look at it until next week. I can always just make suggested edits in a new PR if you merge this one.

leavesandlemmas:
> > @eloch216 Have you had a chance to look through the documentation ? If it helps, I wrote a documentation page for the `time` variable's definition so maybe other parts of the documentation can just point that (in case of future changes).
> > If I don't hear anything by end of tomorrow, I will plan to merge this into develop over the weekend or next monday.
> 
> I haven't been able to look yet -- I've been a bit swamped with other stuff (PRU seminar, PhotoGEA reviewer feedback, traveling back to Illinois from Canada). I probably won't be able to really look at it until next week. I can always just make suggested edits in a new PR if you merge this one.

I don't suppose there's any rush in merging... Let me know next week.

eloch216:
I found a few things to change in the documentation. Several of them were totally unrelated to changes in this PR, but I happened to notice them. For example, there was a "TO-DO" in the "Practical Guide" that could now be addressed by adding a link to the quantity documentation web site.

The biggest changes were to the "BioCro II paper Section 1.1 Example" vignette. As originally written, it discussed three modifications to the example from the BioCro II paper:
1. Switching to a 24 hour time period (from a 12 hour time period)
2. Switching the units of time from seconds to days (to match the typical BioCro meaning of time)
3. Switching to a particular day of the year rather than using elapsed time

With the changes in this PR, Modification 2 is no longer possible. We cannot change `time` to the usual BioCro units (now hours rather than days), because the intervals between successive `time` values must be equal to the `timestep` (this is checked by `run_biocro`), and the `timestep` must be in seconds (because the module returns a rate in units of per second). Since Modification 3 is an extension of Modification 2 (essentially it adds an offset to the `time` values), it is also no longer possible to do. On a different note, `doy` and `hour` are no longer included in the results, so we don't need to discuss them anymore in this vignette.

The discussion about modules that have `time` as an input is still relevant so I tried to keep that.

Before merging, I will wait a few days for any comments from @gsrohde since he originally created this vignette.

eloch216:
I'll go ahead an merge this since there haven't been any other comments since my last one.

PR #144, opened 2025-01-17
Ensure integer doy

eloch216:
Here's my fix for the issue described in this comment: https://github.com/biocro/biocro/pull/143#discussion_r1919247708

There were two suggestions for addressing this. Justin's suggestion was to make the lines in `run_biocro` conditional, so that `doy` and `hour` would only be recomputed when the `drivers` had `doy` and `hour` columns. Scott O's suggestion was to make a module for computing `doy` and `hour` from `time`.

I implemented Justin's version first since it is simpler. I also added some tests to check that `run_biocro` has the correct behavior in a few important situations.

Originally I thought the module approach might be too complicated. If `doy` and `hour` are the outputs of a module, then they cannot be included in the `drivers`. I thought this might create a headache and cause many old scripts to break.

But I think I found a good way to deal with this. I changed `add_time_to_weather_data` so it also removes `doy` and `hour` when adding `time`, and added a new module called `format_time`, whose name was inspired by `format.POSIXlt` in R.

I added the new module to the crop model definitions, so any code that runs those models will function as before. But users running different models (such as myself with my cowpea model) will not automatically know they need to add the `BioCro:format_time` module. To address this, I added a warning to `add_time_to_weather_data`. 

If `doy` and `hour` are removed but the new module is not included, then a warning is sent, advising the user to include `BioCro:format_time`. Otherwise their code would run, but the results would mysteriously lack `doy` and `hour`. I think this is "polite" way to make sure people know how to respond to this change, since I don't expect people to actually look at the change log.

I think the module approach is nice since it will also be effective for C++ users, rather than relying on R code in `run_biocro`.

BTW, I used the new `module_write` function to initialize the new module header file and it worked great.


justinmcgrath:
My first thought is that changing how its interpolating doy and hour removes the need for any of this, but maybe fixing that isn't possible.

This looks good in lieu of that. I don't love it, since it seems pretty complicated for something that is easily calculated after the model run, but I'm not sure that what I suggested is better. format_time here is not particularly similar to format.POSIXlt, so I'm not sure about the name.

add_time_to_weather_data is a convenience function to save a step for our most common use. Thus, it's probably fine for it to add format_time to the drivers itself, rather than warning about it. Then everything will work as before. It might take a bit of reworking the logic, and maybe that happens in two steps in run_biocro, instead of in this function.


if (doy and hour are pesent)
drivers <- add_time_to_weather_data(...) # only adds the time
modules <- add_format_time_to_modules(...)
}

Or put everything in one function and return the two things in a list.

eloch216:
> My first thought is that changing how it's interpolating doy and hour removes the need for any of this, but maybe fixing that isn't possible.

I don't think we want to add a special case to the code for interpolating the drivers. That seems to go against the BioCro philosophy of treating everything the same.

Ultimately, I think `doy` and `hour` just shouldn't be drivers because they cannot be interpolated the same way as other variables like temperature. Yet, they are typically included in weather data, and most people expect them to be in the output from `run_biocro` when using weather data.

> This looks good in lieu of that. I don't love it, since it seems pretty complicated for something that is easily calculated after the model run, but I'm not sure that what I suggested is better. format_time here is not particularly similar to format.POSIXlt, so I'm not sure about the name.

I'm not sure it's really any more complicated than what we did previously. We are still re-calculating `doy` and `hour` when necessary, just via a module instead of R code.
 
> add_time_to_weather_data is a convenience function to save a step for our most common use. Thus, it's probably fine for it to add format_time to the drivers itself, rather than warning about it. Then everything will work as before. It might take a bit of reworking the logic, and maybe that happens in two steps in run_biocro, instead of in this function.
> 
> if (doy and hour are pesent) drivers <- add_time_to_weather_data(...) # only adds the time modules <- add_format_time_to_modules(...) }

That's a good idea. Instead of `add_format_time_to_modules`, we could have a more general function for adding modules to lists if they are not already included. That could potentially be useful in other scenarios as well.

justinmcgrath (approved):
I think it looks good. Thanks.

src/module_library/format_time.h
```diff
@@ -0,0 +1,80 @@
...
+        "hour"             // hours
+    };
+}
+
+void format_time::do_operation() const
+{
+    double constexpr hr_per_day = 24.0;
+
+    int const d = std::floor(time / hr_per_day);
+    double const doy = 1 + d;                   // days
```
leavesandlemmas, line 71: I thought of an issue when calculating DOY this. If simulations are allowed to span calendar years, then computed this way, `doy` won't be correct outside the calendar year of `time  = 0`.  For example, if we count hours from midnight Jan 1, 2025,  `time =  -24` should 24 hours before, so midnight Dec 31, 2024.  The DOY should be 366, since Dec 31, 2024 is the 366th day of that year. But this will give 0.

I can't seem to find the right functions in c++ but in R, this gives the correct value

```r
doy = 1 + ((time %/% 24) %% days_in_year(year)

is_leap_year <- function(year){
    divisible_by_4 <- year %% 4 == 0
    divisible_by_100 <- year %% 100 == 0
    divisible_by_400 <- year %% 400 == 0
    divisible_by_4 & (!divisible_by_100 | divisible_by_400)
}

days_in_year <- function(year){
    return(365 + is_leap_year(year))
}
```

leavesandlemmas: I can't find the c++ functions that returns `364` for ` - 1 % 365 `

src/module_library/format_time.h
```diff
@@ -0,0 +1,80 @@
...
+    double const hour = time - hr_per_day * d;  // hours
```
leavesandlemmas, line 72: Is hour allowed to have a fractional part?

leavesandlemmas (approved):

src/module_library/format_time.h
eloch216: > I thought of an issue when calculating DOY this. If simulations are allowed to span calendar years, then computed this way, `doy` won't be correct outside the calendar year of `time = 0`.

I vote that we ignore this for now, since we don't usually run multi-year simulations right now. I think the long-term solution is probably to specify `time` as a UNIX timestamp or something similar, and then hopefully use a time library to correctly extract year, day of year, and hour from the timestamp.

PR #146, opened 2025-01-22
Example of Datetime functions

leavesandlemmas:
I created this pull request as a possible example of how we could implement better datetime in the R code of biocro. What this code does is create an S4 class for `time` that has a vector of hours and an origin year.  There are functions to convert from this type to other. This is a work in progress / prototype.  Might be better to implement this idea using R's S3 classes rather than time alone. 

Let me know what you think


eloch216:
Could you give a bit more description of how you would intend this to be used?

My main comment about the code itself is that I don't see why this requires a new class. I think `POSIXlt` and `POSIXct` already do everything that `biocro_time` does. Rather than introducing a new class, the getter/setter functions could be implemented as convenience functions for interacting with `POSIX` objects in convenient ways. For example, `get_day_of_year` could be something like

```R
# Define function
get_day_of_year <- function(posix_time) {
    as.numeric(format(posix_time, '%j'))
}

# Use today as an example
time <- Sys.time()
get_day_of_year(time)
```

Creating a new class can be a hassle, and can be a barrier for people who are not familiar with it. I know this through experience, having created a new class in the PhotoGEA R package (called `exdf`). For any desired functionality, if there's a way to achieve it without making a new class, I think that's usually the best option.

eloch216:
> it is functionally impossible for us to properly implement a time class. It is complex

There's an XKCD for that: https://xkcd.com/2867/

leavesandlemmas:
> Could you give a bit more description of how you would intend this to be used?

> Regarding the implementation itself, it is functionally impossible for us to properly implement a time class. It is complex, more so than what's here.

> This appears to be replicating what is already provided by the POSIX time classes. Why are those not suitable for whatever purpose this class in intended?

Regarding these points, the purpose was to facilitate the use of the POSIX time class. Especially when looking at BioCro's output.  But the POSIX time class currently can't be passed to BioCro directly.  And the conversion to POSIX requires specifying the origin / epoch year. I also wrote most of this before #144 when the output of `run_biocro` in #143  produced non-integer days for DOY. So basically it was intended to fix that and facilitate conversions to and from POSIX. 


This was also a trial run in specifying new generics for R objects, so it's not strictly speaking necessary to use the objects. I had written a lot of the code like this:

```r
to_datetime <- function(x, origin){
  origin_year <- paste0(origin, '-01-01')
  as.POSIXct(origin_year) + as.difftime(x, units='hours')
}

from_datetime <- function(x, origin){
  origin_year <- paste0(origin, '-01-01')
  difftime(x, origin_year, units='hours') |> as.numeric()
}
```
It's usually 


It can be tabled, I made this a draft pull request more as suggested code than a serious implementation.  However, I think the definition of `time` will need to change again in the future in order to accommodate simulations that span calendar years or more than one year.  Currently, `run_biocro` will raise an error if a user attempts to do either.  It seems to me that unless the c++ code is modified to use some C++ implementation similar to POSIX, the `time` variable will effectively become like such a class (apparently R's versions are based on / implemented using a C library, this might be relatively straightforward) . If I understand them correctly, datetime classes count time as an integer from some epoch / origin. For example R's POSIXct class counts the number of seconds as an integer from midnight Jan 1,1970  UTC, (the numpy library for python uses a 64 bit integer for the number of nanoseconds from the same date) . Then,  simply convert back and forth between that and a human readable format.  But that's effectively what we already do with `time`, we have code that allows users to specify the time as a DOY and an hour of day, then we convert that into `time` for biocro, then convert back afterward. It seems to me that the complexity in datetime classes comes from timezones and daylight savings time (which we mostly already ignore), maybe possibly leap seconds.  And even if we mostly wrap another library's datetime class, we then control the interface.

justinmcgrath:
People have done simulations across years in the past - willow growth for example. That's been handled by running multiple simulations, using the values at the end of one simulation as the starting values for the next one. That is cumbersome, so a different approach would be nice. However, no one is currently doing that work, and I know of no project with that sort of work on the horizon. I am not seeing other benefits that would help what we're doing now, so based on that, this would not be a priority.

Cross-year simulations would also require substantial changes to the modules themselves to simulate the winter period. I suspect that even with the ability to use time across years, the work involved in creating suitable models wouldn't be of interest itself, and people would simulate separate years anyway. So I'd also like to see that someone has a demonstrated desire for this before pursuing it.

justinmcgrath:
And regarding the definition of time, the framework only imposes that the value be numeric, and thus includes the POSIX definition. The specific definition comes from the modules. If the modules are written to handle POSIX time, that can easily be based by the framework using `as.numeric(as.POSIXct(...))`. Getting that to work requires changing the modules, but nothing about the framework would need to change. And we wouldn't want the framework to impose POSIX time, because that typically is not necessary and would create a burdensome requirement for models that don't need it.

leavesandlemmas:
Remove this branch & pull request?

eloch216:
> Remove this branch & pull request?

Sounds good to me

eloch216:
This PR is closed, so I will delete the corresponding branch.

PR #141, opened 2025-01-08
Add PR advice

eloch216:
This PR adds a new section to the developer documentation that describes a few conventions we tend to follow when making and reviewing PRs. Hopefully this will be useful for people who are getting started with contributing to BioCro.

@leavesandlemmas As the newest contributor, it would be great to get your thoughts on this -- do you think it is clear and helpful?

PR #148, opened 2025-01-27
Add more advice about building vignettes

eloch216:
While following my own advice from PR https://github.com/biocro/biocro/pull/141, I realized that the suggested workflow was missing a step.

Because of CRAN package size requirements, we had to move most of our vignettes to `vignettes/web_only`, which is listed in `.Rbuildignore`. These vignettes are not built when running `R CMD check`, but are included when `pkgdown` builds the BioCro documentation web site.

So, even when `R CMD check` passes locally, there is still a possible surprise when the `pkgdown` site is built during the GitHub checks.

I updated some of the developer documentation to reflect this. There is now a new step in the suggested PR workflow. I also rearranged the "Building the vignettes" chapter because the main method listed (`R CMD build`) actually only builds one vignette, skipping most of them.

### DIFF
--- a/developer_documentation/contribution_guidelines.Rmd
+++ b/developer_documentation/contribution_guidelines.Rmd
@@ -206,10 +206,10 @@ found to be helpful.
 * **Submitted code should work:** It takes time and effort to produce a
   thoughtful review of a pull request, so requests to review a PR should
   generally only be made for working code. In a BioCro context, "working" code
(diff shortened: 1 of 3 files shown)


leavesandlemmas (approved):
Looks good to me! I don't think I have any comments

justinmcgrath (approved):

PR #150, opened 2025-01-30
Refactoring the biomass conversion factor from the photosynthesis modules

leavesandlemmas:
Currently, BioCro converts carbon fluxes such as photosynthesis (or respiration) to changes in biomass by assuming a conversion factor (biomass is a fixed ratio of the amount of carbon in the plant). The modules which compute the `canopy_assimilation_rate` return a value with dimensions mass / area / time.  Moreover this conversion factor is hard-coded into these modules, and not easily treated as a parameter. However, the conversion means that effectively the carbon content of any crop in BioCro is fixed to 40% C by mass.  

The development of future models which add non-structural carbohydrates or nutrient components would require modifications to how biomass is calculated for each organ. To maintain the utility of the photosynthesis modules, it would be useful to access the canopy assimilation rate measured in moles of CO2.  

This pull request modifies the canopy photosynthesis modules so that they no longer perform a biomass conversion (they still convert transpiration from moles of water to mass of water). Instead the photosynthesis modules return `canopy_assimilation_rate_mol_CO2`  and `canopy_photorespiration_rate_mol_CO2` and `GrossAssim_mol_CO2`.  A new direct module will convert these fluxes into changes in biomass.  The new module also allows the conversion factor to be modified. To obtain the previous behavior, set `CHO_carbon_molar_mass = 30.026` g/mol .  FWIW the typical carbon molar mass in plant biomass is closeer to 27 g/mol (or 45% C by mass).  The logic of the conversion is described in the documentation of this module and I have copied that below. 

The new module also performs the area and time unit conversions, and perhaps it shouldn't. Maybe the multilayer integrator modules should return `canopy_assimilation_rate` in mol CO2 / Ha / hour rather than micromol CO2 / m^2 / s. Thoughts?

```
 * Models of photosynthesis compute the rate of CO2 assimilation.
 * However, many BioCro modules require the rate at which photosynthesis
 * increases plant biomass. Unless a model computes different components
 * of plant biomass explicitly, the calculation is assumed to be a simple
 * conversion factor.
 *
 * The conversion is based on the fact that plant
 * biomass is primarily composed of carbohydrates and the ratio of carbon
 * to oxygen and hydrogen is nearly the same for all carbohydrates. Thus,
 * given a number of carbon atoms (i.e., moles of carbon), we can estimate
 * the carbon-hydrogen-oxygen mass (the CHO mass) which will account for
 * ~95% of plant biomass. We call this ratio the carbon molar mass.
 *
 * (carbon molar mass) = (molar mass) / (carbon atoms per mol)
 *
 * For example, glucose has a molar mass of 180 g/mol and 6 carbon atoms.
 * So every mol C contributes 12 g of C + 2 g H + 16 g O = 30 g CHO mass.
 * Or (180 g / mol glucose) / (6 mol C / mol glucose) = 30 g CHO / mol C
 * All monosaccharides have this value. All disaccharides are 28.5 g/mol.
 * Starch, cellulose, and hemi-celluloses have a value of 27 g/mol.
 *
 * Any mixture of these substances will have a carbon molar mass between
 * 27 and 30 g/mol.
 *
 * Thus, given a carbon molar mass, this module will convert
 * the CO2 assimilation rate in micromol C / m^2 land / s to a biomass
 * accumulation rate in Mg / Ha / hr.
 *
 * To convert micromol / m^2 / s into Mg / ha / hour, we also must
 * convert the units of time and land area as well as to shift the SI
 * prefixes. The following factor gives the correct conversion:
 * (carbon molar mass) * (3600 s / hr) * (1e-6 mol / umol) *
 * (1e-3 Mg / kg) * (1e4 m^2 / ha)
 * = (carbon molar mass) * 3.6e-5 (s * mol * Mg * m^2) / (hr * micromol * g * ha)
 *
 * Note that if other elements are assumed to be in constant ratio with
 * carbon, then this conversion factor gives total biomass, not just
 * CHO mass.
 *
 * For instance, the amino acid glutamate has a carbon molar mass of 29.4
 * g / mol C. But every 5 mol C contributes 1 mol N. However, the carbon
 * molar mass varies more for amino acids than for carbohydrates.
 * Proteinogenic amino acids vary from 18 - 41 g / mol C.
 *
 * The carbon molar mass of a biomass sample can be measured by identifying
 * the number of carbon moles contained in the sample:
 *
 * (carbon molar mass) = (biomass)/(# mol C)
 *
 * Given a C mass fraction, the number of mol C in the sample is:
 *
 * (# mol C) = (biomass) * (C mass fraction) / (molar mass of carbon)
 *
 * Substituting yields:
 *
 * (carbon molar mass) = (molar mass of carbon) / (C mass fraction)
 *
```


yufenghe22 (approved):
This is a good idea to link to future non-structural carbon modules. Please make sure the changed version generates identical results or very close to the previous version. For example, check some soybean biomasses between the old and new. I think the `tests` can do this, but since there are new variables introduced to the test files, it seems difficult to check whether the changes on those test files are due to additions of variables or changes of values.

eloch216:
I will take a look at the code itself soon, but in the meantime I have a few immediate comments:

> This pull request modifies the canopy photosynthesis modules so that they no longer perform a biomass conversion (they still convert transpiration from moles of water to mass of water). Instead the photosynthesis modules return `canopy_assimilation_rate_mol_CO2` and `canopy_photorespiration_rate_mol_CO2` and `GrossAssim_mol_CO2`.

Usually we avoid mentioning units in quantity names. With that convention, these should probably just be called `canopy_assimilation_rate`, `canopy_photorespiration_rate` and `GrossAssim`.

> The new module also performs the area and time unit conversions, and perhaps it shouldn't. Maybe the multilayer integrator modules should return `canopy_assimilation_rate` in mol CO2 / Ha / hour rather than micromol CO2 / m^2 / s. Thoughts?

I think there's no "correct" approach here. Personally, I like `micromol / m^2 / s` because it is more directly comparable to canopy gas exchange measurements (for example, https://doi.org/10.1016/j.agrformet.2023.109532 or https://doi.org/10.1016/j.agrformet.2023.109532). It's unlikely that an experimentalist would report an assimilation rate using hectares for area and hours for time.

On the other hand, rates from BioCro differential modules are all in "per hour," so I can see a potential benefit of switching to `mol / Ha / hr` for consistency.

Maybe @justinmcgrath has an opinion here?

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,138 @@
...
+
+
+
+    constexpr double sec_per_hour = 3600;
+    constexpr double mol_per_micromol = 1e-6;
+    constexpr double square_meters_per_hectare = 1e4;
+    constexpr double megagram_per_gram = 1e-6;
+    constexpr double a = sec_per_hour * mol_per_micromol * square_meters_per_hectare * megagram_per_gram;
+
+    const double cf = CHO_molar_mass_per_mol_C * a;
```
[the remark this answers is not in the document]
eloch216: >     * Shall we remove `molar_mass_of_glucose` in the `constants.h`?

There could still potentially be other uses for the `molar_mass_of_glucose` constant, so I don't think there's a pressing need to remove it. Plus, `constants.h` is in the `framework` git submodule, so modifying it is more complicated.

eloch216:
Please run `clang-format` on the following files:
- `src/module_library/CanAC.cpp`
- `src/module_library/c3CanAC.cpp`
- `src/module_library/c3_canopy.cpp`
- `src/module_library/c4_canopy.h`
- `src/module_library/carbon_assimilation_to_biomass.h`
- `src/module_library/multilayer_canopy_integrator.h`

I can tell it has not been run on these files because there are successive lines where the comment characters (`//`) are not lined up.

eloch216:
I really like the documentation for the new module. I think it's very helpful.

One comment is that in the resulting Doxygen page, the equations and variable names are not formatted different from the regular text, so they aren't so easy to pick out (see screenshot). It's possible to include some LaTeX style markup in Doxygen comments to make them stand out more. Here's an example of a module that uses different formatting for variable names and equations: https://biocro.github.io/BioCro-documentation/latest/doxygen/doxygen_docs_modules_public_members_only/classstandard_b_m_l_1_1soil__sunlight.html

You don't need to use fancier formatting, but I just wanted to bring it up as a possibility.

Screenshot 2025-01-30 120108

leavesandlemmas:
> Usually we avoid mentioning units in quantity names. With that convention, these should probably just be called `canopy_assimilation_rate`, `canopy_photorespiration_rate` and `GrossAssim`

We're going to have to pick different names because those variables because the module I added to convert biomass has those variables as the outputs. Thoughts?  I haven't thought of anything good 






> I really like the documentation for the new module. I think it's very helpful.

Thank you

> One comment is that in the resulting Doxygen page, the equations and variable names are not formatted different from the regular text, so they aren't so easy to pick out (see screenshot). It's possible to include some LaTeX style markup in Doxygen comments to make them stand out more. Here's an example of a module that uses different formatting for variable names and equations: https://biocro.github.io/BioCro-documentation/latest/doxygen/doxygen_docs_modules_public_members_only/classstandard_b_m_l_1_1soil__sunlight.html
> 
> You don't _need_ to use fancier formatting, but I just wanted to bring it up as a possibility.

I am more than happy to add latex formatted math. Just hadn;t looked into how to.

eloch216:
> > Usually we avoid mentioning units in quantity names. With that convention, these should probably just be called `canopy_assimilation_rate`, `canopy_photorespiration_rate` and `GrossAssim`
> 
> We're going to have to pick different names because those variables because the module I added to convert biomass has those variables as the outputs. Thoughts? I haven't thought of anything good

Oh, I didn't notice that, but I now I understand.

I have run into this problem before. See, for example, the `energy_balance_outputs` struct, defined in `src/module_library/leaf_energy_balance.h`. It has two members called `boundary_layer_conductance` and `gbw`. These are the same thing expressed in different units. In hindsight, that naming scheme is confusing, and in fact, it did cause a mistake in the "radiation use efficiency" module because I used the wrong struct member in one spot. See this PR: https://github.com/biocro/biocro/pull/151

I would argue that in the `energy_balance_outputs` and in this module, it's not really a difference in units, but a difference in physical quantities. A unit difference could be something like seconds vs. hours, or meters vs. feet, because these are still measuring the same kind of thing (distance or time). But a rate of "molecules per area per time" is really a different thing than a rate of "mass per area per time."

Anyway, this is a long-winded way of saying that I would be okay with including the "kind of thing" in the quantity name, but not the units. So, for example, we could distinguish between `canopy_assimilation_rate_molecules` and `canopy_assimilation_rate_mass`, but I wouldn't like `canopy_assimilation_rate_mol` or `canopy_assimilation_rate_Mg` as much. (For `gbw`, it's a difference between a conductance for mass fluxes (units are `m / s`) and a conductance for molecular fluxes (units are `micromol / m^2 / s`).)
 
> > One comment is that in the resulting Doxygen page, the equations and variable names are not formatted different from the regular text, so they aren't so easy to pick out (see screenshot). It's possible to include some LaTeX style markup in Doxygen comments to make them stand out more. Here's an example of a module that uses different formatting for variable names and equations: https://biocro.github.io/BioCro-documentation/latest/doxygen/doxygen_docs_modules_public_members_only/classstandard_b_m_l_1_1soil__sunlight.html
> > You don't _need_ to use fancier formatting, but I just wanted to bring it up as a possibility.
> 
> I am more than happy to add latex formatted math. Just hadn;t looked into how to.

There's another example here that has a bit more stuff: https://github.com/biocro/biocro/blob/develop/src/module_library/conductance_limited_assim.h

leavesandlemmas:
> Anyway, this is a long-winded way of saying that I would be okay with including the "kind of thing" in the quantity name, but not the units. So, for example, we could distinguish between `canopy_assimilation_rate_molecules` and `canopy_assimilation_rate_mass`, but I wouldn't like `canopy_assimilation_rate_mol` or `canopy_assimilation_rate_Mg` as much. (For `gbw`, it's a difference between a conductance for mass fluxes (units are `m / s`) and a conductance for molecular fluxes (units are `micromol / m^2 / s`).)

But molecules would also specify a different unit than moles, since it is possible to specify the number of molecules of CO2  assimilated per leaf area per time, although they have the same unit dimensions.  Some other possible names are 

+ `canopy_assimilation_molar_flux` since its a flux in moles. 
+ `canopy_molar_assimilation` (though this sounds like it should be per mole). 
+ `canopy_assimilation_CO2`  because it has dimensions # of CO2 per area per time.
 

I would prefer to leave `canopy_assimilation_rate` with its old meaning / units for compatibility reasons.

eloch216:
> > Anyway, this is a long-winded way of saying that I would be okay with including the "kind of thing" in the quantity name, but not the units. So, for example, we could distinguish between `canopy_assimilation_rate_molecules` and `canopy_assimilation_rate_mass`, but I wouldn't like `canopy_assimilation_rate_mol` or `canopy_assimilation_rate_Mg` as much. (For `gbw`, it's a difference between a conductance for mass fluxes (units are `m / s`) and a conductance for molecular fluxes (units are `micromol / m^2 / s`).)
> 
> But molecules would also specify a different unit than moles, since it is possible to specify the number of molecules of CO2 assimilated per leaf area per time, although they have the same unit dimensions. Some other possible names are
> 
>     * `canopy_assimilation_molar_flux` since its a flux in moles.
> 
>     * `canopy_molar_assimilation` (though this sounds like it should be per mole).
> 
>     * `canopy_assimilation_CO2`  because it has dimensions # of CO2 per area per time.

I like `canopy_assimilation_CO2`. The other two still sound like units to me.

> I would prefer to leave `canopy_assimilation_rate` with its old meaning / units for compatibility reasons.

This sounds good to me.

NEWS.md
```diff
@@ -122,6 +122,14 @@ be directly added to this file to describe the related changes.
 - Added a new function for generating C++ header files for new module classes:
   `module_write`
 
+- The conversion of CO2 assimilation to biomass is no longer hard coded into
+  the photosynthesis modules: c3CanAC, CanAC, multilayer_canopy_integrator, etc.
+  These modules now produce canopy assimilation rates in micromol CO2 / m^2 / s
+  A new module called `biocro:carbon_assimilation_to_biomass` now performs the
```
eloch216, line 128: The module is called `BioCro:carbon_assimilation_to_biomass` (names are sensitive to capitalization)

src/module_library/c3_canopy.cpp
```diff
@@ -78,11 +78,11 @@ string_vector c3_canopy::get_inputs()
 string_vector c3_canopy::get_outputs()
 {
     return {
-        "canopy_assimilation_rate",     // Mg / ha / hr
-        "canopy_transpiration_rate",    // Mg / ha / hr
-        "canopy_conductance",           // mol / m^2 / s
-        "GrossAssim",                   // Mg / ha / hr
-        "canopy_photorespiration_rate"  // Mg / ha / hr
+        "canopy_assimilation_rate_CO2",     // Mg / ha / hr
+        "canopy_transpiration_rate",            // Mg / ha / hr
```
eloch216, line 82: This file needs `clang-format`

src/module_library/c4_canopy.h
```diff
@@ -196,11 +196,11 @@ string_vector c4_canopy::get_inputs()
 string_vector c4_canopy::get_outputs()
 {
     return {
-        "canopy_assimilation_rate",     // Mg / ha / hr
-        "canopy_transpiration_rate",    // Mg / ha / hr
-        "canopy_conductance",           // mol / m^2 / s
-        "GrossAssim",                   // Mg / ha / hr
-        "canopy_photorespiration_rate"  // Mg / ha / hr
+        "canopy_assimilation_rate_CO2",     // micromol CO2 / m^2 / s
+        "canopy_transpiration_rate",            // Mg / ha / hr
```
eloch216, line 200: This file needs `clang-format`

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,155 @@
...
+    void do_operation() const;
+};
+
+string_vector carbon_assimilation_to_biomass::get_inputs()
+{
+    return {
+        "canopy_assimilation_rate_CO2",      // micromol CO2 / m^2 / s
+        "GrossAssim_CO2",                    // micromol CO2 / m^2 / s
+        "canopy_photorespiration_rate_CO2",  // micromol CO2 / m^2 / s
+        "CHO_carbon_molar_mass"                  // g CHO / mol C
```
eloch216, line 123: This file needs `clang-format`

src/module_library/multilayer_canopy_integrator.h
```diff
@@ -137,11 +137,11 @@ string_vector multilayer_canopy_integrator::get_inputs(int nlayers)
 string_vector multilayer_canopy_integrator::get_outputs(int /*nlayers*/)
 {
     return {
-        "canopy_assimilation_rate",     // Mg / ha / hr
-        "canopy_transpiration_rate",    // Mg / ha / hr
-        "canopy_conductance",           // mmol / m^2 / s
-        "GrossAssim",                   // Mg / ha / hr
-        "canopy_photorespiration_rate"  // Mg / ha / hr
+        "canopy_assimilation_rate_CO2",     // micromol CO2 / m^2 / s
+        "canopy_transpiration_rate",            // Mg / ha / hr
```
eloch216, line 141: This file needs `clang-format`

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * \f[
+ *  \kappa = \frac{\mu}{(\text{carbon atoms per mol})}
+ * \f]
+ *
+ * Where \f$\mu\f$ is the molar mass.
+ *
+ * For example, glucose has a molar mass of 180 g/mol and 6 carbon atoms.
+ * So every mol C contributes 12 g of C + 2 g H + 16 g O = 30 g CHO mass.
+ * Or (180 g / mol glucose) / (6 mol C / mol glucose) = 30 g CHO / mol C
```
eloch216, line 33: I think there should be a period after `30 g CHO / mol C`.

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,155 @@
...
+ * increases plant biomass. Unless a model computes different components
+ * of plant biomass explicitly, the calculation is assumed to be a simple
+ * conversion factor.
+ *
+ * The conversion is based on the fact that plant
+ * biomass is primarily composed of carbohydrates and the ratio of carbon
+ * to oxygen and hydrogen is nearly the same for all carbohydrates. Thus,
+ * given a number of carbon atoms (i.e., moles of carbon), we can estimate
+ * the carbon-hydrogen-oxygen mass (the CHO mass) which will account for
+ * ~95% of plant biomass. We call this ratio \f$\kappa\f$ the carbon molar mass.
```
eloch216, line 23: Is the "carbon molar mass" a standard term in the literature? If so, is there a textbook or paper that could be cited?

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,155 @@
...
+ *  \kappa = \frac{\mu}{(\text{carbon atoms per mol})}
+ * \f]
+ *
+ * Where \f$\mu\f$ is the molar mass.
+ *
+ * For example, glucose has a molar mass of 180 g/mol and 6 carbon atoms.
+ * So every mol C contributes 12 g of C + 2 g H + 16 g O = 30 g CHO mass.
+ * Or (180 g / mol glucose) / (6 mol C / mol glucose) = 30 g CHO / mol C
+ * All monosaccharides have this value. All disaccharides are 28.5 g/mol.
+ * Starch, cellulose, and hemi-celluloses have a value of 27 g/mol.
```
eloch216, line 35: Are there sources to cite for these values? I'm not doubting them, I just think it could be helpful to provide some references for anyone who wants to learn more.

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,155 @@
...
+ * = (carbon molar mass) * 3.6e-5 (s * mol * Mg * m^2) / (hr * micromol * g * ha)
+ *
+ * Note that if other elements are assumed to be in constant ratio with
+ * carbon, then this conversion factor gives total biomass, not just
+ * CHO mass.
+ *
+ * For instance, the amino acid glutamate has a carbon molar mass of 29.4
+ * g / mol C. But every 5 mol C contributes 1 mol N. However, the carbon
+ * molar mass varies more for amino acids than for carbohydrates.
+ * Proteinogenic amino acids vary from 18 - 41 g / mol C.
```
eloch216, line 59: Is there a source for the amino acid carbon molar mass values?

eloch216:
This is looking pretty good! Just a few minor style comments.

Besides the ones I made above as single comments, I think it could also be helpful to break up the documentation for the new module into sections, since it covers a few different topics. I would say the documentation has general info about carbon molar masses, info about converting time and area units, and info about calculating values of the carbon molar mass. It could be broken into sections following this categorization, or another one if you have something else in mind. Sections can be separated by horizontal lines (see the documentation for `RHprof()` in `src/module_library/AuxBioCro.cpp`) and they can be given bold names (see the documentation for the `solar_position_michalsky` module).

src/module_library/carbon_assimilation_to_biomass.h
leavesandlemmas: No, as far as I know I made it up.  I made some changes to clarify it should be the "carbon molar mass of  x". I couldn't find anything referring to this quantity by this name.  Googling carbon molar mass only returns the molar mass of carbon. This is a related quantity. 

We could rename it to something else. E.g. 
+ `CHO_molar_mass_per_carbon`
+ `CHO_per_carbon_molar_mass`
+ `CHO_molar_mass_per_mole_carbon`.
But I couldn't think of anything that's shorter, descriptive, and doesn't violate our convention on no unit names in the quantity's name.  The conversion between mass and moles is normally called a molar mass.

src/module_library/carbon_assimilation_to_biomass.h
leavesandlemmas: For the molar mass of glucose ? Or the number of carbon atoms in glucose ?  Or the other carbohydrates?

The `constants.h` file doesn't give a citation for the molar masses of water or glucose, either. But we could cite pubchem I guess. https://pubchem.ncbi.nlm.nih.gov/compound/D-Glucose  Or maybe IUPAC for the periodic table of elements.

src/module_library/carbon_assimilation_to_biomass.h
leavesandlemmas: I computed these from their molar masses.

src/module_library/carbon_assimilation_to_biomass.h
eloch216: The discussion about glucose doesn't need any more clarification. I meant for these statements:
- All monosaccharides have this value.
- All disaccharides are 28.5 g/mol.
- Starch, cellulose, and hemi-celluloses have a value of 27 g/mol.

I think they are not immediately obvious to people who don't think about carbohydrate chemistry very often. (Me!) You don't have to cite something for the molecular masses or chemical formulas. But maybe just clarify how you got these values? E.g. "Using lactose and cellulose as representative examples, we can also determine that all disaccharides and polysaccharides have..." (or whatever best describes how you did this).

src/module_library/carbon_assimilation_to_biomass.h
eloch216: > I computed these from their molar masses.

You could just say that you computed them. Is there a link to a wiki page about proteinogenic amino acids with their masses and chemical formulas?

src/module_library/carbon_assimilation_to_biomass.h
leavesandlemmas: I added some extra text describing the calculation more. And actually I was wrong to say monosaccharide. Rather I shouldve said all hexoses. Apparently pentoses, trisoes, etc. are also considered monosaccharides.  

They all have the same value because they all have the same molecular formula by definition. A hexose is C6H12O6 and a disaccharide of two hexoses is C12H22O11 (the glycosidic bond cleaves off a H2O so every time a hexose is connected to another there's one less O and two less H). 

I have seen a lot of papers do this calculation (or versions of it) but they just don't draw a box around it and give it a name. I have a table in a google sheet where I did most of the calculations last summer.

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,198 @@
...
+
+#include "../framework/module.h"
+#include "../framework/state_map.h"
+
+namespace standardBML
+{
+/**
+ * @class carbon_assimilation_to_biomass
+ *
+ * @brief
```
eloch216, line 12: Now the brief description is empty. It should be a single short sentence (not even necessarily a complete sentence), such as "Converts the canopy assimilation rate from a flux of CO2 molecules (micromol / m^2 / s) to a rate of dry biomass accumulation (Mg / ha / hr)."

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * For example, glucose has a molar mass of 180 g/mol and 6 carbon atoms,
+ * assuming isotopes 1-H, 12-C, 16-O. The molar mass varies with the ratio of
+ * isotopes, but the average isotopic ratio is close to the same molar mass.
+ * So every mol C contributes 12 g of C + 2 g H + 16 g O = 30 g CHO mass.
+ * Or (180 g / mol glucose) / (6 mol C / mol glucose) = 30 g CHO / mol C.
+ * All hexoses (e.g., glucose, fructose, galactose) all have the same molecular
+ * formula C6H12O6, and thus the same molar mass (180 g / mol) and all consist
+ * of 6 carbons, again ignoring any variation in isotope distribution.
+ * Thus, they all have a carbon molar mass of about 30 g / mol this value.
```
eloch216, line 48: "Thus, they all have a carbon molar mass of about 30 g / mol this value." -> "Thus, they all have a carbon molar mass of about 30 g / mol."

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,198 @@
...
+ * Disaccharides consisting of two hexose unitssuch as sucrose, maltose, and
```
eloch216, line 49: "unitssuch" -> "units such"

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,198 @@
...
+ * (1e-3 Mg / kg) * (1e4 m^2 / ha)
+ * = (carbon molar mass) * 3.6e-5 (s * mol * Mg * m^2) / (hr * micromol * g * ha)
+ *
+ *
+ * ### Generalization of CHO carbon molar mass
+ *
+ * Note that if other elements are assumed to be in constant ratio with
+ * carbon, then this conversion factor gives total biomass, not just
+ * CHO mass. The `CHO_carbon_molar_mass` is the amount of dry biomass (grams)
+ * per carbon (mol C).
```
eloch216, line 89: > Note that if other elements are assumed to be in constant ratio with carbon, then this conversion factor gives total biomass, not just CHO mass. The `CHO_carbon_molar_mass` is the amount of dry biomass (grams) per carbon (mol C).

This seems to suggest that this quantity should be called something else such as `biomass_per_assimilated_CO2` (or something like that), since it is not necessarily the mass of CHO per C.

src/module_library/carbon_assimilation_to_biomass.h
```diff
@@ -0,0 +1,198 @@
...
+ * Proteinogenic amino acids vary from 18 - 41 g / mol C.
+ *
+ * Mixtures of amino acids and carbohydrates will have a carbon molar mass
+ * in this range, but given that carbohydrates are the largest fraction of plant biomass
+ * the carbon molar mass of plant biomass will be close to 27 g / mol C as this
+ * is the carbon molar mass of starch, cellulose, and hemi-celluloses.
+ *
+ * ### Measuring the CHO carbon molar mass of a tissue sample.
+ *
+ * The carbon molar mass \f$\kappa\f$ of a biomass sample of mass \f$m\f$ can be measured by identifying
```
eloch216, line 103: This line, and a few others in the documentation, are longer than 80 characters

eloch216 (approved):
I have no more comments about this. @leavesandlemmas you can merge when you're ready.

justinmcgrath:
Thanks for doing this. Here are some late comments.

>> The new module also performs the area and time unit conversions, and perhaps it shouldn't. Maybe the multilayer integrator modules should return canopy_assimilation_rate in mol CO2 / Ha / hour rather than micromol CO2 / m^2 / s. Thoughts?

> Maybe @justinmcgrath has an opinion here?

Given the generally messy state of units and the big undertaking to change that, it doesn't make much difference. micromol / m^2 / s is probably more familiar for most people.

> Anyway, this is a long-winded way of saying that I would be okay with including the "kind of thing" in the quantity name, but not the units.

My preference is that the quantity name would describe the kind of thing, and that's pretty common: height, mass, etc. Some of the compound things have conventional names (mass fraction, irradiance) that we use, but for many of the rates we don't often do that or maybe there aren't standard names. My general approach to this is that good quantity names specify the dimensions: carbon mass, wind velocity. The dimension "amount of substance" is difficult for this, because no one is going to write that, and "amount" alone is easily mistaken for a broader meaning.

Although I don't see other people using "molar flux" I immediately understood what it meant, and that seems fine to distinguish if from mass flux, which is a commonly used phrase. As for things like "CO2" as a unit, SI convention is that these are not units, and the name of the quantity indicates the type of molecule -- CO2 mass flux, CO2 molar flux, etc. I get their point and usually I try to adhere to that, but I can see cases where it gives less than it takes. Here though, adding CO2 doesn't clarify anything for me about what makes these two different. They're both CO2 fluxes. I don't have a preference whether this gets changed, but I figured I'd share my thoughts on it.

PR #151, opened 2025-01-30
Fix RUE

eloch216:
Apparently I introduced a typo into `rue_leaf_photosynthesis.cpp` as part of this PR: https://github.com/biocro/biocro/pull/109

This typo caused the `ten_layer_rue_canopy` to predict very small values of `canopy_assimilation_rate` on the `develop` branch. In turn, these small assimilation rates caused the optimization in the "Quantitative Comparison Between Two Photosynthesis Models" to fail. In that vignette, the issue is that the optimal value of `alpha_rue` had become extremely large (>1, which doesn't make sense physically). This all took a while to figure out, but fortunately the solution was very simple.

This PR fixes the typo and re-adjusts a few axis limits in the vignette. Since Justin is busy and this is pretty straightforward, I will just merge it when I have confirmed that the checks are all passing.


eloch216:
Following a comment on another PR (https://github.com/biocro/biocro/pull/150#issuecomment-2628027809), I tried to clarify the differences between the two different units used for boundary layer conductance in the output from energy balance calculations. Most of the instances of `gbw` refer to "mass conductances" in `m / s`. I renamed the one exception to `gbw_molecular`, since this is a "molecular conductance" in `mol / m^2 / s`.

Conductances used throughout BioCro do not always follow this convention, but I still think this is a little bit of progress.

PR #152, opened 2025-02-01
Add model testing

eloch216:
This PR moves the crop model testing functions from `tests/testthat/crop_model_testing_helper_functions.R` into the package namespace. These functions will be useful to anyone working on BioCro models in other repositories, such as the nitrogen model or the cowpea model. It would be nice to have these functions in the next BioCro release.


eloch216:
This PR is now ready to be evaluated

leavesandlemmas (approved):
All looks fine to me! It's a good idea to make testing easier.

Would it be possible to change `testthat.r` so that it can be run from the same directory as `R CMD INSTALL BioCro` ?  It would be a QOL improvement to be able to run tests from commandline immediately after installing without having to change directories.

eloch216:
> Would it be possible to change `testthat.r` so that it can be run from the same directory as `R CMD INSTALL BioCro` ? It would be a QOL improvement to be able to run tests from commandline immediately after installing without having to change directories.

I'm not sure about that. I think the contents and location of `testthat.R` are determined by the `testthat` package. The documentation for `testthat::test_check` says:

```
To run testthat automatically from ⁠R CMD check⁠, make sure you have a tests/testthat.R that contains:

library(testthat)
library(yourpackage)

test_check("yourpackage")
```

I haven't tried changing it, but it seems like altering `testthat.R` might mess up `R CMD check`.

My solution to this issue is to have a text file with some common commands so I can just copy/paste when needed:

```r
remove.packages('BioCro')
install.packages('biocro', repos=NULL, type='SOURCE')

library(BioCro)
library(testthat)
test_dir('biocro/tests/testthat')
```

I run these from a fresh R session in the parent directory of `biocro` to install the package and run the tests.

eloch216:
I am planning to merge this PR on Tuesday, February 11 unless I hear from someone before then.

yufenghe22 (approved):
I've been constantly using `crop_model_testing_helper_functions.R` for commiting changes to modules. Adding it to namespace makes this process easier.

PR #154, opened 2025-02-11
Remove time inputs

eloch216:
In this PR, any modules that use `time` as an input are changed to use other time-related quantities instead, such as `doy`, `hour`, or `fractional_doy`. This is more clear than using `time`. See https://github.com/biocro/biocro/pull/137#issuecomment-2625332426

I decided to just remove the `BioCro:fake_solar` module altogether since I am not using it anymore. While doing that, I realized it would be helpful to check the module test case directory to make sure there aren't any "extra" test cases for modules that aren't in the library. I modified `test_module_library` to do this, and this feature turned out to be useful, since it identified a test case file for a module that no longer exists (`BioCro:soil_type_selector`).


leavesandlemmas (approved):
Seems fine to me!

PR #155, opened 2025-02-11
Update version of ubuntu used for testing

eloch216:
I got an email from GitHub about the Ubuntu-20.04 runner image closing down:

```
The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025. To raise awareness of the upcoming removal, we will temporarily fail jobs using Ubuntu 20.04. Builds that are scheduled to run during the brownout periods will fail. The brownouts are scheduled for the following dates and times:

    March 4 14:00 UTC – 22:00 UTC
    March 11 13:00 UTC – 21:00 UTC
    March 18 13:00 UTC – 21:00 UTC
    March 25 13:00 UTC – 21:00 UTC

What you need to do

Jobs using the ubuntu-20.04 YAML workflow label should be updated to ubuntu-22.04, ubuntu-24.04 or ubuntu-latest.
```

In this PR, I have changed all instances of ubuntu-20.04 to ubuntu-22.04. This seemed like the smallest change to our current setup. I could also try ubuntu-24.04 or ubuntu-latest if people think that might be better.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -92,7 +92,7 @@ jobs:
           # [
           #   {
           #     "configName": "R 3.6.0 on Ubuntu",
-          #     "os": "ubuntu-20.04",
+          #     "os": "ubuntu-22.04",
           #     "r": "3.6.0"
           #   }
           # ],
@@ -101,7 +101,7 @@ jobs:
           #     [
           #       {
           #         "configName": "R 3.6.0 on Ubuntu",
-          #         "os": "ubuntu-20.04",
+          #         "os": "ubuntu-22.04",
           #         "r": "3.6.0"
           #       }
           #     ]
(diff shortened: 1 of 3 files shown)


eloch216:
> I don't know what support we want to try to maintain overall, but the Linux computer will have 24.04 now so I'd like to at least check for that. Could someone change the workflow at some point? Thank you.
>
> Cheers,
> Justin

I changed it to 22.04

justinmcgrath:
> I changed it to 22.04

By the time you wrote this comment, you had changed it to 24.04. Is 22.04 a typo here?

I think checking an LTS release is sensible rather than the latest. I don't know whether it particularly matters whether that's 22.04 or 24.04, except if we use 24.04 then we won't need to do anything with this again for at least 3 years.

justinmcgrath (approved):

eloch216:
> > I changed it to 22.04
> 
> By the time you wrote this comment, you had changed it to 24.04. Is 22.04 a typo here?
> 
> I think checking an LTS release is sensible rather than checking the latest. I don't know whether it particularly matters whether that's 22.04 or 24.04, except if we use 24.04 then we won't need to do anything with this again for at least 3 years.

Whoops, yes, that was a typo!

Issue #161, opened 2025-03-05
Fixed Point Iteration in the C3 model in BioCro


leavesandlemmas:
Hi everyone,

I am trying to understand the fixed point iteration used here: https://github.com/biocro/biocro/blob/31a414542d64513ba0c5425ee2b72fe651670fe5/src/module_library/c3photo.cpp#L120 

I would appreciate it if someone would confirm my thinking here: @eloch216  or @justinmcgrath  ? No rush though. 

### What equation does the fixed point iteration solves? 

What's the equation that's being solved exactly?  If I understand correctly, we have this fundamental equation for the intercellular CO2 $C_i$ : 

$$ \dot{C}_i = g_s (C_a - C_i ) - \tilde{V}_c + R_d $$

where $C_a$ is the atmospheric CO2 outside the leaf. The $\tilde{V}_c = V_c - V_o/2$ is the reduced carboxylation rate ( carboxylation reduced by the photorespiration rate). And $R_d$ is the leaf respiration rate. This equation contracts to a steady state within a minutes (at least that's the hypothesis of our photosynthesis models), so that 

$$ 0 = g_s (C_a - C_i ) - \tilde{V}_c + R_d \implies A_n = g_s (C_a - C_i )  = \tilde{V}_c - R_d  $$
 
The Ball-Berry stomata conductance model holds that $g_s$  satisfies:

$$ g_s = b_0 + b_1 \frac{A_n h_s }{C_a }  $$

And the FcVB model computes $\tilde{V}_c  - R_d$ from $C_i$.  It seems like there's another detail with $C_s$, the leaf surface CO2 concentration.  So maybe this isn't quite right. What's the purpose of lines 98-99 ? What process does this correspond to ?

It seems like from the comment (lines 93-97) that this isn't a fundamental process, but rather an addition to prevent numerical problems from negative Ci. Is that correct?

### The fixed point iteration

So the fixed point iteration starts with guesses for $g_s$ and $C_i$, computes $g_s (C_a - C_i )$ (line 98) and $\tilde{V}_c - R_d$  (line 101) then takes the minimum of the two (line 115). Using this estimate of $A_n$, the method then computes the corresponding value of $g_s$ under the ball berry model, and using the new estimates of $A_n$ and $g_s$, it finally computes a new estimate of $C_i$.  The solution is a fixed point of this process, so if the fixed point iteration converges, then this method produces a solution ? 

[...]

Issue #110, opened 2024-06-05
Investigate unrealistic leaf temperatures in enegy balance


justinmcgrath:
Below is a copied comment from a different PR. When solving the energy balance equations, occasionally leaf temperatures are unrealistic. Using a different numerical method (Newton-Raphson instead of fixed point iteration) is a likely first step to address this. It happens very rarely, so it isn't high priority.

> > So I kept a constraint but set it to 30 degrees C.
> 
> A difference of 30 degrees C would be a lot. If it's producing such unrealistic values, then presumably something is wrong with the model or the way it's solved. That seems worth making a note of to revisit later, if there's not a plan to address it here.

There's definitely an issue with the solution method, but there could be something wrong with the model as well. I think switching to a Newton-Raphson solver would fix many of the unreasonable leaf temperature values, but I don't have any intention of doing that as part of this PR.

I took a look at the soybean leaf temperature differences ($\Delta T = T_{leaf} - T_{air}$) on the `main` branch and on this branch. It looks like the hard limit of $\pm 5^{\circ}$ C was being hit pretty frequently on `main`. On this branch, most of the leaf temperatures are within $5^{\circ}$ C of the air temperature, but that isn't being set by an artificial constraint. There is a small fraction of leaves hitting the hard limit of $\pm 30^{\circ}$ C. I could change the limit to 10 since it seems like the real solution is almost always within that range.

![2024-06-04 delta_T](https://github.com/biocro/biocro/assets/48919455/449eaf9f-f238-40f0-ae1a-4fb4a0ad4935)

The code I used is:

```r
library(BioCro)
library(lattice)

res <- with(soybean, {run_biocro(
    initial_values,
    parameters,
    soybean_weather[['2002']],
    direct_modules,
    differential_modules,
    ode_solver
)})

leaf_temperature_names <- grep('leaf_temperature', colnames(res), value = TRUE)

delta_T <- unlist(lapply(leaf_temperature_names, function(lt) {
    res[[lt]] - res$temp
}))

histogram(
    delta_T,
    freq = FALSE,
    breaks = seq(-30, 30, by = 0.5),
    ylim = c(0, 100),
    grid = TRUE
)
```

_Originally posted by @eloch216 in https://github.com/biocro/biocro/issues/109#issuecomment-2148468329_

leavesandlemmas:
[...]

PR #166, opened 2025-03-12
No more fixed point iteration

leavesandlemmas:
This pull request removes the fixed-point iteration used to solve for the balance equation for intercellular the CO2 in `c3photo` and in `c4photo` . This addresses the discussion in  #161  but it doesn't address the leaf temperature as discussed in #110 

 Instead, the intercellular CO2 is computed using the [secant method](https://en.wikipedia.org/wiki/Secant_method).  The rate of convergence of the secant method is almost as fast as Newton's method but it doesn't require an expression for the first derivative, as it effectively approximates the first derivative. It only requires one function evaluation per iteration unlike [Steffensen's method](https://en.wikipedia.org/wiki/Steffensen%27s_method) which is what the [Sun et al. (2012)](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2012JD018059) which does two function evaluations per iteration (Sun et al. don't call their method, instead they say they are using an approximation for the derivative and doing newton's method which is a bit confusing).

The secant method here uses a function `check_assim_rate` that tests whether or not an estimated assimilation rate is correct, rather than `Ci`. It seemed easier to implement this way from the existing code. 

Although we could and should implement multi-dimensional root solvers (see #110), this pull request fixes this particular one and it only took a few days to implement (most of which was actually trying to understand what was going wrong with the ball berry model rather than the secant method itself).

While most of the code for models/modules is unchanged, I did make one small modification to the ball-berry model. If you pass a value of assimilation to the Ball-Berry Model function that causes the leaf surface CO2 `Cs` to be zero or negative, then the model now returns `Inf` because this is the limit of the stomatal conductance as `Cs -> 0`. Infinite conductance is the same as zero resistance; this doesn't cause any issues because the total conductance just equals the the boundary layer conductance (which is never zero nor infinite). 

Minor comments: 
* The variable `Assim_conductance` isn't used anymore so I removed it. 
* I added the variable `Assim_check`; this is the last value of the `check_assim_rate` and it should equal zero (to within tolerances) if a root was correctly identified.  
* I had to update the stored model tests. But the biomass results were unchanged when plotted.
* I also had to update the automatically generated test changes, but these all use very low Ca values where the differences are most likely to appear. 
* I tested the method much more with the C3 modules than the C4 modules, but I did update both. 
* I designed initial guesses which seem to work well, but again these were more thoroughly tested with the C3 modules.
 

### Comparison of the secant method and the fixed point iteration

Here's a plot comparing atmospheric CO2 `Catm` to the computed assimilation rate by both methods:

image

While they agree at high CO2, they differ slightly at low CO2. Zooming on the low CO2 values:

image

We can see that the values differ by a small amount. In fact, the fixed point iteration is reaching its maximum iteration count here. So it hasn't really converged, whereas the secant method has. We can be sure the secant method has correctly converged because we can plot `Assim_check`:

image

Note the fixed point iteration, as originally coded, didn't have a check value. All `Assim_check` values are below the tolerance of `1e-12` which is unfortunately hardcoded. So it cant be easily changed from R but it can be easily changed as needed. Zooming out we see this is true everywhere.

image

We can also compare the number of iterations needed to reach the tolerance. 

image

Note the telltale signs of non-convergence. Zooming in we can see that in general, the secant method requires fewer iterations. 

image

The secant method also has trouble at low `Catm` requiring many more iterations; but it still converges. Although the amount of computation per iteration should be about the same, this isn't a perfect apples-to-apples comparison because the secant method uses a much higher tolerance for convergence than the fixed point iteration (which only requires about 4 digits of accuracy).  

 From some other tests, I believe the increase in iterations needed is mostly due to the nature of the problem becoming somewhat pathological at `Catm = 0`.  The ball berry model is a little weird because if it didn't clamp stomatal conductance to a minimum value, I think it actually introduces two valid roots, though I don't have a graph on hand to show this. The clamping to a minimum value isn't smooth, so it could cause problems where a method approaches the wrong root, then gets throw off course by the clamping.  
 
I should note that even if the Ball Berry model generates two possible roots, since the roots are supposed to be the steady state of a set of differential equations (vector field) see #161 then the stable root would be the correct one under this theory.

Presumably the pathologies could be fixed by using a different stomatal conductance model. But the low CO2 condition rarely happens during a normal BioCro simulation. The light curve doesn't suffer the same pathologies, whatever they may be.

image

Plotting `Assim_check`

image

and the number of iterations:

image

### Is it faster ?

While the secant method is faster than fixed point iteration, it doesn't seem to make  a big difference in simulation time. This is probably because the secant method reduces the number of iterations only by 50% or so under typical conditions.


leavesandlemmas:
Oh, and something that turned out to be super useful for testing the c++ code was to use `Rcpp` to make quick and dirty bindings of the c++ that can then be interacted with in R. The following source file can be compiled and loaded into R using the `Rcpp::sourceCpp('test_ci_balance.cpp')`. It only works with source files, so you have to include any headers you want.

The wrappers are marked with comment `// [[Rcpp::export]]`;  `Rcpp` has some pre processing that replaces those comments with the actual C-to-R binding. `Rcpp` also has a number of classes which mirror the data types in R so it automatically does the conversion: a list is given [here](https://teuder.github.io/rcpp4everyone_en/070_data_types.html).  While Rcpp has got a lot of documentation, it doesn't have much in the way of tutorials beyond simple one-and-done bindings. In principle, it's supposed to facilitate R packages which bind c++ libraries.

The advantage is I was able to compile the actual c++ code that BioCro uses to compute `check_assim_rate` in order to plot `check_assim_rate`  without having to add a module for it, or add it to a module to expose it. It could've have been used to test the implementation of the secant method as well, except I did that part first before I realized I could use `Rcpp` this way. The compilation time is much, much faster than recompiling all of biocro. 

I don't know why, and didn't bother to figure out why, but it seems that the `#include` file paths are only parsed correctly if I put this in the same directory as the header and source files. Ideally I would've kept this file out of the package, and used an explicit file path for imports.  But it's easy enough to delete when you're done testing (or move out of the biocro repository).  

```cpp
// test_ci_balance.cpp

#include <cmath>                             // for pow, sqrt, std::abs
#include <algorithm>                         // for std::min
#include <limits>
#include "ball_berry_gs.h"                   // for ball_berry_gs
#include "FvCB_assim.h"                      // for FvCB_assim
#include "secant_method.h"
#include "../framework/constants.h"          // for dr_stomata, dr_boundary
//#include "conductance_limited_assim.h"       // for conductance_limited_assim
using physical_constants::dr_boundary;
using physical_constants::dr_stomata;

#include <Rcpp.h>
using namespace Rcpp;

double check_assim_rate(double assim, double Ca, double gbw, double RH, double J){
  stomata_outputs BB_res = ball_berry_gs(
    assim * 1e-6,
    Ca * 1e-6,
    /* RH */ RH,
    0.008,
    10.6,
    gbw,
    25,
    25);
  
  double Gs = BB_res.gsw;  // mol / m^2 / s
  
  // Using the value of stomatal conductance,
  // Calculate Ci using the total conductance across the boundary layer
  // and stomata
  double Ci = Ca - assim *
    (dr_boundary / gbw + dr_stomata / Gs);  // micromol / mol
  
  
 FvCB_outputs FvCB_res = FvCB_assim(
    Ci,
    /* Gstar */ 38.6,
    /* J */ J,
    /* Kc */ 259,
    /* Ko */ 179,
    /* Oi*/ 200,
    /*Rd*/ 1,
    /*TPU*/ 11 ,
    /*Vcmax*/ 100,
    /* alpha_TPU */ 0,
    /* electrons_per_carboxylation */ 4,
    /* electrons_per_oxygenation  */ 4);
  
 
  return FvCB_res.An - assim;
}


// // [[Rcpp::export]]
NumericVector check_assim(NumericVector x,double Ca, double gbw=0.2, double RH=0.3, double J = 170) {
  NumericVector out(x.size());
  for (int i = 0; i < x.size(); ++i){
    out[i] = check_assim_rate(x[i], Ca, gbw, RH, J);
  }
  return out;
}
 
// [[Rcpp::export]]
double assim_ub(double Ca,double J){
  return FvCB_assim(
    Ca,
    /* Gstar */ 38.6,
    /* J */ J,
    /* Kc */ 259,
    /* Ko */ 179,
    /* Oi*/ 200,
    /*Rd*/ 1,
    /*TPU*/ 11 ,
    /*Vcmax*/ 100,
    /* alpha_TPU */ 0,
    /* electrons_per_carboxylation */ 4,
    /* electrons_per_oxygenation  */ 4).An;

}

// [[Rcpp::export]]
NumericVector check_bb(NumericVector x,double Ca,double RH, double gbw) {
  NumericVector out(x.size());
  stomata_outputs BB_res;
  for (int i = 0; i < x.size(); ++i){
      BB_res = ball_berry_gs(
      x[i] * 1e-6,
      Ca * 1e-6,
      /* RH */ RH ,
      0.008,
      10.6,
      gbw,
      25,
      25);
    out[i] = BB_res.gsw;
  }
  return out;
}
```

justinmcgrath:
Thanks. This looks promising.

> While most of the code for models/modules is unchanged, I did make one small modification to the ball-berry model. If you pass a value of assimilation to the Ball-Berry Model function that causes the leaf surface CO2 Cs to be zero or negative, then the model now returns Inf because this is the limit of the stomatal conductance as Cs -> 0. Infinite conductance is the same as zero resistance; this doesn't cause any issues because the total conductance just equals the the boundary layer conductance (which is never zero nor infinite).

Was this done in the c++ code and is it necessary for the secant method to work?

The low catm portion demonstrates an adea where this performs better than fixed point. It would be nice to have another physiologically relevant situation that would cause that, for example water stress and thus low conductance. I'm trying to think of something where we can point out that we get better results rather than a hypothetical situation.

Do you have quantification of the speed difference? You can run the model in a loop and system.time() in R to get the total time. You need to change some input so that R doesn't do clever caching stuff. Can make a list of slightly different temperatures for each iteration or something like that, and use the same list for both tests.

Rcpp looks handy for running C++ code from R. For use in the c++ code, it is redundant with the R libraries and only adds a dependency. When compliling you can give a flag to indicate where to look for headers. I assume you can give a similar option to Rcpp.

yufenghe22:
> Is it faster ?
> While the secant method is faster than fixed point iteration, it doesn't seem to make a big difference in simulation time. This is probably because the secant method reduces the number of iterations only by 50% or so under typical conditions.

I think with this 50% less iterations, you would see a lot of performance change when using the ePhotosynthesis instead of FvCB. So it is still a big improvement.

justinmcgrath:
> > Is it faster ?
> > While the secant method is faster than fixed point iteration, it doesn't seem to make a big difference in simulation time. This is probably because the secant method reduces the number of iterations only by 50% or so under typical conditions.
> 
> I think with this 50% less iterations, you would see a lot of performance change when using the ePhotosynthesis instead of FvCB. So it is still a big improvement.

Is this used when ePS is coupled to BioCro?

leavesandlemmas:
> Was this done in the c++ code and is it necessary for the secant method to work?

Yes, it is done in the c++ code: https://github.com/biocro/biocro/blob/daa8a28222e66bdd5c4ce1b485896d89077b0a2b/src/module_library/ball_berry_gs.cpp#L118

And yes, it seems to be necessary for the secant method to reliably find the assimilation value.  

Initially the code checked if `Cs==0` and then threw an error. The fixed point iteration guaranteed `Cs > 0` for all iterations, but the secant method can try values of `assim` that give `Cs <= 0`.  Because that can't happen, it means the value for `assim`  is too big. But the original code would throw an error, or worse compute stomatal conductance using the negative value.  The former case prevents the secant method from refining the guess; and the latter leads to undefined behavior, in the sense that the Ball Berry model doesn't make any sense for negative `Cs`. 
You can see an example here, where the stomatal conductance becomes infinite at a finite value of `assim`: 

image

where this plot is computed using
```r
bb <- function(assim, Catm, rh, gbw, b0 = 0.008, b1=10.6){
  A <- assim * 1e-6; #micromol to mol    
  b0 + b1 * rh *A/(Catm*1e-6 - 1.37*A/gbw) * (assim > 0) 
}
```
which is effectively what the `ball_berry_gs` function does.  I got lazy and didn't bother translating the quadratic solution for leaf surface `rh` but it shows the same behavior. 

You can see there's a singularity at `assim = Catm * gbw/1.37` because for this value of `assim` , we have `Cs = 0`. Now, the issue is worse as the boundary layer conductance goes to zero `gbw` because it shifts the singularity to the left. And it seems that when `gbw > 0.2`, the singularity is far enough to right to not be noticeable. And I think under normal conditions, `gbw` is not usually that low, but I didn't really look into the modules that compute the boundary layer conductance. So now the `ball_berry_gs` will return the [positive infinity value](https://en.wikipedia.org/wiki/IEEE_754#Infinities).

Alternatively stated, at `assim = Catm * gbw/1.37` ,  the ball-berry model predicts zero stomatal *resistance* (reciprocal of conductance). And for larger values of `assim`, the ball-berry model predicts negative resistances. 

image

But what should happen is that stomatal resistance goes to zero, then stays there for all higher values of `assim`. So in terms of stomatal conductance, it has to go to infinity and then stay there. Quite similar to the TPU limitation in the FvCB model. 

In theory, the secant method could check to ensure it doesn't try values of `assim` that are too large. But that means the secant method has to be aware of this values. And it seemed easier to just modify the ball berry model. 

-----

It might be possible to modify the starting guesses to get slightly better performance.  Currently it uses the lower bound and a upper bound. The lower bound is `assim = Rd` or the CO2 assimilation rate is equal to the leaf respiration rate. That is usually the case whenever light is low (like at night). So the method will terminate within 0-2 iterations.  However, normally this will be way too low. 

The upper bound is basically the rate of assimilation if `Ca== Ci` which will almost always be too high.  Since the stomatal conductance becomes infinite at `assim = Catm * gbw / 1.37`  and that will also almost always be too high, it should not try anything larger than that, so it takes whichever it smaller (but not less than zero).  

```cpp
    // compute upper bound as guess
    double assim_ub = FvCB_assim(
                          Ca, Gstar, J, Kc, Ko, Oi, Rd, TPU, Vcmax, alpha_TPU,
                          electrons_per_carboxylation,
                          electrons_per_oxygenation)
                          .An;
    if (assim_ub < 0) {
        assim_ub = 0;
    }
    assim_ub = std::min(Ca * gbw / dr_boundary, assim_ub);
```

However, it might be better to use something like `Ci = 0.8 * Ca` as a starting guess, and maybe only use `Rd` at night. The equations are hairy enough; that I haven't been able to come up with a more sophisticated approximation.

eloch216:
This sounds pretty cool, but I probably won't be able to look at it in depth until late March. Unless it's urgent, I would appreciate if this PR was not merged until I'm able to look at it closely.

For now I have one short comment about something that caught my eye: the purpose of using `co2_assimilation_rate = std::min(FvCB_res.An, an_conductance);` was to prevent `Ci` from becoming negative, which should also prevent `Cs` from becoming negative. This is why the check for `Cs < 0` in the Ball-Berry code was never triggered before. See the documentation for the `conductance_limited_assim()` function for more info.

From prior experience with various parts of BioCro, things like `if (Cs < 0) {Cs = 0}` tend to cause more problems than they solve. It's better to understand why the weird condition has occurred, rather than just overriding it.

Why was `an_conductance` removed; was it somehow causing a problem for the secant method?

src/module_library/secant_method.h
```diff
@@ -0,0 +1,180 @@
...
+ * from two points.
+ *
+ *  \f[ f(x) \approx f(x_1) + \left(\frac{f(x_1) - f(x_0)}{x_1 - x_0}\right) \left(x -x_1\right) \f]
+ *
+ * The root of this approximation is computed to produce to refine a guess for
+ * the root. Let \f$y_n = f(x_n)\f$; the secant method's update rule is:
+ *
+ * \f[ x_{n+1} = x_n - y_n \left(\frac{x_n - x_{n-1}}{y_n - y_{n-1}}\right)\f]
+ *
+ * This formula can be rearranged into
```
justinmcgrath, line 73: Something is missing here.

src/module_library/secant_method.h
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * @return Approximate root. If the method fails, it will return it's last (and
+ *      presumably best) guess for the root. If `check` member field in the
+ *      secant_parameters object is not zero, then the method has failed.
+ *      This routine will terminate if NaN or non-finite guesses are encountered.
+ */
+template <typename F>
+double find_root_secant_method(
+    F fun, double x0, double x1, secant_parameters& par)
+{
```
justinmcgrath, line 103: I have found that using structs like this introduces more inconveniences than they remove. If you have a good reason for it, then it could be OK, but here I don't see how this is helpful. Structs are used in C in a different style of object-orient programming, and in C++ classes are used to the exclusion of that style. Struct generally are not used simply as a way to group variables.

justinmcgrath: It looks like the struct is also being used as a way to have default values, and that can be done in the function signature.

src/module_library/secant_method.h
```diff
@@ -0,0 +1,180 @@
...
+    // check inputs for trivial solutions
```
justinmcgrath, line 104: I don't think we have guidelines about this, but I feel that comments should start with a capital letter.

src/module_library/secant_method.h
```diff
@@ -0,0 +1,180 @@
...
+    if (is_zero(y1, par.atol)) {
+        par.counter = 0;
+        par.check = y1;
+        return x1;
+    }
+
+    for (size_t n = 1; n <= par.max_iter; ++n) {
+        double secant_slope = (y1 - y0) / (x1 - x0);
+
+        // SECANT UPDATE FORMULA;
```
justinmcgrath, line 129: I also think that all caps is not appropriate, except for the usual cases in English. All caps names for constants are OK too.

src/module_library/secant_method.h
justinmcgrath: I see that the struct is also used to pass information out, and there's not a good replacement for that. I would pass it as a pointer though. It's not a convention everywhere, but it's common to pass by reference when the object isn't intended to be modified and pass as a pointer when it is intended to be modified. Although the absence of "const" also indicates that, that's easy to take as someone forgetting to add it. Making it a pointer gives a hint that modification is intended.

src/module_library/c4photo.cpp
```diff
@@ -81,23 +71,24 @@ photosynthesis_outputs c4photoC(
...
+    // during the secant method's iterations.
+    // Here we make an initial guess that Ci = 0.4 * Ca.
+    stomata_outputs BB_res;
+    double InterCellularCO2{0.4 * Ca_pa};  // Pa
+    double Assim{};                        // micromol / m^2 / s
+    double Gs{};                           // mol / m^2 / s
 
+    // This lambda function equals zero
+    // only if assim satisfies both collatz assim and Ball Berry model
+    auto check_assim_rate = [&](double assim) {
```
justinmcgrath, line 87: I don't see what a lambda adds here. It's called directly, so it's creation wasn't needed to satisfy requirements of some other function. It's only used once, so it could have been written in place.

src/module_library/secant_method.h
leavesandlemmas: Oops yeah, I removed that. And forgot to remove this comment.

src/module_library/c4photo.cpp
leavesandlemmas: Written in place as the first argument of `find_root_secant_method` ? Wouldn't that be a little hard to read? Or the whole method should be embedded here?

justinmcgrath: I misread the later code. You can ignore my comment about the lambda.

leavesandlemmas: For what it's worth, I originally tried defining a normal function. But you can't declare / define functions inside other functions for some reason. So it has to be a lambda or function object.  Also as is, the `find_root_secant_method` won't accept a function with any signature other than `double(double)`. But I wanted to keep the secant method code separate so it can be reused. I think the design of `c3photo` could be improved, but I wanted to limit the number of changes

src/module_library/secant_method.h
leavesandlemmas: Originally, this function's declaration was
```cpp
template<typename T>
double find_root_secant_method(F fun, double x0, double x1, size_t max_iter, double atol, double rtol) 
```
and I added default values for the last three arguments.  But since `c3photo` already returned the number of iterations, and that was useful as a diagnostic, I then passed  `iterCounter` to the function by reference, so it could set the number of iterations.  And the check was computed outside the function, by evaluating the returned solution. 

But I thought the interface was too messy, and computing the check outside is a bit inefficient since `find_root_secant_method` would have already evaluated the function at the solution when it decided to terminate. But to return two floats would require a struct as the return object.  So the struct `secant_parameters` mostly exists fix both problems.  

> It does seem confusing to me to group together things that are passed in and passed out into a single object.

I agree that it's not ideal to have outputs and inputs contained in the same object. Maybe a better design is something like

```cpp

class secant_method {
  double atol = 1e-12;
  double rtol = 1e-12;
  double max_iter;
  
  public:
    
    // expose 
    int counter;
    double check; 
    
    // some constructor   
    secant_method(double maxiter, double atol, double rtol) {}

    template<typename F>
    double find_root(F fun, double x0, double x1); 
};
```

The function `fun` could also be saved as a member field. I didn't want to use a class template though because in C++11, the template instantiations of class templates have to be explicit i.e. something like `secant_method<double(double)> solver;` which is a bit verbose. `secant_parameters` could be split into two structs, one for outputs and one for inputs.

> It's not a convention everywhere, but it's common to pass by reference when the object isn't intended to be modified and pass as a pointer when it is intended to be modified.

What I learned from c++ books was to pass by non-constant reference if modification is intended and by constant reference if no modification is intended, and basically never pass by raw pointer (unless one has to).  

> Structs are used in C in a different style of object-orient programming, and in C++ classes are used to the exclusion of that style. Struct generally are not used simply as a way to group variables.

What is the purpose of a struct in C then?  I haven't done much C programming, but it seems like `struct` is intended to be a product data type, the cartesian product of other data types. Especially since there's `enum` (data type of finite sets) and `union` (union/coproduct of data types).

leavesandlemmas:
> Do you have quantification of the speed difference? You can run the model in a loop and system.time() in R to get the total time. You need to change some input so that R doesn't do clever caching stuff. Can make a list of slightly different temperatures for each iteration or something like that, and use the same list for both tests

Yeah I used that, although not with anything sophisticated in terms of benchmarking. Using this code to time a single simulation (and then repeating that a few times to measure the variance)
```r
system.time(
  run_biocro(
    soybean$initial_values,
    soybean$parameters,
    soybean_weather$`2006`,
    soybean$direct_modules,
    soybean$differential_modules,
    soybean$ode_solver
  )
)
```
I get 1.65 --1.66 secs on my desktop for this pull request. And like 1.58-1.60 on the old one. So if anything, it's a bit slower for some reason. Though maybe that's caching and a better benchmark would show similar behavior. I don't know why . Hard to say without profiling the c++ code.

yufenghe22:
> I get 1.65 --1.66 secs on my desktop for this pull request. And like 1.58-1.60 on the old one. So if anything, it's a bit slower for some reason. Though maybe that's caching and a better benchmark would show similar behavior. I don't know why . Hard to say without profiling the c++ code.

This might suggest the time cost of the Secant method itself outweighs the savings on the number of iterations. The time cost of FvCB is pretty small, and reducing iterations from 8 to 5 times may not be very significant. The stability of the Secant is the major benefit.

justinmcgrath:
> > Do you have quantification of the speed difference? You can run the model in a loop and system.time() in R to get the total time. You need to change some input so that R doesn't do clever caching stuff. Can make a list of slightly different temperatures for each iteration or something like that, and use the same list for both tests
> 
> Yeah I used that, although not with anything sophisticated in terms of benchmarking. Using this code to time a single simulation (and then repeating that a few times to measure the variance)
> 
> ```r
> system.time(
>   run_biocro(
>     soybean$initial_values,
>     soybean$parameters,
>     soybean_weather$`2006`,
>     soybean$direct_modules,
>     soybean$differential_modules,
>     soybean$ode_solver
>   )
> )
> ```
> 
> I get 1.65 --1.66 secs on my desktop for this pull request. And like 1.58-1.60 on the old one. So if anything, it's a bit slower for some reason. Though maybe that's caching and a better benchmark would show similar behavior. I don't know why . Hard to say without profiling the c++ code.

I'd run several times, 10 to 100, in a loop. Check what people say about the details of that to avoid R noticing that that's pointless and speeding it up for you.

src/module_library/secant_method.h
justinmcgrath: At some point it would be noce to generalize this, and then a class design would be helpful, but I'd start with the abstract class that would cover what we're thinking about in general. Offhand, this doesn't look like what I had vaguely imagined and I would do something simpler - ungrouped arguments for all inputs and a struct of outputs, but you could probably pass all individually, and use pointers for the output.

Opinions vary on style for outputs in the argument. Some people really dislike pointers and don't ever want to see them regardless of the context. I don't have that opinion, and I think they convey information about "ownership", the area of the program responsible for the memory of the variable. If you're passing pointers, the receiving function doesn't own them. That's not exactly what's going on here, but it's a bit of a hint that it can be modified.

structs are groups of types, but it's not used simply to group things together, as in to make a function of 5 arguments into a function of 1 argument. They're used in a manner similar to classes, but in C++ the language does a lot of the writing that you would have had to do in C. I never learned the details of this, and only know the general aspects. Basically, if I see it in C code, I understand what they're doing, but I couldn't do it myself, and thus I also can't really explain it.

leavesandlemmas:
> This sounds pretty cool, but I probably won't be able to look at it in depth until late March. Unless it's urgent, I would appreciate if this PR was not merged until I'm able to look at it closely.

Sure thing. I'll put my answer to your question below for whenever you get a sec.

> For now I have one short comment about something that caught my eye: the purpose of using `co2_assimilation_rate = std::min(FvCB_res.An, an_conductance);` was to prevent `Ci` from becoming negative, which should also prevent `Cs` from becoming negative. This is why the check for `Cs < 0` in the Ball-Berry code was never triggered before. See the documentation for the `conductance_limited_assim()` function for more info.
> 
> From prior experience with various parts of BioCro, things like `if (Cs < 0) {Cs = 0}` tend to cause more problems than they solve. It's better to understand why the weird condition has occurred, rather than just overriding it.
> 
> Why was `an_conductance` removed; was it somehow causing a problem for the secant method?

So the trickiest part was trying to recast the existing into a function whose root would be the same as the fixed point.  In the issue #161 , I was thinking out loud / asking questions about what the exact equation is, so let me know if I misunderstood something. 

Mathematically, we have the FvCB model as `FvCB(Ci) = assim` (suppressing other arguments) and we have `ball_berry_gs(assim, Ca) = gsw` which can be used to compute `Ci = get_ci(assim, gsw,  Ca) = Ca -  assim/gtotal`  (this function isn't named) where so using `assim` one can compute `gsw`  then compute `Ci` and using `Ci` one can compute `assim`.  The loop is what the fixed point iteration does.

So starting with `assim`, we can compute `Ci` using `ball_berry_bs` and `get_ci`, then use `FvCB` to compute `assim` but that should match the value of `assim` we guessed. So using pseudo code as math notation:

```
check_assim_rate(assim, Ca)  = FvCB(Ci) - assim 
  where
    Ci = get_ci(assim, Ca, gsw) 
    gsw = ball_berry_gs(assim, Ca)
```

According to the documentation, the conductance limited assimilation rate doesn't limit the assimilation rate so it shouldn't appear here I think. And that's why I ditched it.  It doesn't seem to be necessary.  Presumably, it could be used to ensure the starting guesses aren't too high. Also from looking at the documentation of `an_conductance_limited`, it seems like we could just clamp `Ci` to zero in case `Ca - assim/gtotal` happens to be negative. I might try that if I have time.

The secant method doesn't have any in-built bounds checking because that requires careful implementation to avoid breaking stuff.  But I could implement some sort of hybrid between the secant method and the [false position method](https://en.wikipedia.org/wiki/Regula_falsi). The latter uses bracketing to ensure convergence to a root, at least if the initial bracket spans one simple root. The false position method isn't usually as fast to converge. 

If we were to use bracketing, then having bounds on `assim` for any given situation is very useful because the bounds should span the solution.

leavesandlemmas:
> As you point out, the same end result can be achieved by clamping `Cs` within the Ball-Berry code, and clamping `Ci` in `c3photoC`. BTW, I did not know there was a name for this before. But now I see it has a wiki page! https://en.wikipedia.org/wiki/Clamp_(function)

Also sometimes called the clip function.  

> As I mentioned above, using conductance_limited_assimilation prevents negative Cs and Ci from occurring in the Ball-Berry calculations.

But doesn't this just clamp the value of assim if the conductance limited assimilation is never actually limiting ?  It doesn't seem fundamentally different.  

> But, I have found that solving numeric issues by clamping often causes other problems to pop up downstream that can be very difficult to diagnose and fix. I like to see this kind of clamping code removed from BioCro, not added to it.

I agree that it's not ideal.  But throwing errors might also prevent the secant method code from successfully finding the correct `assim` value. The secant method is going to try assim values that arent consistent with the 1D gas flow steady state to check how "wrong" the value is. But if `Catm` is low, the correct value of `Ci`  or `Cs` will be near physically impossible values. I suspect the conductance limited assimilation helps by clamping `assim` to valid values. Maybe the secant method could be specialized here to use this information though, so it won't try an invalid value but can recover and continue.

eloch216:
> Sorry about the failed tests... I didn't think the code would fail the tests. But I made a small change (fixing an error) to the initial guess which caused the module test case to take one fewer iterations, which then raises an error.
> 
> Is there a way to have the tests ignore the number of iterations ?

At the moment, there is not a way to have the module tests ignore any output quantities. It may be annoying at times, but I do think it is for the best -- I would want to know if the loop suddenly needed more iterations to converge, for example.

eloch216:
> > As I mentioned above, using conductance_limited_assimilation prevents negative Cs and Ci from occurring in the Ball-Berry calculations.
> 
> But doesn't this just clamp the value of assim if the conductance limited assimilation is never actually limiting ? It doesn't seem fundamentally different.

A fundamental difference is that on this branch, the Ball-Berry module can sometimes return results that are not self-consistent. Here's what I mean:

There are two "consistency checks" we can make for the outputs from the Ball-Berry module. One is to calculate the CO2 flux across the boundary layer, which is given by `gbc * (Ca - Cs) = (gbw / 1.37) * (Ca - Cs)`. This flux should always be equal to the net CO2 assimilation rate, since we assume steady-state conditions. The other is to recalculate `Cs` using `Ca - An / gbc = Ca - An / (gbw / 1.37)`. This should always be equal to the value of `Cs` returned by the module.

For very large values of the net CO2 assimilation rate, we can see that the fluxes and CO2 concentrations diverge; in other words, that the consistency checks fail (see figure below). This is likely to cause problems.

On the other branch, these inconsistent results are never returned by the Ball-Berry module because an error would be thrown for large values of `An` where `Cs` becomes negative.

I guess it's probably wrong to say that clamping causes the issue; it's that allowing negative values of `Cs` and replacing them with 0 doesn't have a physical basis. The conductance-limited assimilation approach does have a physical basis. It's important to remember that our Ball-Berry code should always produce reasonable results; it has an existence outside of the convergence loop in `c3photoC`.

Picture1

Here's the code I used to make these figures:

```r
# Load packages
library(BioCro)
library(PhotoGEA)
library(lattice)

# Run Ball-Berry model for different values of A
res <- module_response_curve(
    'BioCro:ball_berry',
    within(soybean$parameters, {
        gbw = 1.2
        leaf_temperature = 30
        rh = 0.7
        temp = 27
    }),
    data.frame(net_assimilation_rate = seq(-20, 400))
)

# Calculate the CO2 flux across the boundary layer and recalculate the CO2
# concentration at the leaf surface
res <- within(res, {
    co2_flux_bdry = (gbw / 1.37) * (Catm - cs)
    cs_calculated = Catm - net_assimilation_rate / (gbw / 1.37)
})

# Make plots
pdf_print(
    xyplot(
        leaf_stomatal_conductance ~ net_assimilation_rate,
        type = 'l',
        data = res
    )
)

pdf_print(
    xyplot(
        hs ~ net_assimilation_rate,
        type = 'l',
        data = res
    )
)

pdf_print(
    xyplot(
        net_assimilation_rate + co2_flux_bdry ~ net_assimilation_rate,
        type = 'l',
        data = res,
        xlab = 'Net CO2 assimilation rate (micromol / m^2 / s)',
        ylab = 'CO2 flux (micromol / m^2 / s)',
        par.settings = list(
            superpose.line = list(lwd = 2, lty = c(1, 2, 3))
        ),
        auto.key = list(
            space = 'top',
            text = c(
                'Net CO2 assimilation rate',
                'CO2 flux across the boundary layer'
            )
        )
    )
)

pdf_print(
    xyplot(
        cs + cs_calculated ~ net_assimilation_rate,
        type = 'l',
        data = res,
        xlab = 'Net CO2 assimilation rate (micromol / m^2 / s)',
        ylab = 'CO2 concentration at leaf surface (micromol / m^2 / s)',
        par.settings = list(
            superpose.line = list(lwd = 2, lty = c(1, 2, 3))
        ),
        auto.key = list(
            space = 'top',
            text = c(
                'Cs from Ball-Berry module',
                'Ca - An / gbc'
            )
        )
    )
)
```

eloch216:
I tried using some different starting guesses in `c3photoC`, which are based on the FvCB model predictions at `Ci = 0` and `Ci = inifity`:

```c++
    // Find upper and lower bounds on the net assimilation rate, to be used as
    // initial guesses for the secant method
    double const assim_ub = std::min(
        std::min(
            Vcmax - Rd,         // The maximum value of Ac, which occurs at Ci = infinity
            J / 4 - Rd),        // The maximum value of Aj, which occurs at Ci = infinity
        Ca * gbw / dr_boundary  // The maximum conductance-limited An, which occurs for gsw = infinity
    );                          // micromol / m^2 / s

    double const assim_lb = std::max(
        -Gstar * Vcmax / (Kc * (1 + Oi / Ko)) - Rd,  // The value of Ac when Ci = 0
        -J / 8 - Rd                                  // The value of Aj when Ci = 0
    );                                               // micromol / m^2 / s
```

I tested the results using the code in this comment: https://github.com/biocro/biocro/pull/169#issuecomment-2749651767

With these guesses, there is a small improvement in the number of iterations. But more important, the secant method converges for all the `Catm` values tested.

@leavesandlemmas , if you like this change, please add it in a new commit. I won't be making any changes here since this is your PR.

justinmcgrath:
Since this is for C3 specifically, I think the upper bound could be at Ci = Ca. Maybe that doesn't work in the dark? Scrap this idea if you think it sacrifices robustness.

eloch216:
> Since this is for C3 specifically, I think the upper bound could be at Ci = Ca. Maybe that doesn't work in the dark? Scrap this idea if you think it sacrifices robustness.

I tried it out and it seems to work well. Using the test example, the loop converges for all Ca, with fewer iterations at low Ca. Here is the code that I used:

```c++
    // Find upper and lower bounds on the net assimilation rate, to be used as
    // initial guesses for the secant method. We add a small offset (1e-3) to
    // ensure the guesses are distinct, since they would otherwise be identical
    // when Ca = 0
    double const assim_ub =
        1e-3 + std::min(
                   FvCB_assim(
                       Ca, Gstar, J, Kc, Ko, Oi, Rd, TPU, Vcmax, alpha_TPU,
                       electrons_per_carboxylation,
                       electrons_per_oxygenation)
                       .An,                // The predicted value of An when Ci = Ca
                   Ca * gbw / dr_boundary  // The maximum conductance-limited An, which occurs for gsw = infinity
               );                          // micromol / m^2 / s

    double const assim_lb =
        -1e-3 + std::max(
                    -Gstar * Vcmax / (Kc * (1 + Oi / Ko)) - Rd,  // The value of Ac when Ci = 0
                    -J / (2.0 * electrons_per_oxygenation) - Rd  // The value of Aj when Ci = 0
                );                                               // micromol / m^2 / s
```

leavesandlemmas:
I also increased the maximum number of iterations to 1000. It is unlikely to ever need that many.   I don't have any issues with the guesses. 

 I also tried designing some better ones than the original ones I had.  These don't have to be upper and lower bounds, but the `check_assim` function is pretty linear when Ca > 100, gbw > 0.2 , and  RH > 0.2 . So basically it doesn't matter very much under the ideal conditions.  

Low Ca, gbw, or RH each cause the `check_assim` function's first derivative to vary around `assim = 0`.  Low gbw and low RH appear to be more pathological than low Ca.  I also am fairly certain a bifurcation happens, and several solutions become possible, although it's possible Ed's changes prevent that from happening. 

A lower and upper bound should give a secant that's roughly the average of first derivative of the `check_assim` so it should be fairly robust in terms of reliably converging. Explicit bracketing would make the stability / robustness even better; and we would want good bounds, since bracketing refines the bounds.  

However,  guesses which start closer to the correct solution could converge faster than the lower and upper bound.

> Since this is for C3 specifically, I think the upper bound could be at Ci = Ca. Maybe that doesn't work in the dark? Scrap this idea if you think it sacrifices robustness.

It is tricky to design guesses which work for all edge cases. If Ca is sufficiently low ,or at night, then Ci > Ca.

justinmcgrath:
> > Since this is for C3 specifically, I think the upper bound could be at Ci = Ca. Maybe that doesn't work in the dark? Scrap this idea if you think it sacrifices robustness.
> 
> I tried it out and it seems to work well. Using the test example, the loop converges for all Ca, with fewer iterations at low Ca. Here is the code that I used:
> 
> ```c++
>     // Find upper and lower bounds on the net assimilation rate, to be used as
>     // initial guesses for the secant method. We add a small offset (1e-3) to
>     // ensure the guesses are distinct, since they would otherwise be identical
>     // when Ca = 0
>     double const assim_ub =
>         1e-3 + std::min(
>                    FvCB_assim(
>                        Ca, Gstar, J, Kc, Ko, Oi, Rd, TPU, Vcmax, alpha_TPU,
>                        electrons_per_carboxylation,
>                        electrons_per_oxygenation)
>                        .An,                // The predicted value of An when Ci = Ca
>                    Ca * gbw / dr_boundary  // The maximum conductance-limited An, which occurs for gsw = infinity
>                );                          // micromol / m^2 / s
> 
>     double const assim_lb =
>         -1e-3 + std::max(
>                     -Gstar * Vcmax / (Kc * (1 + Oi / Ko)) - Rd,  // The value of Ac when Ci = 0
>                     -J / (2.0 * electrons_per_oxygenation) - Rd  // The value of Aj when Ci = 0
>                 );                                               // micromol / m^2 / s
> ```

Can you force a situation where Ci  might be larger than Ca? Does this bound break things in the dark?

leavesandlemmas:
> Can you force a situation where Ci might be larger than Ca?  Does this bound break things in the dark?

Yes, it's easily forced by setting `Ca = 0` or `Qabs = J = 0` , this causes net efflux of CO2 out of the leaf and therefore `Ci > Cs > Ca` if the defining equations are satisfied.  Since they are diffusion equations, flux must go against the gradient of CO2 concentration. 

But it's not really a bound, and the secant method can easily go out of bounds if the secant's zero lies outside the interval spanned by the initial guesses. The method could be modified to require bounds and use them explicitly. The method of false position or *regula falsi* does this.  Bracketing provides a mathematical guarantee of convergence, but it can also slow down convergence.  So I was sort of envisioning a hybrid method where the algorithm would try the secant method on initial guesses, but also check each iteration against bounds. If a secant's zero lies outside of the bounds, then the algorithm can fall back to regula falsi.

justinmcgrath:
Using bounds can lead to other issues, and I tend to use them conservatively. My concern was that for situations where Ci > Ca, it wouldn't work if if these were bounds. I didn't know they were just guesses. If it's an extreme guess, and not a bound, then Assim at Ci = Ca is a good extreme guess, even if it's not quite as extreme as it could be.

leavesandlemmas:
Here's an excellent GIF that illustrates how the secant method works: 


Although in this example, the initial guesses `a, b` do also form a bracket (an interval that spans the root) and the secant method never generates a secant whose zero is outside this bracket.  But the secant method still works if `b` were to the left of the root. 

Bracketing methods use the fact that if signs of `f(a)` and `f(b)`  are different, then at least one root must lie inside `[a,b]`. They generally preserve this property at every iteration so if you can guess a starting bracket, then convergence to one of the roots inside this starting bracket is guaranteed.  The basic secant method doesn't use this information. For problems that aren't too pathological it's not an issue. Issues occur if the function has points where the derivative is zero (stationary points), since the secants can then have zeros that lie far away from the current points.

leavesandlemmas:
I have updated the NEWS and added a comment about Sun et al. 

@eloch216  or   @justinmcgrath  do you have any other changes to this pull request ?  If not, I will merge it into develop.

eloch216:
> I have updated the NEWS and added a comment about Sun et al.
> 
> @eloch216 or @justinmcgrath do you have any other changes to this pull request ? If not, I will merge it into develop.

I am happy with the main content of this PR. There may still be a few style changes.

1. I'm not sure how I feel about the new "changelog" comments inside `c3photo.cpp` and `c4photo.cpp`. Having a history in multiple places (`NEWS.md` and these files) might make it more difficult to maintain. We haven't done this for any other files -- the documentation within each file generally just describes how it currently works.

2. Earlier, Justin said "I don't think we have guidelines about this, but I feel that comments should start with a capital letter." I fixed some of these in my PRs, but I think there are still some new comments that don't follow this rule.

leavesandlemmas:
> I am happy with the main content of this PR. There may still be a few style changes.
> 
>     1. I'm not sure how I feel about the new "changelog" comments inside `c3photo.cpp` and `c4photo.cpp`. Having a history in multiple places (`NEWS.md` and these files) might make it more difficult to maintain. We haven't done this for any other files -- the documentation within each file generally just describes how it currently works.

Just trying to incorporate a comment about Sun et al somewhere. It could get easily buried in the NEWS, so I thought it should be added to the code where it's used. But there's a possibility that this becomes out of dated in the future, hence I put a date on it. Easy to remove though.

justinmcgrath:
In the PS function it could say something like "We use the secant method because of known convergence issues when using fixed-point iteration with this set of equations (Sun et al. ...)."

src/module_library/c3photo.cpp
```diff
@@ -1,15 +1,25 @@
...
+#include "FvCB_assim.h"                 // for FvCB_assim
+#include "secant_method.h"              // for find_root_secant_method
 #include "c3photo.h"
 
 using physical_constants::dr_boundary;
 using physical_constants::dr_stomata;
 
+/*
+
+  Secant method used instead of based on the problem identified by Sun et al.
```
eloch216, line 16: > Secant method used instead of based on the problem identified by Sun et al.

This doesn't seem quite right. Ditto for the comment in `c4photo.cpp`.

leavesandlemmas: Oops didn't read that closely.

What about this?
```
  The secant method is used to solve for assimilation, Ci, and stomatal conductance, 
  because of known convergence issues when using fixed-point iteration, based on
  Sun et al. (2012) "A numerical issue in calculating the coupled carbon and 
  water fluxes in a climate model." *Journal of Geophysical Research*
  https://dx.doi.org/10.1029/2012JD018059
```

eloch216 (approved):

PR #169, opened 2025-03-24
No more fixed point iteration ebl

eloch216:
This PR makes a few modifications to the new secant method code.

----

In the first commit, I made the following changes:
- In `ball_berry_gs`, an error is thrown when `Cs` is not positive instead of clamping `Cs` to 0.
- `c3photoC` has gone back to using `conductance_limited_assim` to make sure `Cs` is positive.

After these changes, the outputs from `c3photoC` had barely changed -- just one test needed to be updated. For a few low values of `Catm`, the calculated assimilation rates are slightly different than on the `no-more-fixed-point-iteration` branch, but this doesn't seem like a big deal since both versions sometimes fail to converge at low `Catm`. The code on this branch also needs one more iteration to run in most cases.

Picture1

----

In the second commit, I made a few other changes.

First, I noticed that there were two variables in `c4photoC` with similar names:
- `assim` is an input argument to the `check_assim_rate` lambda.
- `Assim` is used to store the final value of `assim` found by the secant method, but is defined before the lambda.

I found one spot where `Assim` was used instead of `assim` within the lambda, and fixed this. This caused some changes in the C4 assimilation calculations and I needed to update a few stored test results.

Similarly, I noticed that there were two variables named `InterCellularCO2`; one is an input argument to the `collatz_assim` lambda, while the other is updated during the convergence loop. I found this to be confusing, so I renamed the variable used in the loop to `Ci_pa`, using a similar convention to `Ca_pa`. I don't think this has any impact on the calculations but it makes the code easier to follow.

----

In the third commit, I changed `c4photoC` to go back to using `conductance_limited_assim`, as I did for `c3photoC`.

----

In the fourth commit, I added `const` specifiers to `c3photoC` and `c4photoC` when appropriate, and changed the lambda captures to explicitly show which variables must be captured by reference. I think these changes make it more clear which variables are being altered by the lambdas.


eloch216:
Here's the code I used to generate the figure above:

```r
library(BioCro)
library(PhotoGEA)

aci_curves <- module_response_curve(
    'BioCro:c3_assimilation',
    within(soybean$parameters, {
        Catm = NULL
        gbw = 1.2
        Qabs = 1500
        rh = 0.7
        StomataWS = 1
        temp = 30
        Tleaf = 30
    }),
    data.frame(
        Catm = seq(0, 2000)
    )
)

pdf_print(
  lattice::xyplot(
    Assim ~ Catm,
    data = aci_curves,
    auto = TRUE,
    type = 'b'
  ),
  width = 12,
  height = 4
)

pdf_print(
  lattice::xyplot(
    Assim ~ Catm,
    data = aci_curves,
    auto = TRUE,
    type = 'b',
    xlim = c(0, 150),
    ylim = c(-1, 5)
  ),
  width = 12,
  height = 4
)

pdf_print(
  lattice::xyplot(
    iterations ~ Catm,
    data = aci_curves,
    auto = TRUE,
    type = 'b'
  ),
  width = 12,
  height = 4
)

pdf_print(
  lattice::xyplot(
    iterations ~ Catm,
    data = aci_curves,
    auto = TRUE,
    type = 'b',
    ylim = c(0, 6)
  ),
  width = 12,
  height = 4
)

pdf_print(
  lattice::xyplot(
    Assim_check ~ Catm,
    data = aci_curves,
    auto = TRUE,
    type = 'b',
    ylim = c(-1e-12, 1e-12)
  ),
  width = 12,
  height = 4
)

pdf_print(
  lattice::xyplot(
    Gs ~ Catm,
    data = aci_curves,
    auto = TRUE,
    type = 'b'
  ),
  width = 12,
  height = 4
)

pdf_print(
  lattice::xyplot(
    Ci ~ Catm,
    data = aci_curves,
    auto = TRUE,
    type = 'b'
  ),
  width = 12,
  height = 4
)
```

src/module_library/ball_berry_gs.cpp
```diff
@@ -116,13 +116,7 @@ stomata_outputs ball_berry_gs(
...
 
     if (Cs <= 0.0) {
-        // Stomatal conductance becomes infinite as Cs -> 0.0
-        double inf = std::numeric_limits<double>::infinity();
-        return stomata_outputs{
-            /* .cs = */ 0,    // micromol / mol
-            /* .hs = */ 0,    // dimensionless
-            /* .gsw = */ inf  // mol / m^2 / s
-        };
+        throw std::range_error("Thrown in ball_berry_gs: Cs is not positive.");
```
leavesandlemmas, line 119: I think I was running to problems here when plotting the lambda function. Because sufficiently high enough values of `assim` cause this code to throw an error if `Cs<0` is encountered at any point during the secant method's iteration, even if it would still converge afterward.

But I am glad it doesn't  seem to be necessary.

src/module_library/c3photo.cpp
```diff
@@ -78,16 +78,27 @@ photosynthesis_outputs c3photoC(
     // these are updated as a side effect in the secant method iterations
     FvCB_outputs FvCB_res;
     stomata_outputs BB_res;
-    double Gs{1e3};  // mol / m^2 / s      (initial guess)
-    double Ci{0.0};  // micromol / mol     (initial guess)
+    double an_conductance{};  // mol / m^2 / s
+    double Gs{1e3};           // mol / m^2 / s  (initial guess)
```
leavesandlemmas, line 82: Should we remove the comments here ?  Technically this is no longer an initial guess, just variables to store the output.

src/module_library/c3photo.cpp
```diff
@@ -78,16 +78,27 @@ photosynthesis_outputs c3photoC(
...
+    double Ci{0.0};           // micromol / mol (initial guess)
 
     // this lambda function equals zero
     // only if assim satisfies both FvCB and Ball Berry model
-    auto check_assim_rate = [&](double assim) {
+    auto check_assim_rate = [=, &FvCB_res, &BB_res, &an_conductance, &Gs, &Ci](double const assim) {
```
leavesandlemmas, line 87: This is a good change.  I don't like having functions with side effects but at least its a bit more explicit now. I couldn't think of a better way to write this, not without changing the overall design of `c3photo.cpp`.

src/module_library/c3photo.cpp
```diff
@@ -78,16 +78,27 @@ photosynthesis_outputs c3photoC(
...
+        // The net CO2 assimilation is the smaller of the biochemistry-limited
+        // and conductance-limited rates. This will prevent the calculated Ci
+        // value from ever being < 0. This is an important restriction to
+        // prevent numerical errors during the convergence loop, but does not
+        // seem to ever limit the net assimilation rate if the loop converges.
+        an_conductance = conductance_limited_assim(Ca, gbw, Gs);  // micromol / m^2 / s
+
+        double const assim_adj =
+            std::min(assim, an_conductance);  // micromol / m^2 / s
```
leavesandlemmas, line 96: Did you find a situation where this affects the solution ? Or convergence?

src/module_library/ball_berry_gs.cpp
leavesandlemmas: Actually, so when running the script you shared, I get this error:
```
Error in evaluate_module(module_name, temp_input_quantities) :
  Caught exception in R_evaluate_module: Thrown in ball_berry_gs: Cs is not positive.
Calls: module_response_curve ... apply -> FUN -> evaluation_function -> evaluate_module
Execution halted
```

Are you not getting this error?

eloch216: Whoops. The `Catm` sequence should begin at 1, not 0. `Catm = 0` triggers the error. I have changed the code above and it should work now.

I think we could probably change the check to trigger on `Cs < 0`, and use infinite `gsw` when `Cs` is zero, as you had suggested before. I will test this out.

src/module_library/c3photo.cpp
eloch216: I think they are still initial guesses, because the values here will be passed to `conductance_limited_an` the first time `check_assim_rate` is called.

src/module_library/c3photo.cpp
eloch216: Yeah, this is easy to test. Just replace lines 95-96 with `double const assim_adj = assim;  // micromol / m^2 / s` to bypass the conductance limited assimilation. Doing this causes tons of `Cs is not positive` errors all over the place.

src/module_library/ball_berry_gs.cpp
eloch216: Update: I changed the error condition in the Ball-Berry code, so it is only triggered for negative `Cs`. Now the example code works with `Catm = 0`.

src/module_library/c3photo.cpp
leavesandlemmas: Oh right. Never mind

src/module_library/ball_berry_gs.cpp
leavesandlemmas: Okay cool.

leavesandlemmas:
I resolved the conflict.

eloch216:
> I resolved the conflict.

I will update the tests

eloch216:
Got some compiler warnings that are causing checks to fail:
```
Warning: Found the following significant warnings:
  module_library/secant_method.h:33:25: warning: 'secant_parameters::rtol' will be initialized after [-Wreorder]
  module_library/secant_method.h:29:20: warning:   'double secant_parameters::check' [-Wreorder]
  module_library/secant_method.h:36:5: warning:   when initialized here [-Wreorder]
```

leavesandlemmas:
I am good with merging this pull request into the original request. So whenever you are ready to merge.

eloch216:
> I am good with merging this pull request into the original request. So whenever you are ready to merge.

Awesome!

I am going to re-include the conductance limited assimilation rate as an output of `c3photoC`, `c4photoC`, `BioCro:c3_assimilation`, and `BioCro:c4_assimilation`, and then I will merge.

It's important to have this as an output -- the only basis for the claim that the conductance limited rate "does not seem to ever limit the net assimilation rate if the loop converges" is that I have compared the values of `An` and `An_conductance`, and I have always seen `An_conductance >> An`. But I don't know if this is truly always the case.

justinmcgrath:
Is it's not mentioned already, please add a comment somewhere that this change is based on whichever from the Sun et al. paper.

eloch216:
> Is it's not mentioned already, please add a comment somewhere that this change is based on whichever from the Sun et al. paper.

I think these changes are not based on the Sun et al. paper. They used Newton's method, and here we are using the secant method. Or would you say we were inspired by them to move away from the fixed point iteration method?

justinmcgrath:
> > Is it's not mentioned already, please add a comment somewhere that this change is based on whichever from the Sun et al. paper.
> 
> I think these changes are not based on the Sun et al. paper. They used Newton's method, and here we are using the secant method. Or would you say we were inspired by them to move away from the fixed point iteration method?

Yes, they point out the convergence issues with fixed point iteration with the PS equations that led to the changes here, and the secant method is very similar to Netwon's method conceptually, even if they differ substantially in implementation details.

leavesandlemmas:
>  Or would you say we were inspired by them to move away from the fixed point iteration method?

I would say the inspiration for the change still comes from that paper. Especially just solving for the Ci root only, and not temperature and Ci simultaneously. 

>  They used Newton's method, and here we are using the secant method.

Technically, they don't use Newton's method. Newton's method refers to evaluating the first derivative directly (or at least it should). They say they use Newton's method but what they describe is [Steffensen's method](https://en.wikipedia.org/wiki/Steffensen%27s_method), because they approximate the first derivative using that method's approximation.  Basically, the secant method and Steffensen's methods are both quasi Newton methods. And there's many quasi-newton methods, especially for function optimization.

On the one hand, Steffensen's method can achieve higher convergence rates per iteration, but at the cost of an additional function evaluation per iteration. The secant method is slightly faster in terms of best-case improvement per function evaluation. However, the secant method should be a bit more stable than steffensen's method.

PR #171, opened 2025-03-26
Alternate secant method code

eloch216:
I also found it a bit confusing that the `secant_parameters` struct included inputs and outputs. Here's one approach to addressing this, where I just added more arguments to `find_root_secant_method`, following a suggestion from Justin.

Another possibility could be to use a `struct` for returning the final value of `x`, the check, and the counter.

### DIFF
--- a/src/module_library/c3photo.cpp
+++ b/src/module_library/c3photo.cpp
@@ -139,15 +139,23 @@ photosynthesis_outputs c3photoC(
         Ca * gbw / dr_boundary                 // The maximum conductance-limited An, which occurs for gsw = infinity
     });                                        // micromol / m^2 / s
 
-    secant_parameters secpar{1000, 1e-12, 1e-12};
+    // Run the secant method
+    double Assim_check{};  // Will be modified by find_root_secant_method
+    size_t iterations;     // Will be modified by find_root_secant_method
 
-    // find_root_secant_method will update secpar as a side-effect
     double const co2_assim_rate = find_root_secant_method(
-        check_assim_rate, assim_guess_0, assim_guess_1, secpar);
+        check_assim_rate,
+        assim_guess_0,
+        assim_guess_1,
+        1000,
+        1e-12,
+        1e-12,
+        Assim_check,
+        iterations);
 
     return photosynthesis_outputs{
         /* .Assim = */ co2_assim_rate,              // micromol / m^2 / s
-        /* .Assim_check = */ secpar.check,          // micromol / m^2 / s
+        /* .Assim_check = */ Assim_check,           // micromol / m^2 / s
         /* .Assim_conductance = */ an_conductance,  // micromol / m^2 / s
         /* .Ci = */ Ci,                             // micromol / mol
         /* .Cs = */ BB_res.cs,                      // micromol / m^2 / s
@@ -155,7 +163,7 @@ photosynthesis_outputs c3photoC(
         /* .Gs = */ Gs,                             // mol / m^2 / s
         /* .RHs = */ BB_res.hs,                     // dimensionless from Pa / Pa
         /* .Rp = */ FvCB_res.Vc * Gstar / Ci,       // micromol / m^2 / s
-        /* .iterations = */ secpar.counter          // not a physical quantity
+        /* .iterations = */ iterations              // not a physical quantity
     };
 }
 
--- a/src/module_library/c4photo.cpp
+++ b/src/module_library/c4photo.cpp
@@ -125,18 +125,26 @@ photosynthesis_outputs c4photoC(
     double const assim_guess_0 = collatz_assim(0.4 * Ca_pa);
     double const assim_guess_1 = collatz_assim(Ca_pa);
 
-    secant_parameters secpar{1000, 1e-12, 1e-12};
+    // Run the secant method
+    double Assim_check{};  // Will be modified by find_root_secant_method
+    size_t iterations;     // Will be modified by find_root_secant_method
 
-    // find_root_secant_method will update secpar as a side-effect
     double const Assim = find_root_secant_method(
-        check_assim_rate, assim_guess_0, assim_guess_1, secpar);
+        check_assim_rate,
+        assim_guess_0,
+        assim_guess_1,
+        1000,
+        1e-12,
+        1e-12,
+        Assim_check,
+        iterations);
 
-    // unit change
-    double const Ci = Ci_pa / atmospheric_pressure * 1e6;  // micromole / mol
+    // Convert Ci units
+    double const Ci = Ci_pa / atmospheric_pressure * 1e6;  // micromol / mol
 
     return photosynthesis_outputs{
         /* .Assim = */ Assim,                       // micromol / m^2 /s
-        /* .Assim_cehck = */ secpar.check,          // micromol / m^2 / s
+        /* .Assim_check = */ Assim_check,           // micromol / m^2 / s
         /* .Assim_conductance = */ an_conductance,  // micromol / m^2 / s
         /* .Ci = */ Ci,                             // micromol / mol
         /* .Cs = */ BB_res.cs,                      // micromol / m^2 / s
@@ -144,6 +152,6 @@ photosynthesis_outputs c4photoC(
         /* .Gs = */ Gs,                             // mol / m^2 / s
         /* .RHs = */ BB_res.hs,                     // dimensionless from Pa / Pa
         /* .Rp = */ 0,                              // micromol / m^2 / s
-        /* .iterations = */ secpar.counter          // not a physical quantity
+        /* .iterations = */ iterations              // not a physical quantity
     };
 }
--- a/src/module_library/photosynthesis_outputs.h
+++ b/src/module_library/photosynthesis_outputs.h
@@ -15,7 +15,7 @@ struct photosynthesis_outputs {
     double Gs;                 //!< Stomatal conductance to water vapor (mol / m^2 / s)
     double RHs;                //!< Relative humidity at the leaf surface (dimensionless)
     double Rp;                 //!< Rate of photorespiration (micromol / m^2 / s)
-    int iterations;            //!< Number of iterations used by convergence loop
+    size_t iterations;         //!< Number of iterations used by convergence loop
 };
 
 #endif
--- a/src/module_library/secant_method.h
+++ b/src/module_library/secant_method.h
@@ -7,34 +7,6 @@
 // Helper function declarations
 inline bool is_close(double x, double y, double atol, double rtol);
 inline bool is_zero(double x, double atol);
-/**
- * @struct A struct to hold parameters related to the secant method.
- *
- * @param check The value of the function at its last evaluation. If a
- *              root has been found, this should equal zero to within
- *              tolerance.
- * @param counter The number of iterations completed at termination of the secant
- *              method. Zero indicates no iterations. A value of 10
- *              indicates 10 iterations were completed.
- * @param max_iter The total number of iterations allowed. The secant method
- *              will terminate when this number is reached.
- * @param atol The absolute tolerance used to test for floating point number
- *              equality; primarily used to test if a zero has been found:
- *              Terminate if `abs(check) < atol`
- * @param rtol The relative tolerance used to test if iterates are not
- *              improving: Terminate if `abs(x0 - x1) <= atol + rtol * abs(x1)`
- */
-struct secant_parameters {
-    double check = 0;
-    int counter = 0;
-    const size_t max_iter = 100;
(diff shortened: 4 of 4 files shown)


leavesandlemmas (approved):
I don't have any objective objections.

Well I suppose my stylistic/aesthetic preferences are a little different. BioCro has many functions with a dozen or so arguments, and I have found working with them difficult. It's harder to find where arguments are actually used and defined, when there's a lot of arguments getting passed around. I have found making changes trickier. However, it's probably better to defer to the majority's preference.

In addition, I think the initial design isn't all that great stylistically either.  Preferably outputs should be together, and separate from the inputs. And there's two set of inputs: `fun, x0, x1` are enough to generate a sequence of iterates. The inputs `max_iter`, `atol` and `rtol` are for deciding when to terminate. If I get the time, I might try for a better design.

You've got my approval to merge. Even if I do a rewrite, this would be the default if the rewrite is rejected.

justinmcgrath:
> In addition, I think the initial design isn't all that great stylistically either. Preferably outputs should be together, and separate from the inputs. And there's two set of inputs: `fun, x0, x1` are enough to generate a sequence of iterates. The inputs `max_iter`, `atol` and `rtol` are for deciding when to terminate. If I get the time, I might try for a better design.

A way to treat this more generically is for all root finding algorithms to have the form g(f), and that is created by some other function h(...) -> g(f), where what's goes in the ... depends on the details of the algorithm. Things like max_iter, atol and rtol would certainly be passed to the h() function. I don't see a way for it to work without also passing it the starting values, but conceptually that doesn't seem right to me.

Although this is a sort of purist way to program it, I don't know that it has any strong benefits. I like the logic of it more, but I do not see a likely scenario where the flexibility of it is necessary. This stuff is hard coded, and you don't need to pass around different versions of g(f)-formed functions dynamically. And the approach is certainly not common, with most people passing everything to a single function.

leavesandlemmas:
> > In addition, I think the initial design isn't all that great stylistically either. Preferably outputs should be together, and separate from the inputs. And there's two set of inputs: `fun, x0, x1` are enough to generate a sequence of iterates. The inputs `max_iter`, `atol` and `rtol` are for deciding when to terminate. If I get the time, I might try for a better design.
> 
> A way to treat this more generically is for all root finding algorithms to have the form g(f), and that is created by some other function h(...) -> g(f), where what's goes in the ... depends on the details of the algorithm. Things like max_iter, atol and rtol would certainly be passed to the h() function. I don't see a way for it to work without also passing it the starting values, but conceptually that doesn't seem right to me.

The trouble is that cpp functions aren't objects that can be returned by other functions. But function objects can be: 

Here's an implementation that does that:
```cpp
struct secant_method {
    
    const size_t max_iter;
    const double atol;
    const double rtol;

    secant_method(size_t max_iter, double atol, double rtol) :
        max_iter{max_iter}, atol{atol}, rtol{rtol} {}

    template<typename F>
    double operator()(F fun, double x0, double x1){
        /* secant method code */
        return solution\; 
    }

};
```

The constructor is the function `h`  and the call method is `g`.  Starting values have to be provided somewhere, and since those are problem dependent, they'd need to be passed with `fun`.  Function objects seem to be the best way to have higher order functions in C++. The only thing that's not great about this version is that the call method should return the last `check` and the iteration count it reached.  

Alternatively, the method specific arguments could be templated:

```cpp

template<typename F, size_t max_iter, double atol, double rtol>
double secant_method(F fun, double x0, double x1){
    /* secant method code */
    return solution; 
}

``` 
But this probably doesn't really reduce the number of arguments, since you'd have to specify the template arguments at the same time.  There might be some template type tricks that Im not aware of. 


General interfaces are tricky because root-solving methods different in what needs to be provided in each problem, and what parameters are used.

eloch216:
> > > In addition, I think the initial design isn't all that great stylistically either. Preferably outputs should be together, and separate from the inputs. And there's two set of inputs: `fun, x0, x1` are enough to generate a sequence of iterates. The inputs `max_iter`, `atol` and `rtol` are for deciding when to terminate. If I get the time, I might try for a better design.
> > 
> > 
> > A way to treat this more generically is for all root finding algorithms to have the form g(f), and that is created by some other function h(...) -> g(f), where what's goes in the ... depends on the details of the algorithm. Things like max_iter, atol and rtol would certainly be passed to the h() function. I don't see a way for it to work without also passing it the starting values, but conceptually that doesn't seem right to me.
> 
> The trouble is that cpp functions aren't objects that can be returned by other functions. But function objects can be:
> 
> Here's an implementation that does that:
> 
> ```c++
> struct secant_method {
>     
>     const size_t max_iter;
>     const double atol;
>     const double rtol;
> 
>     secant_method(size_t max_iter, double atol, double rtol) :
>         max_iter{max_iter}, atol{atol}, rtol{rtol} {}
> 
>     template<typename F>
>     double operator()(F fun, double x0, double x1){
>         /* secant method code */
>         return solution\; 
>     }
> 
> };
> ```
> 
> The constructor is the function `h` and the call method is `g`. Starting values have to be provided somewhere, and since those are problem dependent, they'd need to be passed with `fun`. Function objects seem to be the best way to have higher order functions in C++. The only thing that's not great about this version is that the call method should return the last `check` and the iteration count it reached.

Couldn't `secant_method` have members for `check` and `counter` that could be accessed by the user after it was run? Perhaps it would be better to implement this as a `class` rather than a `struct`, with `check` and `counter` private, and with getter functions for retrieving their values.

If you like the struct/class approach better, and want to write the implementation, I would be fine with using that instead of the version I made in this PR. To me the main thing is to avoid the `secant_parameters` struct that was used before. Let me know if you want to implement it -- otherwise I will merge this PR.

> Alternatively, the method specific arguments could be templated:

It's best to avoid templating things unless it's absolutely necessary. Scott R. has pointed out that templates can result in very unhelpful error messages. See this for an example: https://github.com/ebimodeling/biocro-dev/pull/445. I think the function `F` needs to be templated, but it's hard to imagine a scenario where the tolerances would be anything other than doubles.

(As an aside, I went through a "template phase" after reading some `boost` examples, which use templates very frequently. So I wrote those trig functions using templates, which later needed to be removed.)

leavesandlemmas:
> Couldn't `secant_method` have members for `check` and `counter` that could be accessed by the user after it was run? Perhaps it would be better to implement this as a `class` rather than a `struct`, with `check` and `counter` private, and with getter functions for retrieving their values.
>
> If you like the struct/class approach better, and want to write the implementation, I would be fine with using that instead of the version I made in this PR. To me the main thing is to avoid the `secant_parameters` struct that was used before. Let me know if you want to implement it -- otherwise I will merge this PR.

Okay I think you should go ahead and merge. Let's stick with this version for now. But I think we can rethink the design as part of #172   

A function object could store the counter and check as state and provide accessor methods. But that's very much not functional programming, since the state of the check and counter would only make sense if `operator()` is called first. And the values correspond to a specific call of the check and counter.  

This side effect could cause problems. While currently, we only call the object once per instantiation, but future versions of the photosynthesis modules could shuffle the code around. 

But it also occurred to me that the three inputs `max_iter` `atol` and `rtol` and the two outputs `check` and `counter` aren't specific to the secant method. Moreover the code that uses that information decides if convergence has been reached.  If we had a newton method, we would basically use the same code for convergence.

In essence, the secant method and newton method both generate a sequence $(x_0, y_0), (x_1, y_1), (x_2, y_2), \ldots$ where their respective update formulas provide the next $x_n$ and then $y_n = f(x_n)$. The convergence code only needs this sequence. Im thinking the code could be restructured along these lines, but it's probably worth having different methods.

justinmcgrath:
Classses as used in C++ are very heavyweight, and the above example has the complexity of classes but no benefit over using functions since it's not generic or abstracted. You could use a C++ ethos with abstract classes and could probably use something C like with void types and some expectations. I still don't see tangible benefits to this though. I don't find this, in my opinion, easier to understand or more logical than a function that accepts both "types" of arguments and I don't see where we'd use an abstract idea of a root finder.

justinmcgrath:
I asked an AI for a way to implement my general idea for this. The overall approach is constructors that accept arguments for the problem setup and return a generic root finding function, so g(...) -> h(f). This is what it came up with, prefacing functions with "g_" or "h_" to indicate what's what.

```c++
#include <iostream>
#include <functional>
#include <cmath>
#include <optional>
#include <tuple>

// Result type: root + convergence info
struct RootResult {
    double root;
    int iterations;
    double residual;
};

// Factory for Newton-Raphson
auto g_newton(double x0, double atol, double rtol, int max_iter) {
    return [=](std::function<double(double)> f) -> std::optional<RootResult> {
        double x = x0;
        for (int i = 0; i < max_iter; ++i) {
            double fx = f(x);
            if (std::abs(fx) < atol) 
                return RootResult{x, i, fx};  // Converged
            
            // Approximate derivative (or pass df/dx as another arg to g_newton)
            double dfx = (f(x + 1e-8) - fx) / 1e-8;
            if (dfx == 0) break;  // Prevent division by zero
            x -= fx / dfx;
        }
        return std::nullopt;  // No convergence
    };
}

// Factory for Bisection
auto g_bisection(double a, double b, double atol, int max_iter) {
    return [=](std::function<double(double)> f) -> std::optional<RootResult> {
        if (f(a) * f(b) >= 0) 
            return std::nullopt;  // Invalid interval

        for (int i = 0; i < max_iter; ++i) {
            double c = (a + b) / 2;
            double fc = f(c);
            if (std::abs(fc) < atol || std::abs(b - a) < atol) 
                return RootResult{c, i, fc};  // Converged

            (f(a) * fc < 0) ? (b = c) : (a = c);
        }
        return std::nullopt;  // No convergence
    };
}

int main() {
    auto f = [](double x) { return x * x - 2; };  // f(x) = x² - 2 (root ≈ √2)

    // Configure Newton-Raphson
    auto h_newton = g_newton(1.0, 1e-6, 1e-6, 100);
    if (auto result = h_newton(f)) {
        std::cout << "Newton: root=" << result->root 
                  << ", iterations=" << result->iterations 
                  << ", residual=" << result->residual << "\n";
    }

    // Configure Bisection
    auto h_bisect = g_bisection(1.0, 2.0, 1e-6, 100);
    if (auto result = h_bisect(f)) {
        std::cout << "Bisection: root=" << result->root 
                  << ", iterations=" << result->iterations 
                  << ", residual=" << result->residual << "\n";
    }
}
```

I like the lack of classes, which is done by using lambdas and is essentially identical to how you'd do it in R, but with some annoying syntax. I'm not familiar with lambdas in C++, and maybe other people find this easier to read.

I'm not sure about returning the struct. I asked for something more generic and what it suggested was worse.

I asked about the utility of this, and it was excited about it, but I'm still not sold. At any rate, this is lightweight enough that I'm less resistant to incorporating this approach, even if of little utility, than I am to a class based approach, as those seem to get deeply tied into the program and hard to change or remove.

The full "conversation" is here: https://www.perplexity.ai/search/i-want-to-talk-through-an-idea-gh1VNn9JTX2Je0qlCSnlAA

leavesandlemmas:
This code doesn't compile, using this c++ shell: https://cpp.sh/  
1. `std::optional` is a c++ 17  addition, and deduced return types are a C++14 extension, to quote the compiler error. 
2. It doesn't compile in c++ 17 standard either. I get an error `cannot assign to a variable captured by copy in a non-mutable lambda`. 

I rewrote it like this:

```cpp
// Factory for Bisection
auto g_bisection(double atol, int max_iter) {
    return [=](std::function<double(double)> f, double a, double b) -> std::optional<RootResult> {
        if (f(a) * f(b) >= 0) 
            return std::nullopt;  // Invalid interval

        for (int i = 0; i < max_iter; ++i) {
            double c = (a + b) / 2;
            double fc = f(c);
            if (std::abs(fc) < atol || std::abs(b - a) < atol) 
                return RootResult{c, i, fc};  // Converged
            (f(a) * fc < 0) ? (b = c) : (a = c);
        }
        return std::nullopt;  // No convergence
    };
}
```

Which then does work. The typing seems to be okay. The std::optinal would have to be dropped. I think the AI was trying to generate a c++ version of the maybe monad using std::optional. 

I don't see any way around the fact that the result would need to be a struct similar. Basically the solving method should return all three, and then the unneeded information can be discarded. There would be some unnecessary copying but it's a  int and a float so not a big deal. 

However, in my opinion, this design is basically the same as the classes, although I suppose it's a bit less wordy than the struct/class definitions. Lambda functions in C++ are implemented as function objects. IIRC the compiler generates a struct with some unique type name and the struct has a `operator()` method with its return type deduced from the expression. Outside of the variable capture, I don't think lambda functions are that hard to read; the syntax is wordier than the syntax in haskell or python, but that's basically true for everything in c++.

Issue #168, opened 2025-03-19
Boost Euler and RK45 ode solvers fail on default soybean-biocro.


leavesandlemmas:
I noticed the other day that both of these calls to `run_biocro` fail with the `boost_euler` and the `boost_rk4` method both fail. They print the following error:

```
The ODE solver reports the following:
boost::numeric::odeint::integrate_const encountered an error and has returned a partial result:
Max number of iterations exceeded (-2147483648).
The dynamical system reports the following:
1 derivatives were calculated
```

This script will reproduce the error. 

```r
library(BioCro)

# fails
euler_result <- run_biocro(
  soybean$initial_values,
  soybean$parameters,
  soybean_weather$`2006`,
  soybean$direct_modules,
  soybean$differential_modules,
  within(default_ode_solvers$boost_euler, {output_step_size <- 1}),
  verbose=TRUE
)

# fails
rk4_result <- run_biocro(
  soybean$initial_values,
  soybean$parameters,
  soybean_weather$`2006`,
  soybean$direct_modules,
  soybean$differential_modules,
  within(default_ode_solvers$boost_rk4, {output_step_size <- 1}),
  verbose=TRUE
)

# success
euler_result <- run_biocro(
  soybean$initial_values,
  soybean$parameters,
  soybean_weather$`2006`,
  soybean$direct_modules,
  soybean$differential_modules,
  within(default_ode_solvers$homemade_euler, {output_step_size <- 1}),
  verbose=TRUE
)

```

Let me know if these aren't working on your BioCro installation.

eloch216:
I also had failures when running the examples.

I think they must be related to having `adaptive_max_steps` set to `NA` in the `boost_euler` and `boost_rk4` solver lists. It seems that `NA` is being misinterpreted as `-2147483648` when it is passed to the C++ code, so these solvers automatically fail. 

I verified that this is the case by setting `adaptive_max_steps` to 1 (see code below). With this change, the solvers work.

I haven't looked at the code to figure out why this is happening; it's a bit surprising since these non-adaptive solvers shouldn't care about the value of `adaptive_max_steps`. I would guess that this issue popped up after this PR: https://github.com/biocro/biocro/pull/149.

```R
# success
euler_result <- run_biocro(
  soybean$initial_values,
  soybean$parameters,
  soybean_weather$`2006`,
  soybean$direct_modules,
  soybean$differential_modules,
[...]

PR #173, opened 2025-03-27
Make sure all default solvers are working

eloch216:
This PR addresses issue https://github.com/biocro/biocro/issues/168.

I considered casting `adaptive_max_steps` to a `size_t` instead of an `int`. But, the `ode_solver` class expects an `int`, so this causes compiler warnings. Modifying the `ode_solver` class would require changing the C++ framework, which is complicated since it is a git submodule, so I would prefer to avoid doing this for now.

The new tests I added revealed another issue, where `adaptive_abs_error_tol` was misspelled as `adaptive_abs_error_tole` in two of the default solvers. Not sure how this happened, but now the mistake is fixed!


leavesandlemmas (approved):
Seems good!

PR #176, opened 2025-04-09
1d root solver library

leavesandlemmas:
This PR was originally created as a draft to help work out some weird bugs. I have updated the initial comment here to make the current version easier to understand,

### New PR Overview 

This PR implements a header library for solving 1D equations. Here's an example of how to use it. 

```cpp
#include <iostream>
#include "root_onedim.h"

// convenience function to print the result object.
std::ostream& operator<<(std::ostream& os, const root_algorithm::result_t& r){
// the result struct has four member fields: root, residual, iteration, and flag. 
    os << "root = " << r.root << '\n'; // the root or solution 
    os << "residual = " << r.residual << '\n'; //  residual = f(root) ; should be zero if a root was found successfully.
// the residual can be used to detect an error
    os << "iteration = " << r.iteration << '\n'; // number of iterations at termination; convergence is checked before the first iteration, so zero means the initial info is either a solution or an invalid state.

// r.flag is a value of type `root_algorithm::Flag` and is implemented as a scoped enum; it indicates the error states. 
// A function that checks if the flag indicate a successful termination, i.e., probable convergence.
    auto s =  root_algorithm::successful_termination(r.flag) // Does the algorithm think it found a root ? 
        ? "Termination was successful"
        : "Termination was unsuccessful" ;

    os << s << '\n';
// the flags are values of a scoped enum type; which basically means they are integers. This function returns a string describing // the flag's meaning. 
    os << "    " << root_algorithm::flag_message(r.flag) << "\n\n"; // Why did the algorithm terminate ? 
// the library won't throw exceptions, but will detect and signal errors. You decide how to handle the error. 
// I think compiler optimizations should keep the extra overhead from the error flags very low, and lower than using strings.   
    return os;
}

// Example Usage:
// Example problem function 
double f(double x){
    return (x + 1) * (x - 1);
}

int main() {

    // declare a solver object with the desired method as a template argument.
    // arguments set the maximum iteration count and the tolerances to use.  
    root_algorithm::root_finder<secant> solver(500, 1e-12, 1e-12);

    // the `solve` method accepts a function, or function object (lambda or class with the call method operator())
    // `easy_test{}` is a function object. 
    // the `solve` method accepts method-specific arguments for the problem. 
    root_algorithm::result_t result = solver.solve(f, 0, 3); // secant method, first two guesses are 0, 3. 



    std::cout << "==================================================\n";
    std::cout << "Easy Test-----------------------------------------\n";
    std::cout << result;

    // also accepts function objects or lambda functions

    auto g = [](double x) {return  x > 0 ? std::pow(x, 4) - 1 : -1; } ; 

    std::cout << "Hard Test-----------------------------------------\n";
    result = solver.solve(g, 0, 5);
    std::cout << result;
    std::cout << "==================================================\n";
    
    std::cout << "Hard Test with the bisection method---------------\n";
    // declare a solver using a different algorithm; can't reuse same name in same namespace
    root_algorithm::root_finder<root_algorithm::bisection> 
        solver_2(500, 1e-12, 0);
    result = solver_2.solve(g, 0, 5);  // can resue result 
    std::cout << result; 
}

```
which prints 
```
==================================================
Easy Test-----------------------------------------
root = 1
residual = 0
iteration = 9
Termination was successful
    Residual is zero.

Hard Test-----------------------------------------
root = 0.0159872
residual = -1
iteration = 5
Termination was unsuccessful
    Change in guess is zero.

==================================================
Hard Test with the bisection method---------------
root = 1
residual = 9.09495e-13
iteration = 43
Termination was successful
    Bracket width is zero.
```

I have tried to abstract away most of the implementation details, so that the user-interface won't need to ever change. I may make some tweaks to the implementation. Or add a method or two. 

It's relatively easy to switch between methods, you typically just need to change one line of code. Note you might need to change the starting guesses.  The bracketing methods like `bisection`, `ridder`, `regula_falsi` require the guesses to be a valid bracket (so one above and one below the root and the signs of the function must be different). 

#### Troubleshooting 

Note that this library itself will not throw exceptions for any reason (unless the function you pass throws an exception). That means you have to check for failure, but it also means you can decide how to handle the failure states. Within the absolute tolerance, we should find `residual = 0` so if you don't get `residual = 0` then there's an issue.  Note  `residual != 0` and `iteration = 0` likely means the initial info is invalid.  If `iteration == max_iterations` then the algorithm reach the max iteration count.  You can also use `successful_termination` and `flag_message` to help troubleshoot. 

Convergence checks are a bit tricky. Many algorithms also check whether or not the change in the root's guess has become small and terminate then. However, this can cause a false-positive termination.  For example, the secant method terminates early for the hard test, because it converges very slowly in that problem (it will exhaust the 500 iterations if allowed). However, the secant method does have to check for guesses that get too close to avoid division by zero problems. 

The other tricky part is that the bracket methods also terminate when the bracket becomes small, but this can occur without the residual becoming zero, at a singularities and jump discontinuities in non-continuous functions. However, currently this counts as a successful termination, so use the `successful_termination` with care. 

#### A Warning 

 I suspect that there might be a bug in my code with the bracketing methods because the bisection method has trouble finding the correct solution in the `c3photo` and `c4photo` functions, when the secant method doesn't.  However, I can't seem to find the bug or reproduce it on a test problem. So use them with care for now. 

The new implementation of the secant method reproduces the correct behavior, so I would stick with that for now. It also typically faster than the bracketing methods, but in principle the bracketing methods should always converge if a solution exists.  But feel free to try out the other methods.  

#### Implementation details 

The `root_finder` class has a `solve` method contains the for loop for all methods and builds the `result_t` as output. In general, its serves partly as an interface. Because each algorithm might save different info between iterations and perform different updates, each algorithm provides the `state` type and functions defining the algorithm's iteration and initializaiton. No need for ABCs or virtual functions, if you pass something without valid methods, the compiler will raise an error.

Currently each algorithm must define the following methods/types with the type signatures:
```cpp
//holds whatever state saved between iterations.
 struct state;
//takes the initial problem info (e.g., a bracket, an
 // initial guess) and instantiates a value of type `state`. If the
 provided info is invalid, then return a flag indicating the error.
with_flag<state> initialize(F&& f, Args... args); // Args... args is swapped for whatever is required.
 //maps state to state. The actual formula of most methods is
// implemented here. If the iteration encounters an error, return the
// flag indicating the error.
// Some methods handle errors that other methods do not.
 with_flag<state>& iterate(F&& f, with_flag<state>& s);
// iterate is implemented like a compound assignment operator,
// updating in-place rather than using an immutable maping.
// It should be equivalent to with_flag<state> iterate(f, const state& s)
 // Where it maps a state to state, and adds a new flag if an error is
// encountered.

// checks the state to see if the algorithm has found a root to within
// tolerance; return the appropriate flag, else return `Flag::valid`
Flag check_convergence(const state& s, _abs_tol, _rel_tol);
// Extract a single value for the root.
double root(const state& s);
// Extract the residual from the state.
double residual(const state& s);
```

The flags which indicate whether or not a termination state has been reached and the nature of the termination state are added a template. This implementation was inspired by error handling in haskell / FP languages but I probably don't need to do it that way.  And there's no easy way or non verbose way to do functors / bind  / partial application which would make using a monad for the error handling simpler.  

Some changes I am considering (and happy to hear your thoughts):
+ Add brent's method / maybe dekker's method or other hybrid methods. 
+ Combine the `check_convergence` and `iterate` methods so that convergence checks happen during an iteration. Originally, it seemed simpler to have these separate. But error states can occur in the  iterate /update steps of many algorithms. Moreover, there's some efficiency to be gained: for instance, check `f(x) = 0` each time `f` is evaluated at a new point. And terminate if `x` is a root.
+ Allow the `root` and `residual` getters to use a `Flag` to decide what to return.   
+ make `state` struct inherit from flags since the current implementation doesn't use unflagged states. Originally `iterate` was a function of the state object and returned a state object.
+ Maybe replace the state with private member fields. 

--------
### Original

This pull request creates a 1D root finding library.  I have more plans but figured I should share it.

I am creating it as a draft pull request because the PR doesn't pass the tests; it fails the `c3_canopy` module test. But this seems like its a bug, so I would like @eloch216  or someone to run the test on their computer. Although I initially thought the culprit is my code, the `c3_assimilation` module test case passes, only the c3_canopy code doesn't. The problem is I didn't make any changes to that code.  

I started this branch from the secant method code PR and Ed changed the c3_canopy module test case last. 

```r
td <- "path to your module_test_case directory"
module_name <- "BioCro:c3_canopy"
tc <- BioCro::cases_from_csv(module_name, td)[[1]]

# Does this pass?
# returns character(0) if passed
test_module()

# If doesn't pass
outputs <- evaluate_module(module_name, tc$inputs)
sort_list <- function(xs)xs[order(names(xs))]
compare_outputs  <- function(xs,ys)cbind(sort_list(xs), sort_list(ys))
compare_outputs(tc$expected_outputs, outputs)

```

Running this script, I get:

```
                                 [,1]        [,2]       
canopy_assimilation_rate_CO2     0           0          
canopy_conductance               1           1.485794   
canopy_photorespiration_rate_CO2 0.004865616 2.58469    
canopy_transpiration_rate        -4.559601   -4.559601  
GrossAssim_CO2                   4.222135    0.006385396
```

Please run the above script, and let me know what you get.


eloch216:
> I am creating it as a draft pull request because the PR doesn't pass the tests; it fails the `c3_canopy` module test. But this seems like its a bug, so I would like @eloch216 or someone to run the test on their computer. Although I initially thought the culprit is my code, the `c3_assimilation` module test case passes, only the c3_canopy code doesn't. The problem is I didn't make any changes to that code.

It's totally reasonable for a change in `c3photoC` to alter the output of the `BioCro_c3_canopy` module, because that module ultimately uses `c3photoC`. I would flip this around and ask the question: Why _does_ the `BioCro:c3_assimilation` module pass its test cases?

Right now, there is just one test case for the `BioCro:c3_assimilation` module, and it is not very realistic. I tried adding some more test cases in PR https://github.com/biocro/biocro/pull/177. That branch is based off `develop`. When I run those test cases on this branch, I get a failure for the `BioCro:c3_assimilation` and `BioCro:c3_canopy` modules.

So I think that the outputs of `c3photoC` have changed on this branch, and we can see this in both the `BioCro:c3_assimilation` and `BioCro:c3_canopy` modules. It seems like a small change that only occurs in particular situations (low light or low CO2) and doesn't seem to influence the overall soybean simulation results. But nevertheless, a real change has occurred, as far as I can tell. I am a little surprised that there were no failures for `BioCro:c3_leaf_photosynthesis`, but I guess the new test cases for that module also don't cover the particular situation where the outputs have changed.

eloch216:
Running your script above, I get the same results you got:

```
                                 [,1]        [,2]       
canopy_assimilation_rate_CO2     0           0          
canopy_conductance               1           1.485794   
canopy_photorespiration_rate_CO2 0.004865616 2.58469    
canopy_transpiration_rate        -4.559601   -4.559601  
GrossAssim_CO2                   4.222135    0.006385396
```

leavesandlemmas:
@eloch216  Thanks, I guess I will have a look at those tests. In principle, this PR shouldn't modify the behavior at all, so it's probably an error in this PR. 

When I testing with a module response curve using C3 assimilation, I don't see any obvious errors in the assimilation rates, although judging from these scripts, the error might be in the calculation of the other variables.  And the code seems to perform correctly on test problems. 

One possible issue is that the `check_assim` function object is capturing the outputs of transpiration, stomatal conductance, etc. by reference, creating a side effect where the function isn't pure. So I am wondering if that's the issue.

eloch216:
> When I testing with a module response curve using C3 assimilation, I don't see any obvious errors in the assimilation rates, although judging from these scripts, the error might be in the calculation of the other variables. And the code seems to perform correctly on test problems.
> 
> One possible issue is that the `check_assim` function object is capturing the outputs of transpiration, stomatal conductance, etc. by reference, creating a side effect where the function isn't pure. So I am wondering if that's the issue.

This inspired me to look into the issue a little bit more. It turns out that for the `c3_assimilation` test case that fails, there is only a difference in the conductance-limited assimilation rate.

It's easy to check which branch is behaving properly, since we can directly calculate `conductance_limited_assim` from the `gbw` and `Gs` outputs from the `c3_assimilation` module, and compare it against the value of `conductance_limited_assim` from the module. It looks like this branch is returning consistent values, but `develop` is not.

So there is a real difference, but it's not a bug on this branch -- this branch actually fixes a mistake.

I added some more constancy checks on PR https://github.com/biocro/biocro/pull/177. Currently one of them needs to be skipped (the one for checking the conductance limited assimilation). When that PR is merged, I can update this branch, un-skip that test, and update the module test cases.

eloch216:
So I feel good about the tests now... but I have not looked at any of the content of this PR yet 😂

leavesandlemmas:
> So I feel good about the tests now... but I have not looked at any of the content of this PR yet 😂

No worries, you found the bug that i suspected might be outside the new code.

leavesandlemmas:
I guess I should explain the structure a little too. 

The basic usage is to instantiate a root finder object with the desired tolerances and maximum iterations, using the template argument to select the desired algorithm to find a root.  

Then, the root_finder object has a solve method which accepts the function whose root is sought, plus whatever problem & method-specific  initial information is needed. So for the secant method, this is two guesses, a bracket for the bisection method etc.  The solve method does the for-loop, and returns a result_t object when convergence is detected.  E.g.

```cpp
root_finder<secant> solver{100, 1e-12, 1e-12}; // max iterations, tolerances 
result_t result = solver.solve(f, x0, x1); 
```

Internally, each method is expressed as five functions, which the following type signatures (using Haskell style notation):
```hs
initialize :: initial_info -> state
iterate :: state -> state
has_converged :: state -> Bool
root :: state -> Double
residual :: state -> Double
``` 

`initialize` takes the initial problem info (like the two guesses in the secant method) and creates an object that holds the state that is kept from one iteration to the next.  `iterate` contains the meat of each method; it refines the guess, etc. `has_converged` checks if `state` has converged; this means the `state` object has to have whatever info is used to make this decision. `root` and `residual` extract the last best guess and the function's residual at that root from the state.

I am not 100% convinced that the last three functions can't be factored out at least partly. But for the time being, this allows for minimum code duplication. 

Methods are written as `struct` but they don't hold any state and no objects are ever constructed.  Instead they basically register the functions with the compiler, so that the `root_finder`  specialized to whatever method is selected at compile-time.

eloch216:
> @eloch216 Hey, so this PR passes an R CMD check on my computer, which is windows with R version 4.3.3, but fails the version 3.6.0 check with a compiler error. But I thought we used the same compiler flags for all version of R?

When using R version 3.6.0, it is also necessary to use an older version of RTools (Rtools35). Rtools35 uses an older version of the gcc compiler (4.9.3). The latest version of RTools uses gcc version 14.2.0, so sometimes there are different compiler messages on R 3.6.0.

>I also can't find whatever it thinks the error is. It's not a c++11 issue because the code definitely compiles (using g++ ).

Here is the error, I found it in the GitHub workflow output:

```
In file included from module_library/c3photo.cpp:8:0:
module_library/root_onedim.h: In instantiation of 'root_algorithm::with_flag<T>::with_flag(T) [with T = root_algorithm::two_step_method::state]':
module_library/root_onedim.h:175:35:   required from 'root_algorithm::with_flag<root_algorithm::two_step_method::state> root_algorithm::two_step_method::initialize(F&&, double, double) [with F = c3photoC(c3_temperature_response_parameters, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double)::<lambda(double)>&]'
module_library/root_onedim.h:133:79:   required from 'root_algorithm::result_t root_algorithm::root_finder<Method>::solve(F&&, Args&& ...) [with F = c3photoC(c3_temperature_response_parameters, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double)::<lambda(double)>&; Args = {const double&, const double&}; Method = root_algorithm::secant]'
module_library/c3photo.cpp:159:22:   required from here
module_library/root_onedim.h:44:48: error: cannot convert 'root_algorithm::two_step_method::state' to 'double' in initialization
     with_flag(T x) : state{x}, flag{Flag::valid} {}
```

It looks like there are also some other warnings related to `c3photoC`:

```
module_library/c3photo.cpp: In function 'photosynthesis_outputs c3photoC(c3_temperature_response_parameters, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double)':
module_library/c3photo.cpp:153:12: warning: unused variable 'Assim_check' [-Wunused-variable]
     double Assim_check{};  // Will be modified by find_root_secant_method
            ^
module_library/c3photo.cpp:154:12: warning: unused variable 'iterations' [-Wunused-variable]
     size_t iterations;     // Will be modified by find_root_secant_method
```

> BTW this PR is ready for review. I will convert to a full pull request after I finish organizing the file, adding more documentation, switching from the standalone secant_method to this library, etc.

If there is a compiler error and you're still changing the file organization and documentation, it doesn't really sound like it's ready for review!

eloch216:
> > When using R version 3.6.0, it is also necessary to use an older version of RTools (Rtools35). Rtools35 uses an older version of the gcc compiler (4.9.3). The latest version of RTools (Rtools45) uses gcc version 14.2.0, so sometimes there are different compiler messages on R 3.6.0.
> 
> So I should install Rtools35 ? Is there a way to switch between versions ? I have been looking at these compiler errors, but I don't understand where it's gone wrong.

You can have multiple installations of both R and RTools. I have R 3.6.0 and Rtools35 installed on my laptop so I can troubleshoot these kinds of issues locally.

It can be a little tricky to have multiple versions of R. One issue is that each version generally needs to have its own directory for installed packages, since they may require different versions of some packages.

I have a `.Rprofile` file with some code that helps automate this:

```R
rlib_name <- paste0(
    '~/RLib-',
    with(R.Version(), {paste(major, minor, sep = '.')})
)

Sys.setenv(R_LIBS_USER=rlib_name)

if (!dir.exists(rlib_name)) {
    dir.create(rlib_name)
}

.libPaths(c(rlib_name, .libPaths()))
```

This will make sure each R version uses its own package directory.

Another thing is that new versions of R and Rtools are "smart" enough that R can locate Rtools. But for older versions, including 3.6.0 and Rtools35, Rtools needs to be on the PATH. I wouldn't recommend actually adding it to your system path. Instead, you can use the following command in R to temporarily add it:

```R
Sys.setenv(PATH = paste("C:/Rtools/bin", Sys.getenv("PATH"), sep=";"))
```

Then you will be able to install BioCro in R version 3.6.0.

eloch216:
> Okay, I got that to work. The Rtools43 was being added by my .profile file, which is now empty. So I have been able to reproduce the same error message.

Nice!
 
> But I don't understand what it doesn't like; it's something about the `with_flag` template initializers but I don't understand why it says it can't convert from `state` to `double` this shouldn't ever occur. And if it's a syntax error, why does it fail on this version but not with the newer version of Rtools? I am wondering if it isn't a compiler bug.

I haven't tried looking into it. But even if it is a compiler bug, that doesn't really matter -- we can't change the compiler used with R 3.6.0, so the code will have to change to address the error.

leavesandlemmas:
I can't believe this fixed it. I changed this:

```cpp
template <typename T>
struct with_flag {
    T state;
    Flag flag;

    with_flag() {};
    with_flag(T x) : state{x}, flag{Flag::valid} {}
    with_flag(T x, Flag f) : state{x}, flag{f} {}
    with_flag(Flag f) : state{}, flag{f} {}

};
```
to this:

```cpp
template <typename T>
struct with_flag {
    T state;
    Flag flag;

    with_flag() {};
    with_flag(T x) : state(x), flag(Flag::valid) {}
    with_flag(T x, Flag f) : state(x), flag(f) {}
    with_flag(Flag f) : flag(f) {}

};
```

and it compiled and passed all the checks... so something about using {} versus () to initialize the objects. I don't fully understand the difference TBH. But it works now.

leavesandlemmas:
Though  I broke my R installations somehow.

eloch216:
> Though I broke my R installations somehow.

In what way were they broken? Have you been able to fix them?

BTW, when PR https://github.com/biocro/biocro/pull/178 is done, I will update this branch from `develop`. I think there may be a few merge conflicts.

leavesandlemmas:
> > Though I broke my R installations somehow.
> 
> In what way were they broken? Have you been able to fix them?

There was a weird error where the linker was saying it didn't recognize the file format of the object files. And i somehow deleted rtools43, and then downloaded rtools45 which doesn't work. 

But i think i fixed everything.  I don't fully understand the braces vs parentheses in initializing objects.  But the code compiles now.  


> BTW, when PR https://github.com/biocro/biocro/pull/178 is done, I will update this branch from `develop`. I think there may be a few merge conflicts.

No worries.

eloch216:
It took me a little while to understand this, but I think it all makes sense. I can understand how it reduces duplicated code.

A few general comments:
- Please run `clang-format`.
- The documentation definitely needs to be updated -- it is unclear or wrong in a few places. I found your comment above (that begins with "I guess I should explain the structure a little too") to be helpful and it would be good to incorporate some of this into the code documentation.
- I'm not sure if I find the "testing" function objects to be very helpful as-is. I think it would be really nice to make a module that solves one of these with each algorithm and returns all the results. The outputs could be qualified by the algorithm name, like `secant_root`, `secant_residual`, `secant_iterations`, `secant_valid`, etc. Maybe there could also be an input whose value is used in the function. E.g. for the "easy test" the user could pass in an offset instead of having it fixed to 1. This would be a great example of how to use the solvers, and a good test to make sure they are all working.

I will have a few more specific comments that I'll make at relevant spots.

I tried to see if I could figure out why that error was occurring for R v3.6.0 but couldn't really understand it either. But I guess it's okay as long as it works!

src/module_library/c3photo.cpp
```diff
@@ -150,22 +150,17 @@ photosynthesis_outputs c3photoC(
     });                                        // micromol / m^2 / s
 
     // Run the secant method
-    double Assim_check{};  // Will be modified by find_root_secant_method
-    size_t iterations;     // Will be modified by find_root_secant_method
-
-    double const co2_assim_rate = find_root_secant_method(
+    // double Assim_check{};  // Will be modified by find_root_secant_method
+    // size_t iterations;     // Will be modified by find_root_secant_method
```
eloch216, line 154: Delete these lines instead of commenting them out, since they aren't used anymore.

src/module_library/c3photo.cpp
```diff
@@ -5,7 +5,7 @@
 #include "c3_temperature_response.h"    // for c3_temperature_response
 #include "conductance_limited_assim.h"  // for conductance_limited_assim
 #include "FvCB_assim.h"                 // for FvCB_assim
-#include "secant_method.h"              // for find_root_secant_method
```
eloch216, line 8: I think `secant_method.h` needs to be deleted now?

src/module_library/root_onedim.h
```diff
@@ -0,0 +1,515 @@
...
+ * @brief Result from root finding algorithm.
+ *
+ * @param root The identified root.
+ *
+ * @param [in] residual The output value (root); should equal zero if
+ * successful.
+ *
+ * @param [in] iteration The number of iterations performed.
+ *
+ * @return result_t
```
eloch216, line 67: Not sure it makes sense to have documentation for a return value since this is not a function. I don't think the `residual` and `iteration` members are input arguments, either.

src/module_library/root_onedim.h
```diff
@@ -0,0 +1,515 @@
...
+ * floating-point zero. E.g., abs_tol = 1e-12 means |x| < 1e-12, or x equals
+ * zero to 12 digits. Used for convergence testing.
+ *
+ * @param [in] rel_tol The relative tolerance, the error threshold for
+ * comparing floating point number equality. E.g., rel_tol = 1e-12 means
+ * x equals y if the first 12 digits match. Used for convergernce testing.
+ *
+ * @return result_t
+ *
+ * @details This class creates a function object whose call method (operator())
```
eloch216, line 97: The call method is `solve()`, not `operator()`, right?

src/module_library/root_onedim.h
```diff
@@ -0,0 +1,515 @@
...
+ */
+template <typename Method>
+struct root_finder : public Method {
+
+    size_t max_iterations = 100;
+    double _abs_tol = std::numeric_limits<double>::epsilon();
+    double _rel_tol = 2 * std::numeric_limits<double>::epsilon();
+
+    root_finder() {}
+    root_finder(size_t max_iter, double abs_tol, double rel_tol) : max_iterations{max_iter}, _abs_tol{abs_tol}, _rel_tol{rel_tol} {}
```
eloch216, line 126: This line is too long

src/module_library/root_onedim.h
```diff
@@ -0,0 +1,515 @@
...
+            std::forward<F>(func), std::forward<Args>(args)...);
+
+        for (size_t i = 0; i < (max_iterations + 1); ++i) {
+            if (is_valid(s)) {
+                s.flag = Method::check_convergence(s.state, _abs_tol, _rel_tol);
+            }
+
+            if (s.flag != Flag::valid) {
+                return make_result(s, i);
+            }
```
eloch216, line 143: I think `s.flag != Flag::valid` is equivalent to `!is_valid(s)`. So I think this could be simplified into an `if-else` block:
```c++
            if (is_valid(s)) {
                s.flag = Method::check_convergence(s.state, _abs_tol, _rel_tol);
            } else {
                return make_result(s, i);
            }
```

src/module_library/root_onedim.h
```diff
@@ -0,0 +1,515 @@
...
+        return state{s.current, new_guess};
+    }
+
+};
+
+struct steffensen : public two_step_method {
+
+
+    template <typename F>
+    with_flag<state> initialize(F&& fun, double x0)
```
eloch216, line 252: I think the Newton and Steffenson classes have the same `initialize` method? Can this duplication be eliminated? And if they need to overwrite the default `two_step_method::initialize` method, are they really two step methods?

src/module_library/root_onedim.h
```diff
@@ -0,0 +1,515 @@
...
+};
+
+struct secant : public two_step_method {
+
+
+    template <typename F>
+    with_flag<state> iterate(F&& fun, const state& s)
+    {
+        double secant_slope = (s.current.y - s.last.y) / (s.current.x - s.last.x);
+        if (secant_slope == 0.0) {
```
eloch216, line 213: Usually this would check for a very small value instead of exact equality with zero. There are a few other cases like this in the code as well.

src/module_library/root_onedim.h
leavesandlemmas: Unfortunately, it can't be simplified into a single if-else.  If the `iterate` method returns an invalid flag, then it doesn't make sense to check the convergence (and overwrite that invalid flag).  The second if-statement checks if for any termination flag (either convergence or invalid state).

src/module_library/root_onedim.h
leavesandlemmas: It depends on whether the change in `x` is used to decide convergence. Several descriptions of these algorithms use change in `x` at each iteration to decide whether or not to terminate. But sometimes this causes a false positive if the function is flat enough to cause very small steps per iteration and thus converges very slowly. The algorithm will decide it's near a root when it's not.

If I remove the small_step as a convergence check, then they do become one step methods.  

Brent's method basically checks for whether the step size is too small and does a bisection.

leavesandlemmas:
@eloch216 

I have added better documentation.  For now, I am holding off on adding Brent's method; there's a numerical recipe for it and I also have the original paper. But it's a bit tricky to understand and adapt.

Unfortunately, I think I should change the structure so that the `check_convergence` and `iterate` do updates in place rather than returning something. In FP, functions aren't supposed to do in-place updates since values are supposed to be immutable. But in FP languages, the compiler might do optimizations that remove the implicit copying required. C++ doesn't really do that though; so this code is probably less performant than the numerical recipe codes.

Also this commit fails to compile because when I swapped out the secant_method in the in `c4photo.cpp` code,  I get a compiler error that says the Flags are defined multiple times. I'd like your guidance on fixing this issue, since it means we either need to convert the c4photo.cpp and c3photo.cpp into header files or take the flags and put them in their own source file.

leavesandlemmas:
I will add more detailed descriptions of the methods. I just added a `@brief` tag which describes what needs to be passed to the solver.

eloch216:
Why are the tests failing now? They were passing yesterday after I updated this branch from `develop`.

src/module_library/root_onedim.h
eloch216: Sounds good!

eloch216:
> @eloch216 Initially, there was a linker error where the `successful_termination` and `flag_message` functions were defined in multiple object files (even with the header guards). This wasn't an issue until I changed the `c4photo.cpp` file, initially I hadn't changed that one in case there was a stupid bug. So it would compile correctly, but the linking would fail.
>
> Now, the `assim_conductance` fails for `c4_assimilation`. I'm not sure what the issue is. I think probably the expected value just needs to be updated. Did you change this ?

Oh, I see. We already needed to update the C3 module tests because of a change in `assim_conductance` (https://github.com/biocro/biocro/pull/176#issuecomment-2794960611) so I think the tests just need to be updated again after switching from `find_root_secant_method` in `c4photo`. The new implementation of the secant solver on this branch seems to fix a bug that is causing incorrect values of `assim_conductance` on the `develop` branch.

src/module_library/root_onedim.h
eloch216: Could the second check use `if (!is_valid(s))` though?

leavesandlemmas: yes,  and I've made that change but haven't committed and pushed it to github b/c of the failing tests

eloch216:
Did you use a code styler for `module_testing.R`? There seem to be a lot of style changes to that file.

leavesandlemmas:
> Did you use a code styler for `module_testing.R`? There seem to be a lot of style changes to that file.

Not intentionally, I think I installed some R support extension for VS code that automatically did that `'` with `"`. I think I've turned it off now. I only wanted syntax highlighting. 

The main change was I made a failing test print out the discrepancy in the expected outputs.

eloch216:
> > Did you use a code styler for `module_testing.R`? There seem to be a lot of style changes to that file.
> 
> Not intentionally, I think I installed some R support extension for VS code that automatically did that `'` with `"`. I think I've turned it off now. I only wanted syntax highlighting.
> 
> The main change was I made a failing test print out the discrepancy in the expected outputs.

This is a good idea, but I think the formatting is not quite right. For each quantity where there is disagreement, the module name, test case name, etc get repeated. I noticed this by intentionally creating some failures for the `BioCro:ten_layer_c3_canopy` module:

```r
Failure ('test.Modules.R:5:5'): All modules are functioning correctly
`test_module_library(...)` threw an unexpected error.
Message: Problems occurred while testing modules from the `BioCro` library:
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_0    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_1    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_2    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_3    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_4    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_5    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_6    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_7    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_8    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Assim_layer_9    -0.0638519565916136     -2.6087651216984
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_0       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_1       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_2       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_3       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_4       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_5       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_6       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_7       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_8       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Ci_layer_9       300.115230856943        12222.6794228289
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_0       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_1       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_2       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_3       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_4       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_5       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_6       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_7       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_8       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_Cs_layer_9       300.013067726397        12217.1140572359
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_0  -7.01568763793061       -7.01568215178524
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_1  -7.01568763793061       -7.01568215178524
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_2  -7.01568763793061       -7.01568215178524
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_3  -7.01568763793061       -7.01568215178524
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_4  -7.01568763793061       -7.01568215178524
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_5  -7.01568763793061       -7.01568215178524
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_6  -7.01568763793061       -7.01568215178524
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_7  -7.01568763793061       -7.01568215178524
  Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: calculated outputs do not match expected outputs   
        Expected        Actual  
shaded_EPenman_layer_8  -7.01568763793061       -7.01568215178524
  Mod
Class:   simpleError/error/condition
```

src/module_library/root_onedim.h
```diff
@@ -0,0 +1,697 @@
...
+ * @param [in] abs_tol The absolute tolerance, the error threshold for
+ * floating-point zero. E.g., abs_tol = 1e-12 means |x| < 1e-12, or x equals
+ * zero to 12 digits. Used for convergence testing.
+ *
+ * @param [in] rel_tol The relative tolerance, the error threshold for
+ * comparing floating point number equality. E.g., rel_tol = 1e-12 means
+ * x equals y if the first 12 digits match. Used for convergernce testing.
+ *
+ * @details This class creates a function object whose solve or call method
+ * is an interface for using any of the root finding methods. A class template
```
eloch216, line 98: Once an object of this class has been created, it is an interface for one of the root finding methods, not any of the root finding methods.

I would suggest rewording this to something like

"This class standardizes the interface to each of the root finding methods."

Or

"When instantiated, this class is a function object whose solve method is an interface to one of the root finding methods."

leavesandlemmas:
> This is a good idea, but I think the formatting of the error message is not quite right. For each quantity where there is disagreement, the module name, test case name, etc get repeated. I noticed this by intentionally creating some failures for the `BioCro:ten_layer_c3_canopy` module:

Do you have some advice on how to change the code to improve the formatting? 

I have been futzing with it and can't seem to get a nice table.

eloch216:
> > This is a good idea, but I think the formatting of the error message is not quite right. For each quantity where there is disagreement, the module name, test case name, etc get repeated. I noticed this by intentionally creating some failures for the `BioCro:ten_layer_c3_canopy` module:
> 
> Do you have some advice on how to change the code to improve the formatting?
> 
> I have been futzing with it and can't seem to get a nice table.

This stuff can be pretty annoying. I can give it a try and let you know what I come up with.

leavesandlemmas:
@eloch216  thanks

leavesandlemmas:
> - it could be nice to include fixed-point iteration as an option, so its performance could be compared to the better algorithms.

I have a question about this. Normally, for fixed point iteration, we've got a function object `double f(double x)` that has a fixed point `x` such that `f(x) == x`.  And we just repeatedly call `f` until convergence or max iterations.  But if let `g(x) = f(x) - x` or `g(x) = x - f(x)` then the fixed point of `f` is a zero of `g`.  If I implement a fixed point iteration method, which should it expect as an input `f` or `g` ? 

I am thinking it would probably have to be `f`, even though it would break some of the interface's logic, because there's two possible `g`.

leavesandlemmas:
> Some of my comments from earlier days still have not been addressed, such as running `clang-format` and adding a test to make sure all the root finders are working. (I think a module would be an easy way to do this, since it will be tested against stored results during the module tests without needing to write any dedicated testing stuff.) The old version of the secant finder should also be removed.

I have been working on a test module, but I seem to be having trouble with the module, working properly. I also have this file I used for testing, which can be compiled separately from biocro (so it compiles very fast). Should we consider a c++ library for unit tests of the c++ library? 

Maybe I could adapt the `run_test` function into a module. 🤔 

```cpp
#include <iostream>
#include "../biocro/src/module_library/root_onedim.h"

using namespace root_algorithm;

std::ostream &operator<<(std::ostream &os, const result_t &r);

struct hard_test
{
    double operator()(double x)
    {
        return x < 0 ? -1 : std::pow(x, 10) - 1;
    }

    double derivative(double x)
    {
        return x < 0 ? 0.0 : 10 * std::pow(x, 9);
    }

    double second_derivative(double x)
    {
        return x < 0 ? 0.0 : 90 * std::pow(x, 8);
    }
};

struct linear
{
    double operator()(double x)
    {
        return (1 - x);
    }

    double derivative(double x)
    {
        return -1;
    }

    double second_derivative(double x)
    {
        return 0;
    }
};

struct quadratic
{
    double operator()(double x)
    {
        return (1 - x) * (x + 1);
    }

    double derivative(double x)
    {
        return -2 * x;
    }

    double second_derivative(double x)
    {
        return -2;
    }
};

struct kepler
{
    double epsilon;
    double answer;
    double y;

    kepler(double ep, double ans) : epsilon{ep}, answer{ans}
    {
        y = ans - ep * std::sin(ans);
    }

    double operator()(double x)
    {
        return x - epsilon * std::sin(x) - y;
    }

    double derivative(double x)
    {
        return 1 - epsilon * std::cos(x);
    }

    double second_derivative(double x)
    {
        return epsilon * std::sin(x);
    }
};

struct double_root
{
    double operator()(double x)
    {
        return std::pow(x - 1, 2);
    }

    double derivative(double x)
    {
        return 2 * (x - 1);
    }

    double second_derivative(double x)
    {
        return -2;
    }
};

struct triple_root
{
    double operator()(double x)
    {
        return std::pow(x - 1, 3);
    }

    double derivative(double x)
    {
        return 3 * std::pow(x - 1, 2);
    }

    double second_derivative(double x)
    {
        return 6 * (x - 1);
    }
};

struct step_function
{
    double operator()(double x)
    {
        return x >= 1 ? 1 : -1;
    }

    double derivative(double x)
    {
        return 0;
    }

    double second_derivative(double x)
    {
        return 0;
    }
};

struct singularity
{
    double operator()(double x)
    {
        return 1 / (x - 1);
    }

    double derivative(double x)
    {
        return -1 / std::pow(x - 1, 2);
    }

    double second_derivative(double x)
    {
        return 2 / std::pow(x - 1, 3);
    }
};

// double f(double x){return (2 - x) * (x + 2);}

template <typename Method, typename... Args>
void run_test(root_finder<Method> solver, Args &&...args)
{

    std::cout << "==================================================\n";
    std::cout << "Linear Test---------------------------------------\n";
    auto result = solver.solve(linear{}, std::forward<Args>(args)...);
    std::cout << result;

    std::cout << "Quadratic Test------------------------------------\n";
    result = solver.solve(quadratic{}, std::forward<Args>(args)...);
    std::cout << result;

    std::cout << "Kepler Test--------------------------------------\n";
    result = solver.solve(kepler{0.99, 1}, std::forward<Args>(args)...);
    std::cout << result;

    std::cout << "Hard Test-----------------------------------------\n";
    result = solver.solve(hard_test{}, std::forward<Args>(args)...);
    std::cout << result;

    std::cout << "Double Root---------------------------------------\n";
    result = solver.solve(double_root{}, std::forward<Args>(args)...);
    std::cout << result;

    std::cout << "Triple Root---------------------------------------\n";
    result = solver.solve(triple_root{}, std::forward<Args>(args)...);
    std::cout << result;

    std::cout << "Step Function-------------------------------------\n";
    result = solver.solve(step_function{}, std::forward<Args>(args)...);
    std::cout << result;

    std::cout << "Singularity--------------------------------------\n";
    result = solver.solve(singularity{}, std::forward<Args>(args)...);
    std::cout << result;
    std::cout << "==================================================\n";
}

// convenience function
std::ostream &operator<<(std::ostream &os, const result_t &r)
{
    os << "root = " << r.root << '\n';           // the root or solution
    os << "residual = " << r.residual << '\n';   //  residual = f(root) ; should be zero if a root was found successfully.
                                                 // the residual can be used to detect an error
    os << "iteration = " << r.iteration << '\n'; // number of iterations at termination; convergence is checked before iteration, so zero means the initial info is either a solution or an invalid state.

    // r.flag is a value of type `root_algorithm::Flag` and is implemented as a scoped enum; it indicates the error states.
    auto s = successful_termination(r.flag) // Does the algorithm think it found a root ?
                 ? "Termination was successful"
                 : "Termination was unsuccessful";

    os << s << '\n';
    os << "    " << flag_message(r.flag) << "\n\n"; // Why did the algorithm terminate ?
                                                    // the library won't throw exceptions, but will detect and signal errors. You decide how to handle the error.
                                                    // I think compiler optimizations should keep the extra overhead from the error flags very low, and lower than using strings.
    return os;
}

int main()
{

    root_finder<bisection> solver(500, 1e-12, 1e-12);
    run_test(solver, 0, 3);
}
```

eloch216:
> > * it could be nice to include fixed-point iteration as an option, so its performance could be compared to the better algorithms.
> 
> I have a question about this. Normally, for fixed point iteration, we've got a function object `double f(double x)` that has a fixed point `x` such that `f(x) == x`. And we just repeatedly call `f` until convergence or max iterations. But if let `g(x) = f(x) - x` or `g(x) = x - f(x)` then the fixed point of `f` is a zero of `g`. If I implement a fixed point iteration method, which should it expect as an input `f` or `g` ?
> 
> I am thinking it would probably have to be `f`, even though it would break some of the interface's logic, because there's two possible `g`.

I was thinking the opposite -- that it's better to keep a consistent interface, where the user supplies a function whose root should be found.

I would just expect the user to supply $g(x) = f(x) - x$. This can be clarified in the documentation. Then the fixed point rule $x_{n+1} = f(x_n)$ becomes $x_{n+1} = g(x_n) + x_n$.

If it seems like too much of a bother, we don't need to add the fixed-point method at all.

> I have been working on a test module, but I seem to be having trouble with the module, working properly. 

What's going wrong with it? 

> Maybe I could adapt the `run_test` function into a module. 🤔

The module doesn't need to be complicated. The goal is not to illustrate which type of problem each method is best for, or test each algorithm against tricky functions. We just need one function to test, and then run each method. This makes sure that the code can run, and it will let us know if any outputs change.

I think the `run_test` below is more complicated than the module needs to be. Modules also shouldn't print any results (i.e. they should never need `#include <iostream>`).

> Should we consider a c++ library for unit tests of the c++ library?

This would be a nice thing to have, but would be more difficult and time-consuming to create by at least one order of magnitude compared to making a module. This PR is not the right place to do it.

eloch216:
Do you want to update from `develop`? Otherwise I can do it.

Last time, I did it through GitHub desktop. There will be conflicts regarding test cases; for these, it's easiest to just choose "use the version from `develop`." Then run the tests locally, and update case files as necessary with `update_csv_cases` until they are all passing.

leavesandlemmas:
Got it

eloch216:
This looks pretty good. The code is a little simpler and easier to follow now that the `Flag` is a member of the `result_t` struct. Just a few comments:
- The old `secant_method.h` still needs to be deleted.
- The documentation comment for `root_finder` doesn't list all of the available methods. I didn't do a careful check, but I noticed that `pegasus` and `anderson_bjorck` are missing.
- Likewise, the `root_onedim_test` module does not use all the available methods.

Once these are addressed, I think this will be ready to go, unless you have other changes in mind.

(Side note -- I see you updated from `develop`. There may be some test failures after doing this.)

leavesandlemmas:
> This looks pretty good. The code is a little simpler and easier to follow now that the `Flag` is a member of the `result_t` struct. Just a few comments:

Thanks 

 
>     * The old `secant_method.h` still needs to be deleted.

Addressing now
 
>     * The documentation comment for `root_finder` doesn't list all of the available methods. I didn't do a careful check, but I noticed that `pegasus` and `anderson_bjorck` are missing.

I recently added those, and was debating on keeping them. I pulled them from a reference.  But let me go ahead and add a documentation.

 
> Once these are addressed, I think this will be ready to go, unless you have other changes in mind.

OK

> (Side note -- I see you updated from `develop`. There may be some test failures after doing this.)

Damn, I was trying to keep the tests updates here to prevent that. I got a merge conflict for the tests.

eloch216:
Seems risky to claim something is the "final update" 😂

I see you commented out a function in `root_onedim.h` called `both_almost_zero`. We generally try to avoid doing this -- it's preferable to just delete something if it isn't necessary.

An exception to this rule can be made for pieces of code that might be useful for debugging. But in that case, there should be another comment explaining when someone would want to un-comment the commented-out code.

leavesandlemmas:
> Seems risky to claim something is the "final update" 😂
> 
> I see you commented out a function in `root_onedim.h` called `both_almost_zero`. We generally try to avoid doing this -- it's preferable to just delete something if it isn't necessary.
> 
> An exception to this rule can be made for pieces of code that might be useful for debugging. But in that case, there should be another comment explaining when someone would want to un-comment the commented-out code.

I meant to remove that actually. Wanted to make sure nothing broke first, but then forgot to remove after checking nothing broke.

eloch216 (approved):

yufenghe22 (approved):
Looks good to me. Just FYI, I got an NaN result in one of my optimization for biomass partitioning. This means, at least one of the biomass came back as NaN, which had not happened before. The biocro version I'm using is based on the current `develop` branch, which has the previous `secant` method. Could this PR prevent a potential NaN result? I'm still trying to find out the cause to the NaN result, so feel free to ignore this question since it may not be due to the `secant`.

leavesandlemmas:
> Looks good to me. Just FYI, I got an NaN result in one of my optimization for biomass partitioning. This means, at least one of the biomass came back as NaN, which had not happened before. The biocro version I'm using is based on the current `develop` branch, which has the previous `secant` method. Could this PR prevent a potential NaN result? I'm still trying to find out the cause to the NaN result, so feel free to ignore this question since it may not be due to the `secant`.

Maybe yes. Any computer implementation of the secant method uses this formula to generate a new guess:

$$
x_{n+1} = x_n - f(x_n) \frac{x_n - x_{n-1}}{f(x_n) - f(x_{n-1})}
$$

where the formula comes from using the secant's slope:

$$
\frac{\Delta f_n}{\Delta x_n }= \frac{f(x_n) - f(x_{n-1})}{x_n - x_{n-1}}
$$

to find a new zero:

$$
0 = f(x_{n+1}) \approx f(x_n) + \frac{\Delta f_n}{\Delta x_n } (x - x_n)
$$

However, the differences can yield non-finite floats like Inf or NaN, whenever the difference in x or f(x) becomes zero. In fact, even when the secant method converges correctly, it will eventually generate NaN when the difference $\Delta x_n$ becomes zero, as $\Delta f_n$ will be zero on the next iteration. So it's important that the checks for convergence catch this before it happens.  However, I don't know if that's why you're having trouble. 

I did a lot more thinking about tests for convergence for this PR, although I don't think it should really change the behavior of the secant method.  But you could try a different method and see if that prevents the NaN.   

However, it's probably not the secant code per se. I think many of the models we use become ill-defined or break outside a certain range of parameter values. Like the Ball-Berry model behaves very strangely when RH < 0.15 or atmospheric CO2 < 200 ppm, and low canopy water conductance (gbw < 0.2).  In practice, those values don't normally happen but they aren't impossible; and sooner or later unusual conditions are generated in parameter optimization. So the secant method might be trying to find the zero when there's multiple zeros or singularities or other pathologies that frustrate the method, which could result in NaN potentially.

eloch216:
This has two approvals so it can (and should) be merged.

PR #177, opened 2025-04-10
Add test cases for c3 & c4 photosynthesis modules

eloch216:
This PR adds some more realistic test cases for each of the main C3 and C4 modules. These will be more sensitive to changes in key functions like `c3photoC` and `c4photo`, so we can catch any changes in the behavior of those functions.


leavesandlemmas (approved):
This looks great! After working with the C3 assimilation module for the secant method PR, I was thinking that we probably need more realistic test cases. The input-one-for-all case seem like it could create an issue since the Ball-berry model tends to create strange behavior at RH = 0.

eloch216:
> This looks great! After working with the C3 assimilation module for the secant method PR, I was thinking that we probably need more realistic test cases. The input-one-for-all case seem like it could create an issue since the Ball-berry model tends to create strange behavior at RH = 0.

Yeah, the "one for everything" tests are not the best. But, it's time-consuming to add good test cases!

leavesandlemmas:
How do these constancy checks work?

eloch216:
> How do these constancy checks work?

The goal of `c3photoC` is to find values of `Ci`, `gsw`, and other quantities that simultaneously satisfy the FvCB and Ball-Berry models for a given set of environmental conditions (temperature, boundary layer conductance, etc.) So the idea behind these checks is to make sure the results are really self-consistent.

One half of it is checking the FvCB outputs. Since `c3photoC` and the `BioCro:c3_assimilation` module return values of `Ci`, we can plug those back in to the `FvCB` module and see if we get the same assimilation rates that were returned by `c3photoC`.

The other half is checking the Ball-Berry outputs. Since `c3photoC` gives us assimilation rates, we can plug those back in to the `ball_berry` module and see if we get the same stomatal conductance that was returned by `c3photoC`.

These tests should fail whenever the root finder in `c3photoC` fails to converge. I haven't tried this yet, but I believe that setting an excessively small max iteration number (say, 1) would cause this check to fail. Right now on this branch, part of it fails because the returned values of conductance-limited assimilation are not consistent with the returned values of `gsw` and `gbw`.

leavesandlemmas:
The consistency test sounds good, btw!

eloch216:
> > These tests should fail whenever the root finder in `c3photoC` fails to converge.
> 
> Speaking of, any thoughts on how should we handle such failures in general? Like I'm thinking the root solver code shouldn't throw an exception, in case the client code wants to handle it, but it should signal failure.

I think the best approach is to just return the last guess, which in principle should usually be the "best" guess. That's what we've done in the past.

Throwing an error is pretty annoying, since it would prevent a simulation from running, but there may not be a way to fix the issue. There is sort of a success indicator already -- if the number of iterations has reached the maximum value, then the loop didn't converge. This could be made more explicit with a 0/1 indicator of success -- maybe it would be 1 if iterations < max, 0 otherwise. Then it would be possible to check how often failures occur -- the hope is that they would only happen a few times during a given simulation, so they wouldn't influence the results much.

leavesandlemmas:
> Throwing an error is pretty annoying, since it would prevent a simulation from running,

Yeah, so I figured it would be better for the root solver code to not throw any exceptions at all; rather they always return a result struct with the last best guess, the residual / check (which wont equal zero) and the iteration count. But they could also return a flag (e.g., a scoped enum) that indicates the failure, so whatever code that calls the root solver can see why it terminated, either successfully or in failure, then the caller can decide how to handle the error, if desired.

PR #179, opened 2025-04-16
Cumulative respiration

eloch216:


eloch216:
Here's some code that can be used to test the new features on this branch:

```R
# Load libraries
library(BioCro)
library(lattice)

# Run the soybean model with the cumulative modules
soybean_res <- with(soybean, {run_biocro(
    c(initial_values, list(
        canopy_assimilation = 0,
        canopy_gross_assimilation = 0,
        canopy_non_photorespiratory_CO2_release = 0,
        canopy_photorespiration = 0,
        canopy_transpiration = 0,
        Grain_gr = 0,
        Grain_mr = 0,
        Leaf_gr = 0,
        Leaf_mr = 0,
        Leaf_WS_loss = 0,
        Rhizome_gr = 0,
        Rhizome_mr = 0,
        Root_gr = 0,
        Root_mr = 0,
        Shell_gr = 0,
        Shell_mr = 0,
        soil_evaporation = 0,
        Stem_gr = 0,
        Stem_mr = 0,
        total_precip = 0,
        whole_plant_growth_respiration = 0
    )),
    parameters,
    soybean_weather[['2002']],
    direct_modules,
    c(differential_modules, list(
        'BioCro:cumulative_carbon_dynamics',
        'BioCro:cumulative_water_dynamics'
    ))
)})

dev.new()
xyplot(
    canopy_transpiration + soil_evaporation + total_precip ~ fractional_doy,
    data = soybean_res,
    type = 'l',
    xlab = 'Day of year (2002)',
    ylab = 'Cumulative water flow (Mg / ha)'
)

dev.new()
xyplot(
    canopy_gross_assimilation + canopy_assimilation + canopy_non_photorespiratory_CO2_release + canopy_photorespiration ~ fractional_doy,
    data = soybean_res,
    type = 'l',
    xlab = 'Day of year (2002)',
    ylab = 'Cumulative canopy CO2 exchange flow (Mg / ha)'
)

dev.new()
xyplot(
    Grain_gr + Leaf_gr + Leaf_WS_loss + Rhizome_gr + Root_gr + Shell_gr + Stem_gr + whole_plant_growth_respiration ~ fractional_doy,
    data = soybean_res,
    type = 'l',
    grid = TRUE,
    xlab = 'Day of year (2002)',
    ylab = 'Cumulative growth respiration costs (Mg / ha)'
)

dev.new()
xyplot(
    Grain_mr + Leaf_mr + Rhizome_mr + Root_mr + Shell_mr + Stem_mr ~ fractional_doy,
    data = soybean_res,
    type = 'l',
    xlab = 'Day of year (2002)',
    ylab = 'Cumulative maintenance respiration costs (Mg / ha)'
)
```

Below are two new graphs showing cumulative losses due to growth respiration and maintenance respiration:

image

image

eloch216:
I think it's clear that we need to change the behavior of some modules, especially regarding growth respiration, which effectively becomes negative at night because of a misguided "clamping" operation in `resp()` (which is now `growth_resp()`).

Another weird thing is that leaf water stress reduces leaf growth rates, but the fate of the lost carbon is not really specified. Is the reduction accomplished via increased growth respiration in the leaf? Or should we think of this as a change in partitioning (so more carbon is used for root growth at the expense of leaf growth when the plant is water stressed)?

Finally, the "whole-plant" growth respiration rate is not calculated consistently across the modules -- it is done different in `c3CanAC` and `CanAC` as compared to `multilayer_canopy_integrator`.

We can start to discuss changes here, but I would prefer to save them for the next PR. I think this will all be easier to follow if this PR does not change any behavior.

leavesandlemmas:
Weird that the growth respiration becomes negative... well I guess it's not weird if it's a fraction of $dm/dt$ but I thought the `net_assimilation_rate` was clamped to zero when negative?

eloch216:
> Weird that the growth respiration becomes negative... well I guess it's not weird if it's a fraction of d m / d t but I thought the `net_assimilation_rate` was clamped to zero when negative?

The code for the `resp` function has historically been:

```c++
double resp(double base_rate, double grc, double temp)
{
    double ans = base_rate * (1 - (grc * Q10_temperature_response(temp, 0.0)));

    if (ans < 0) ans = 0;

    return ans;
}
```

When `base_rate` is not negative, we can understand this as:

```
resp_losses = base_rate * grc * Q10_temperature_response(temp, 0.0);
new_rate = base_rate - resp_losses;
return new_rate;
```

But when `base_rate` is negative, `new_rate` is clamped to 0. If we think of this again as `new_rate = base_rate - resp_losses`, the only way to have a `new_rate` of 0 is for `resp_losses` to be equal to `base_rate`. In other words, the respiration rate is negative. Or, another way to think of it is that clamping to 0 effectively pulls a bunch of carbon out of nowhere to cancel out the losses that should have occurred.

This is another great example of why clamping is almost always a bad idea in biological models.

leavesandlemmas:
> But when `base_rate` is negative, `new_rate` is clamped to 0. If we think of this again as `new_rate = base_rate - resp_losses`, the only way to have a `new_rate` of 0 is for `resp_losses` to be equal to `base_rate`. In other words, the respiration rate is negative. Or, another way to think of it is that clamping to 0 effectively pulls a bunch of carbon out of nowhere to cancel out the losses that should have occurred.

But that means the "respiration rate" doesn't represent cellular respiration rate, I think. A negative rate would mean that respiration rate is pulling carbon from outside the plant. But cellular respiration  only emits CO2 as far as I know.

eloch216:
> > But when `base_rate` is negative, `new_rate` is clamped to 0. If we think of this again as `new_rate = base_rate - resp_losses`, the only way to have a `new_rate` of 0 is for `resp_losses` to be equal to `base_rate`. In other words, the respiration rate is negative. Or, another way to think of it is that clamping to 0 effectively pulls a bunch of carbon out of nowhere to cancel out the losses that should have occurred.
> 
> But that means the "respiration rate" doesn't represent cellular respiration rate, I think. A negative rate would mean that respiration rate is pulling carbon from outside the plant. But cellular respiration only emits CO2 as far as I know.

Yes, I agree this is not right. I made a "discussion" for it: https://github.com/biocro/biocro/discussions/180

eloch216:
I also made a discussion for the impact of water stress on leaf growth: https://github.com/biocro/biocro/discussions/181

tests/testthat/test.cumulative_modules.R
```diff
@@ -1,22 +1,118 @@
...
-            direct_modules,
-            c(differential_modules, list(
-                'BioCro:cumulative_carbon_dynamics',
-                'BioCro:cumulative_water_dynamics'
-            ))
-        )})
+# Helping function for checking carbon accounting for a particular growth
+# calculator module using the soybean model as a base. Here we use a nonzero
+# `growth_respiration_fraction` to test that it is properly accounted for.
+test_soybean_carbon_accounting <- function(partitioning_calculator) {
```
leavesandlemmas, line 4: I like the idea of these tests. It's like a simple form of property-based testing. I wonder if we can adapt this sort of thing to other tests.

tests/testthat/test.cumulative_modules.R
```diff
@@ -1,22 +1,118 @@
...
+                    'BioCro:cumulative_water_dynamics'
+                ))
+            )})
+        )
+
+        # Check that all the assimilated carbon (gross assimilation) is balanced
+        # by the sum costs of photorespiration, non-photorespiratory CO2 release
+        # by the leaf, growth respiration, maintenance respiration, tissue
+        # growth, and litter formation.
+        soybean_res$total_carbon_use <- with(soybean_res, {
```
leavesandlemmas, line 55: I am wondering if this could simplified using some modules.   There's a total mass module that computes the total mass,  there could also be a module that computes the total maintenance respiration (or is there already?), then you wouldn't have to implement those sums here.

tests/testthat/test.cumulative_modules.R
```diff
@@ -1,22 +1,118 @@
...
+            Leaf_gr +
+            Leaf_mr +
+            Leaf_WS_loss +
+            Rhizome_gr +
+            Rhizome_mr +
+            Root_gr +
+            Root_mr +
+            Shell_gr +
+            Shell_mr +
+            Stem_gr +
```
leavesandlemmas, line 79: Doesn't this implementation double-count the growth respiration ?  There's both a whole-plant growth respiration and GR terms for each organ.

tests/testthat/test.cumulative_modules.R
```diff
@@ -1,22 +1,118 @@
...
+            (Root - soybean$initial_values$Root) +
+            (RootLitter - soybean$initial_values$RootLitter) +
+            (Shell - soybean$initial_values$Shell) +
+            (Stem - soybean$initial_values$Stem) +
+            (StemLitter - soybean$initial_values$StemLitter) +
+            canopy_non_photorespiratory_CO2_release +
+            canopy_photorespiration +
+            Grain_gr +
+            Grain_mr +
+            Leaf_gr +
```
leavesandlemmas, line 70: Don't the  canopy non_photorespiratory_CO2_release terms already account for leaf growth and maintenance respiration? 

I know soybean biocro  probably fails the test otherwise, but I think we should maybe have this test print a warning rather than raise an exception when building biocro,  but make the test check for accurate accounting.  That way in the future, as we make changes, we don't have to rewrite the test as well as the models.

leavesandlemmas (commented):
Overall the implementation looks good.

src/module_library/no_leaf_resp_neg_assim_partitioning_growth_calculator.h
```diff
@@ -140,54 +163,67 @@ string_vector no_leaf_resp_neg_assim_partitioning_growth_calculator::get_inputs(
...
-    double const net_assimilation_rate_grain{
-        kGrain > 0 ? canopy_assim * kGrain : 0};
-
-    // Calculate the rate of new shell production without any respiratory costs
-    // (Mg / ha / hr)
-    double const net_assimilation_rate_shell{
-        kShell > 0 ? canopy_assim * kShell : 0};
+    // Calculate the base rate of new leaf production, accounting for water
+    // stress but no additional respiratory costs (Mg / ha / hr)
+    double const base_rate_leaf{kLeaf > 0 ? canopy_assim * kLeaf * LeafWS : 0};
```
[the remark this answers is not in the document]
eloch216: The rhizome k coefficient can be negative; see `miscanthus_x_giganteus$parameters$kRhizome1` and `miscanthus_x_giganteus$parameters$kRhizome2`. This means that the rhizome is acting like a carbon source rather than a carbon sink. It happens early in the season, when the plant begins to grow from underground reserves stored up from the previous year. Note that the miscanthus model doesn't use the logistic partitioning coefficients.

In general I don't think this should happen for any other tissues, but if I recall correctly, we added the checks for negative k values for consistency, so each tissue component is treated in a similar way.

tests/testthat/test.cumulative_modules.R
eloch216: > Don't the canopy non_photorespiratory_CO2_release terms already account for leaf growth and maintenance respiration?

Respiration is a very murky concept for me. I think different people have different opinions about this. I'm not very informed and don't have a strong opinion.

The soybean model uses the `BioCro:no_leaf_resp_neg_assim_partitioning_calculator` module. This module does not apply any additional growth respiration costs to the leaf. So you can see that `Leaf_gr_rate` and `Leaf_gr` are always zero in the default soybean model outputs. We do have a separate maintenance respiration cost for leaf, though, so `Leaf_mr_rate` and `Leaf_mr` are generally not zero.

Note that the `BioCro:partitioning_calculator` module _does_ apply additional growth respiration to the leaf. A long time ago, someone realized that this is probably not correct. Hence why the other partitioning calculators were created. But, the original one was kept so older BioCro models could be reproduced.

> I know soybean biocro probably fails the test otherwise, but I think we should maybe have this test print a warning rather than raise an exception when building biocro, but make the test check for accurate accounting. That way in the future, as we make changes, we don't have to rewrite the test as well as the models.

The goal of the test is to check for all possible places carbon can end up in the model. So I'm not sure why we would change it?

src/module_library/CanAC.cpp
```diff
@@ -211,15 +212,15 @@ canopy_photosynthesis_outputs CanAC(
...
-    ans.Rp = canopy_rp;                            // micromol / m^2 / s
-    ans.RL = canopy_RL;                            // micromol / m^2 / s
-    ans.Trans = CanopyT * cf2;                     // Mg / ha / hr
-    ans.canopy_transpiration_penman = CanopyPe;    // mmol / m^2 / s
-    ans.canopy_transpiration_priestly = CanopyPr;  // mmol / m^2 / s
-    ans.canopy_conductance = canopy_conductance;   // mol / m^2 / s
-
-    return ans;
+    return canopy_photosynthesis_outputs{
+        /* .Assim = */ CanopyA * (1.0 - growth_respiration_fraction),  // micromol / m^2 / s
```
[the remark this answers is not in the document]
eloch216: > Does it really make sense to define `assim` to be the net assimilation rate including some but not all sources of respiration ? Or does this only include leaf respiration sources?

I think the reasoning behind the name is that we scale up the leaf-level net CO2 assimilation rate across the canopy, so we call it the canopy-level net CO2 assimilation rate. I see your point about this rate not including some other potential sources of CO2 release in the leaf.

> At the leaf level, it makes a lot of sense to talk about the net assimilation rate since that's what's measured, and the carboxylation rate has to be inferred. But it includes all sources of respiration, so at the whole-plant / canopy scale, it seems to me that the equivalent net assimilation rate should also include all sources of respiration. Or the canopy modules should compute the gross carboxylation rate, and the accounting for respiration can be done in the mass modules.

This sounds like a pretty big rework of several modules. Right now I'm not convinced either way if it should be done. But I think it's definitely outside the scope of this PR.

tests/testthat/test.cumulative_modules.R
eloch216: That's a good suggestion, I will add this

tests/testthat/test.cumulative_modules.R
leavesandlemmas: >Respiration is a very murky concept for me. I think different people have different opinions about this. I'm not very informed and don't have a strong opinion.

> The soybean model uses the BioCro:no_leaf_resp_neg_assim_partitioning_calculator module. This module does not apply any additional growth respiration costs to the leaf. So you can see that Leaf_gr_rate and Leaf_gr are always zero in the default soybean model outputs. We do have a separate maintenance respiration cost for leaf, though, so Leaf_mr_rate and Leaf_mr are generally not zero.

I guess the photosynthetic tissues complicate the picture a bit. But in a cell without chloroplasts, they emit CO2 (and consume O2) via cellular respiration by the citric acid cycle (AKA the krebs cycle). It produces ATP, NADH, and a few other similar molecules used to drive other redox reactions; and emits CO2. As far as I know, this is main source of CO2 emitted from animal and plant cells. There's some kinds of anaerobic respiration but I think those aren't common in plants / animals.

The photosynthetic pathway in the chloroplasts also involves ATP & friends so presumably the photosynthetic tissues can use those directly without exporting sugars. It depends on whether or not they are used in the cytosol of cells with chloroplasts. 

When I studied NSC dynamics, respiration was defined operationally as the consumption of NSC (mobile carbon substrate if you prefer), whereas photosynthesis was the rate of production. Obviously, that's not exactly true, or at least that's not exactly what the gas exchange systems measure. But it was important that photosynthesis and respiration can't be negative rates. Or at least, the rate of NSC production can't be negative and the rate of consumption can't ever be negative. Those assumptions are necessary to guarantee that the model never generates a negative amount of NSC. But that doesn't make any physical sense.  It's possible to have a net negative production rate, but it can only be a temporary state.

I suspect that this idea applies generally, not just to NSCs.

tests/testthat/test.cumulative_modules.R
eloch216: I added a new module (`BioCro:total_growth_and_maintenance_respiration`) that calculates the total amount of carbon used by growth and maintenance respiration. I included the carbon lost due to leaf water stress as part of the growth respiration, since it's proportional to the amount of carbon available to the leaf for growth.

yufenghe22 (approved):
Our `module_library` has grown increasingly modular with functions split into separate files, but the flat structure of the `module_library`—now holding 150+ files—is becoming unwieldy. We should reorganize it into logical subfolders (e.g., by functionality like `photosynthesis/`, `soil/`, `growth/`etc) to improve maintainability and make the codebase more navigable. This refactor would follow incremental steps: first auditing existing files, then defining a clear structure, and finally migrating components while updating documentation—all to ensure we retain modularity without the clutter.

leavesandlemmas:
> Our `module_library` has grown increasingly modular with functions split into separate files, but the flat structure of the `module_library`—now holding 150+ files—is becoming unwieldy. We should reorganize it into logical subfolders (e.g., by functionality like `photosynthesis/`, `soil/`, `growth/`etc) to improve maintainability and make the codebase more navigable. This refactor would follow incremental steps: first auditing existing files, then defining a clear structure, and finally migrating components while updating documentation—all to ensure we retain modularity without the clutter.

I second this idea. I also think maybe clearly separating the files which don't define modules would be good. It was difficult for me to follow the structure of the existing models (and I probably still don't grasp all of it) with the flat structure. It was hard to track down where something is defined. 

Although it's a longer term project, I think we could probably refactor some of the common code used for C3 and C4 photosynthesis to reduce the reduplication.

leavesandlemmas (approved):

PR #182, opened 2025-04-18
Take different approach to formatting

eloch216:
This PR takes a different approach to formatting the error messages when there is an issue with a module test case. These error messages are also optional now, so users can turn them off if they don't want to see them. (Although, who would want to do this?)

You can see how it works by changing an input or output in one of the saved test case files.


leavesandlemmas (approved):
Looks good!

PR #184, opened 2025-04-21
Standardize growth respiration

eloch216:
In this PR, I have made some changes to our growth respiration calculations to ensure that the growth respiration rates are never negative. To do this, I have set them all to zero whenever the base rate of carbon available for growth is negative. This approach is also applied when calculating reductions in leaf growth rate due to water stress; now, these losses will only occur when the leaf is actually growing (i.e., when the net canopy CO2 assimilation rate is positive).

These changes are based on discussions that can be found in:
- https://github.com/biocro/biocro/pull/179
- https://github.com/biocro/biocro/discussions/180
- https://github.com/biocro/biocro/discussions/181

This PR modifies the behavior of several key modules, so the miscanthus, willow, and soybean simulation results have changed relative to the `develop` branch. See the figure below, where `new` is this branch and `stored` is the stored results from `develop`. There is a large reduction in stem mass and a smaller reduction in leaf mass compared to `develop`.

Picture1

This can be understood by looking at the cumulative growth respiration rates (shown in the next figure). The total losses due to leaf water stress are now a little bit higher, and the total losses due to stem growth respiration are now much higher. A similar figure generated on the `develop` branch can be found here: https://github.com/biocro/biocro/pull/179#issuecomment-2810355864.

image

In turn, the increase respiration costs are because we now prevent negative respiration rates from occurring. Effectively, negative respiration rates were adding extra carbon to the crops, and now that extra carbon is gone.


eloch216:
Following these change, we will need to re-parameterize the `soybean` model. @yufenghe22 , can you do this as usual? I would expect that the growth respiration coefficients for leaf and stem will become a little smaller, with a larger effect for stem. Other changes will probably be minor.

yufenghe22:
> Following these change, we will need to re-parameterize the `soybean` model. @yufenghe22 , can you do this as usual? I would expect that the growth respiration coefficients for leaf and stem will become a little smaller, with a larger effect for stem. Other changes will probably be minor.

Yes. Do you want me to commit to this PR or open a new one?

eloch216:
You can just push to this branch
________________________________

yufenghe22:
I re-optimized soybean and the results (_v.s._ observations) can be found [here](https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/142261e8f9eeb4e5aedb59bc58f659c06bbf5fc3/run/optimization/figs/fig_old_water_v101.pdf).

yufenghe22:
> > I re-optimized soybean and the results (_v.s._ observations) can be found [here](https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/142261e8f9eeb4e5aedb59bc58f659c06bbf5fc3/run/optimization/figs/fig_old_water_v101.pdf).
> 
> Thanks! I tried comparing the new simulation output against the stored one from `main`. The biomass values look very similar.
> 
> image

yufenghe22 (approved):

eloch216:
Thanks, Yufeng! I will merge this now. Stay tuned for one more PR related to growth respiration and partitioning. Then I think I will be done with these changes.

PR #185, opened 2025-04-21
A refactor of the 1D root solver library

leavesandlemmas:
This PR refactors a lot of the code for the 1D root solver library.

User facing changes are:

1. Added `root_onedim_test` module that tests all currently available methods. 
2. Added additional documentation.
3. Added a couple of additional methods.


leavesandlemmas:
@eloch216  So I can't get the online tests to pass. There's always a small discrepancy in the iteration count (and sometimes in the assim_conductance). This discrepancy has persisted despite the tests passing locally, and despite me updating the test cases. This branch passes all tests locally. And it passed a CMD build and CMD check. 

I am not really sure what to do, as I am not entirely sure why there's even a difference. I am wondering if it's some difference in the accuracy of floating point arithmetic between my computer and the online checks?  Like 64-bit architecture vs. 32-bit architecture? And that somehow causes one extra iteration?

When you get a chance, maybe try installing this branch on your computer and running the tests?

eloch216:
> When you get a chance, maybe try installing this branch on your computer and running the tests?

I ran them locally and got one failure:

```
Differences found for Module `BioCro:c4_assimilation` test case `typical miscanthus, mid light`:

           Quantity Expected   Actual
1 Assim_conductance 36.90976 36.90976
2        iterations  2.00000  3.00000
```

I also got one warning:

```
Module `BioCro:root_onedim_test` test case `Kepler Equation Test Case`: unexpected outputs were found: illinois_flag, illinois_iteration, illinois_residual, illinois_root
```

----

This warning occurs on all the testing platforms; I guess the test case wasn't updated after the new method was added.

The error appears to be the same one that occurs on ubuntu-24.04 using R version 3.6.0. (There may be other platforms with the same error too, but I haven't carefully checked all the GitHub workflow results.)

The difference in `Assim_conductance` doesn't seem to be real, since all the printed digits are identical. But there is definitely a real difference in the number of iterations.

> I am wondering if it's some difference in the accuracy of floating point arithmetic between my computer and the online checks? Like 64-bit architecture vs. 32-bit architecture? And that somehow causes one extra iteration?

I think OS-specific arithmetic differences are the likely culprit. The tests are able to pass on the `1d-root-solver-library` branch, so I think it should be possible to get them to pass here as well. A potentially important difference is that on this branch, the absolute error tolerance used for the secant method in `c4photo` is `1e-14`, but it is `1e-12` on the `1d-root-solver-library` branch. It's possible that using such a tight tolerance puts the solver into the range where OS differences start to appear.

My suggestion would be to change the tolerance back to `1e-12` (which is still more than sufficient for our purposes!), run the tests locally, use `update_csv` to fix any with errors or warnings, and then push it here.

OS issues are a pain to deal with since the only way to get feedback is to use the GitHub workflow itself. Let's see if the tests can pass with a looser tolerance. If there are still issues, then it's time to keep thinking.

leavesandlemmas:
> My suggestion would be to change the tolerance back to `1e-12` (which is still more than sufficient for our purposes!), run the

That fixed the issue. I think I played around with the tolerances and hadn't realized they were different.  I will probably merge this into the other PR sometime on Wednesday.

eloch216:
> > My suggestion would be to change the tolerance back to `1e-12` (which is still more than sufficient for our purposes!), run the
> 
> That fixed the issue. I think I played around with the tolerances and hadn't realized they were different. I will probably merge this into the other PR sometime on Wednesday.

Sounds good! Glad it was a relatively simple fix. BTW, small differences between OSes are pretty common. That's why the default value of the `relative_tolerance` argument of the `model_test_case` function is so high (1e-3). If it is set any lower, we start to see spurious test failures.

leavesandlemmas:
> > > My suggestion would be to change the tolerance back to `1e-12` (which is still more than sufficient for our purposes!), run the
> > 
> > 
> > That fixed the issue. I think I played around with the tolerances and hadn't realized they were different. I will probably merge this into the other PR sometime on Wednesday.
> 
> Sounds good! Glad it was a relatively simple fix. BTW, small differences between OSes are pretty common. That's why the default value of the `relative_tolerance` argument of the `model_test_case` function is so high (1e-3). If it is set any lower, we start to see spurious test failures.

It feels like there ought to be guidance on choosing a reasonable tolerance so that the calculation is reproducible. But I haven't been able to find anything that really answers the question.  Like people talk about Machine Epsilon, but for 64 bit machines and the IEEE double precision floats, the machine precision is much smaller than 1e-14, but errors in the secant method are apparent there.  

So basically a tolerance that's too large means that we don't really get an accurate zero, but on the other hand, a tolerance that's too low and we get OS differences. Or even failures to terminate for very low tolerances. There's clearly a sweet spot, but I am not sure what the sweet spot should be in theory (even neglecting the trouble of determining in practically). It also would depend on whether 32-bit , 64-bit or 128-bit floats are used. I've seen people suggest the sqrt of the machine epsilon (which is about 1e-8 on my machine), but that actually caused the consistency checks to fail for the `c3photo` code. And I have never seen a theoretical justification for that either.

PR #186, opened 2025-04-23
Consolidate partitioning calculators

eloch216:
This PR makes a few changes to modules related to the partitioning growth model. It adds a few new inputs and outputs, but does not change any key behavior. Note that the stored results for the soybean model did not need to be changed.

----

The main change in this PR is that it consolidates our "partitioning growth calculator" modules so there are just two, rather than three.

We have always wanted to give users an option to apply growth respiratory costs to leaf tissue, or to not apply them. Previously, this required two separate modules -- one of them used the `resp()` function (now the `growth_resp_Q10()` function) for leaf tissue, and one did not. This was necessary because the "clamping" behavior of the `resp()` function meant that setting a growth respiration coefficient to zero was not the same as bypassing the `resp()` function.

Following PR https://github.com/biocro/biocro/pull/184, the "clamping" has been removed, and now leaf growth respiration can be disabled through the parameters by simply setting `grc_leaf` to 0. So there is no longer a need for separate modules. Likewise, a single module can be used to reduce the leaf growth rate in response to water stress, or to ignore leaf water stress. In the latter case, `LeafWS` can simply be set to 1 as a parameter, instead of being calculated from the soil water content via the `BioCro:leaf_water_stress_exponential` or `BioCro:leaf_water_stress_linear` modules.

There is still one aspect of the partitioning growth model that cannot be controlled with a parameter value. This concerns the choice of how to respond to negative canopy assimilation rates. One approach is to take these respiratory costs entirely from the leaf. The other is to distribute them across all tissues. So we do need two modules to cover these possibilities.

So, in this PR, there are now two partitioning growth calculators:

- The `BioCro:partitioning_growth_calculator` module is mostly the same as it has been. When the canopy assimilation rate is negative, respiratory costs are spread across all tissues. When `grc_leaf` is zero, this module is identical to the `BioCro:no_leaf_resp_neg_assim_partitioning_growth_calculator` module, so that module has been removed.

- The `BioCro:no_leaf_resp_partitioning_growth_calculator` module has been replaced by the new `BioCro:partitioning_growth_calculator_leaf_costs` module. (I am open to suggestions for a better name!) When the canopy assimilation rate is negative, the respiratory costs are only taken from the leaf.

Besides consolidating and renaming, this PR also adds some new inputs for the partitioning calculators (`grc_grain` and `grc_shell`). This ensures that each tissue is treated the same way. Now the user can specify through the parameters which tissues do not have growth respiration, by setting the appropriate `grc_` parameter value to zero. Previously the grain and shell coefficients were effectively hard-coded to zero.

----

This PR also expands the `BioCro:partitioning_coefficient_logistic` module so it can now calculate rhizome partitioning coefficients using the logistic model. I found it strange that one tissue type was left out. It's possible to choose input parameter values such that `kRhizome` is always zero, so this doesn't change the outputs from the soybean model.


eloch216:
BTW, this is the last partitioning / respiration thing I am planning to do.

leavesandlemmas (approved):
Seems like a good idea to me. I don't have any thoughts on the implementation.

yufenghe22 (approved):
Looks good to me. A major puzzle to me is the use of `LeafWS`. Like the `StomataWS`, this quantity does not yet have a solid biological basis. Overall, how water stress impacts `gs` and leaf growth could be a major task for future. I think we either find more evidence to support the current simplifications or come up with a more mechanistic way to model them.

eloch216:
> Looks good to me. A major puzzle to me is the use of `LeafWS`. Like the `StomataWS`, this quantity does not yet have a solid biological basis. Overall, how water stress impacts `gs` and leaf growth could be a major task for future. I think we either find more evidence to support the current simplifications or come up with a more mechanistic way to model them.

Agreed!

leavesandlemmas:
> Overall, how water stress impacts `gs` and leaf growth could be a major task for future. I think we either find more evidence to support the current simplifications or come up with a more mechanistic way to model them.

Should we make an issue or discussion about this topic?

eloch216:
> > Overall, how water stress impacts `gs` and leaf growth could be a major task for future. I think we either find more evidence to support the current simplifications or come up with a more mechanistic way to model them.
> 
> Should we make an issue or discussion about this topic?

A discussion would probably be more appropriate. That being said, it would also be a good idea to check through the current issues on `biocro/biocro` and `ebimodeling/biocro-dev` to see if there already is one. It's also not a high priority, unless someone is planning to start working on it soon (I am not.)

yufenghe22:
> Should we make an issue or discussion about this topic?

Open a discussion could be good. I'm currently working on a new soil water module and the `gs` parameters. I have no plan to work on the `leafWS` at the moment.

eloch216:
> > Should we make an issue or discussion about this topic?
> 
> Open a discussion could be good. I'm currently working on a new soil water module and the `gs` parameters. I have no plan to work on the `leafWS` at the moment.

We do already have a discussion for `LeafWS`: https://github.com/biocro/biocro/discussions/181

If you want to discuss `StomataWS`, I will leave it up to you to start a discussion for it 😁

Issue #188, opened 2025-04-24
Should we use a code formatter for R code?


eloch216:
This was inspired by Scott's PR, where VScode automatically re-styled one of our R files: https://github.com/biocro/biocro/pull/176#issuecomment-2814113273

Currently we use `clang-format` to ensure consistent style in our C++ code, but we don't have a similar tool for the R code.

There are at least two R packages that can restyle R code: `formatR` and `styler`. I tried them both out. The `formatR` package made a bunch of weird changes that I didn't like. The `styler` package seems better for our purposes.

To use it, install the package using `install.packages('styler')` and then call the following from an R session running in the BioCro root directory:

```R
styler::style_pkg(
  indent_by = 4L,
  start_comments_with_one_space = TRUE
)
```

The default setting is two spaces for indentation, but our `.editorconfig` file is currently set up to use four, so I set that here too.

There are a few things it does that I don't really like, but I think I could accept them for greater consistency overall. One is that it messes up indentation within comments. For example,

```R
    # Check that the following type conditions are met:
    # - `module_names` should be a vector or list of strings with elements of
    #    length 1
```

becomes

```R
    # Check that the following type conditions are met:
    # - `module_names` should be a vector or list of strings with elements of
    # length 1
```

Another is that it messes up comment alignment across lines. For example,

```R
        module_name,         # %1
        module_library,      # %2
        module_type,         # %3
        module_name_caps,    # %4
```

becomes

```R
        module_name, # %1
        module_library, # %2
        module_type, # %3
        module_name_caps, # %4
```

Likewise with aligned assignment operators:

```R
zenith_angle  <- 40   # degrees
LAI           <- 3    # dimensionless
total_ppfd    <- 2000 # micromol / m^2 / s
beam_fraction <- 0.8  # dimensionless
windspeed     <- 2.0  # m / s
```

becomes

```R
zenith_angle <- 40 # degrees
LAI <- 3 # dimensionless
total_ppfd <- 2000 # micromol / m^2 / s
beam_fraction <- 0.8 # dimensionless
windspeed <- 2.0 # m / s
```

It doesn't seem to process R files in the `data` directory, so it didn't alter our comment alignment in `soybean.R`, which is where it could potentially be the most annoying.

There seem to be a lot of options, so it's possible we could change this behavior, although I'm not sure how.

There is a GitHub workflow that can be used to automatically apply `styler`, although I am not sure we would want to use it: https://github.com/r-lib/actions/tree/v2-branch/examples#style-package

[...]

PR #190, opened 2025-04-28
Use styler

eloch216:
Just making this PR so people can see the changes made by the `styler` package as a follow-up to Issue #188.

It also adds a simple script for using `styler::style_pkg`.


eloch216:
Hm, I found another thing I'm not crazy about. The formatting changes make it less obvious where a long function declaration begins. For example:

```R
run_biocro <- function(
    initial_values = list(),
    parameters = list(),
    drivers,
    direct_module_names = list(),
    differential_module_names = list(),
    ode_solver = BioCro::default_ode_solvers$homemade_euler,
    verbose = FALSE
)
{
    # Make sure weather data is properly handled
    adapted <- adapt_weather_data(drivers, direct_module_names)
    drivers <- adapted$drivers
    direct_module_names <- adapted$direct_module_names
```

becomes

```R
run_biocro <- function(
    initial_values = list(),
    parameters = list(),
    drivers,
    direct_module_names = list(),
    differential_module_names = list(),
    ode_solver = BioCro::default_ode_solvers$homemade_euler,
    verbose = FALSE) {
    # Make sure weather data is properly handled
    adapted <- adapt_weather_data(drivers, direct_module_names)
    drivers <- adapted$drivers
    direct_module_names <- adapted$direct_module_names
```

To me, this makes it hard to see where the code in the function body begins.

justinmcgrath:
The first thing I saw was it putting the opening brace of a function after the signature instead of on a new line. It's a very common convention to have the brace on its own line, and my editor, at least, expects that. Admittedly, this almost immediately gives me a bias against this because it makes me think the authors are not familiar with conventions, but maybe there's a setting for it. It has also misaligned many elements. Skimming this, I don't see that it's done anything I find helpful. It's almost entirely changing ' to ", which is something I do not care about, not even for the sake of standardization because it makes no difference to the program or readability. If there's a way to place the opening brace on its own line and not break alignment, then I think there's no real harm in this, but then I don't really see the benefit either. Maybe this is an already well formatted file, and it would be more impressive with something that we don't like to begin with?

eloch216:
I figured out how to stop changing the quotes and removing line breaks before opening curly brackets:

```R
# Get the basic tidyverse style but with 4 spaces for indenting
transformers <- tidyverse_style(indent_by = 4L)

# We don't want to replace single quotes by double quotes
transformers$token$fix_quotes <- NULL

# We don't want to remove line breaks before opening curly brackets
transformers$line_break$set_line_break_before_curly_opening <- NULL

# Apply customized rules to code
styler::style_pkg(transformers = transformers)
```

But, there's nothing that can be done about the alignment. There is a vignette about it: https://styler.r-lib.org/articles/detect-alignment.html

It looks like `styler` does try to preserve alignment, but only for function calls, and it does not support comment alignment.

I think we will just not use a code styler at this point. Maybe someday `styler` will offer the options we want, but it doesn't have them all yet.

PR #193, opened 2025-05-01
Fix vignettes again

eloch216:
A recent version of `pkgdown` changed the default math rendering engine from mathjax to something else, and it messed up several of our vignettes. I thought I had fixed the issue in PR #187, but it seems I did not. I swear I tested the vignettes and the `pkgdown` site, but now a different problem has popped up, where the equation references are not working correctly.

This PR reverts the math rendering back to mathjax, and it seems to work -- I tried building the vignettes using `tools::buildVignette` and `pkgdown::build_site`, and they seemed to be formatted properly, and the references are working.

### DIFF
--- a/pkgdown/_pkgdown.yml
+++ b/pkgdown/_pkgdown.yml
@@ -1,4 +1,5 @@
 template:
+  math-rendering: mathjax
   bootstrap: 5
   bootswatch: flatly
   theme: arrow-light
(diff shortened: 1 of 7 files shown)


yufenghe22 (approved):
Equation links work fine for me

leavesandlemmas (approved):

Issue #191, opened 2025-04-30
Distinct version number for `develop` branch


eloch216:
Sometimes I have found myself writing code that is intended to work with the latest release of BioCro. In this case it can be helpful to include something that checks the BioCro package version, such as:

```R
expected_version <- '3.2.0'
current_version  <- packageVersion('BioCro')

if (current_version != expected_version) {
    warning(
        'This version of Maize-BioCro was designed for version ',
        expected_version,
        ' of the BioCro R package, but the package tests are being run ',
        'using version ',
        current_version,
        '. This may cause spurious test failures to occur.'
    )
}
```

However, this check doesn't really work because the version of BioCro on the `develop` branch is also `3.2.0`, and the `develop` branch is now quite different from `main`.

So, I propose that when we make a new release of BioCro, we should also update the version number on the `develop` branch to include `-dev`.

For example, after releasing version `3.2.0`, the version on `develop` should become `3.2.0-dev`, indicating that it includes unreleased features that have been made following the release of `3.2.0`. When we make another release, we will choose a 3-part version number following our usual rules. E.g. maybe it will change from `3.2.0-dev` to `3.2.1` or `3.3.0` depending on the types of changes that have been made.

If this sounds like a good idea, I can change the version on `develop` and add some more instructions to our developer guide.

leavesandlemmas:
Sounds like a very good idea to me. 

Out of curiosity, is it possible to have the last version number increment automatically or semi-automatically with PRs committing to the development branch?

justinmcgrath:
Sounds good to me.

eloch216:
> Out of curiosity, is it possible to have the last version number increment automatically or semi-automatically with PRs committing to the development branch?

I'm sure it's possible with some sort of GitHub workflow or action, but it would probably take a while to figure out. It's pretty easy to do this manually. Personally I am content with a manual system at the moment.

eloch216:
This issue was addressed by PR #194

PR #194, opened 2025-05-01
Start using different version numbers on `develop`

eloch216:
This PR addresses Issue #191.

I tried changing the version to `3.2.0-dev`, but R package version designations must be exclusively numeric.

I decided to use 9000 as a starting value following this advice: https://r-pkgs.org/lifecycle.html#sec-lifecycle-version-number-tidyverse

I also decided to use a dash rather than a period to make it look even more different.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: BioCro
-Version: 3.2.0-9000
+Version: 3.2.0-0
 Date: 2025-02-25
 Title: Modular Crop Growth Simulations
 Description: A cross-platform representation of models as sets of equations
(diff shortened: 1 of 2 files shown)


justinmcgrath:
It doesn't really matter what we do, but if "-dev" isn't an option, then "-0" and going up if desired for whatever reason (-1, -2, -3, etc), follows the same logic as semantic versioning, rather than an arbitrary 9000.

eloch216:
> It doesn't really matter what we do, but if "-dev" isn't an option, then "-0" and going up if desired for whatever reason (-1, -2, -3, etc), follows the same logic as semantic versioning, rather than an arbitrary 9000.

That makes sense. I can change to `-0`.

Issue #172, opened 2025-03-26
Secant method plans


eloch216:
Once PR https://github.com/biocro/biocro/pull/166 is complete, there are a few other things that would be nice to do:
1. Use the secant method in `leaf_energy_balance.cpp`.
2. Move the secant root finder to the C++ framework, since it could be useful in many module libraries, not just the main one.

In the long term, it could also be helpful to design an abstract class for root finders, and make a root finder library, as we have done for ODE solvers. Then people could choose from several options. I have an example of this in the `eloch216/edBML` repository, but the version there isn't necessarily the best way to do it.

leavesandlemmas:
Ideally the leaf energy balance and Ci could be solved for together since the solution to each depends on the other. The secant method has a multidimensional version called Broyden's method which would work well. But it's a bit more complex since it requires building an approximate jacobian.

leavesandlemmas:
Also a generic interface is tricky to build because root solving methods have different requirements. Broyden's method works by updating an approximation to the jacobian (or better yet the inverse of the jacobian) (and that might be N + 1 guesses), where's Newton's method requires the jacobian as a funciton/expression that can be evaluated and one starting guess. Bracketing methods (which appear to have higher dimensional analogues) require a starting bracket, but no starting guess.

It's also possible to express a problem in root solving as an optimization problem, and that brings in a zoo of possible methods.

eloch216:
I just remembered there is another fixed-point iteration loop in the `leaf_boundary_layer_conductance_nikolov` function, found in `boundary_layer_conductance.cpp`.

leavesandlemmas:
Do you have a copy of numerical recipes I could borrow? I can investigate the claim. There's a lot of overlap since quasi newton methods for optimization are essentially root solvers looking for zeros of the gradient in the objective function. In fact,  one can even newtons method for optimization, if one can evaluate first and second derivatives. That said,  many of quasi newton methods use the fact that the hessian is symmetric in their approximation. And that's not true for generic systems of nonlinear equations.

eloch216:
No, I don't have a copy. I've just been reading the outdated 2nd version in C, available here: https://numerical.recipes/. I find the "nags" in the 3rd edition to be too annoying. Generally I don't use it for the code itself, but rather its description of the ideas behind the code, which I find to be very nicely done and useful. Most of these are timeless so I'm not too worried about seeing an old version. For example, in the context of root finding, I have found Figure 9.6.1 and surrounding discussion to be super helpful.

[...]

PR #197, opened 2025-05-27
Code for solving multidimensional nonlinear equations: Broyden's good method

leavesandlemmas:
This PR implements an algorithm for finding a zero of a multidimensional function (which is a solution to a nonlinear equation). This is a bit of a work-in-progress, but it's ready to be reviewed.

### The Math / Theory

Given $f:\mathbb{R}^n \to \mathbb{R}^n$, we want to find $x$ such that $f(x) = 0$. This PR implements [Broyden's good method](https://en.wikipedia.org/wiki/Broyden%27s_method)  (see also [numerical recipes](https://numerical.recipes/book.html) 9.7.3 in the 3rd ed.). Broyden's method is essentially the secant method but in more than one dimension.  

Let $df_x$ be the Jacobian (or derivative / differential) of $f$ at a point $x$. Then the secant approximation

$$
\Delta y = df_x \Delta x  + O( |\Delta x|^2)
$$

is used to approximate $df_x$. In 1D, this equation is sufficient to determine an approximation for $df_x$. In higher dimensions, the secant equation is underdetermined, so instead one uses the secant approximation to refine a guess for the Jacobian. Given $df_x \approx  J$, find the smallest change in $\Delta J$ so that $J' = J + \Delta J$ satisfies $\Delta y = J ' \Delta x $. 

1. Use $J$ to refine $x$: $x' = x - J^{-1} y $ where $y = f(x)$.
2.  Evaluate $f$ at the new point: $f(x') = y' $
3. Check if $x$ is a solution: $|y'| \leq \epsilon$ for tolerance $\epsilon$.
4. Using  $\Delta y = y' - y$ and $\Delta x = x' - x$, refine $J$ to produce a new estimate of the Jacobian $J'$

A useful trick is to refine $J^{-1}$ as a guess of $df_x^{-1}$ rather than $J$. Doing so eliminates the need to solve a linear system in step 1. The [Sherman Morrison formula](https://en.wikipedia.org/wiki/Sherman%E2%80%93Morrison_formula) can be used to generate a formula for updating $J^{-1}$.  The formula is 

$$
J^{-1} \mapsto J^{-1}  + \frac{\left(\Delta x - J^{-1} \Delta y\right) \Delta x^* J^{-1} }{\Delta x ^* J^{-1} \Delta y}
$$

Here $x^*$ denotes the dual vector/matrix (i.e., the conjugate transpose) of $x$. This is a rank-1 update since $\Delta J^{-1}$ is a rank-1 matrix.

For convergence, the eigenvalues of the approximation $J$ only needs to have the same signs as the $df_x$ (I think...). Convergence is superlinear in the neighborhood of a zero (and quadratic for Newton's method). While Broyden's method is a little slower than Newton's method in terms of the number of iterations required for convergence, However, Broyden's method can be faster in terms of computation, because it only requires one function evaluation per iteration and this version circumvents solving a linear system  / inverting a matrix. For a system of linear equations, Broyden's method produces a solution in $2n$ iterations ($n$ is the dimension); nonlinear equations take longer, roughly double I believe.

### Implementation Details

The formulas are implemented using `std::array` as the container for vectors and matrices. The user passes a function to a solver class similar to the `root_algorithm` library in PR #176. Currently this function is expected to have the signature:

```cpp
template<typename T, size_t dim>
void f(std::array<T, dim>& y, const std::array<T, dim>& x){
 y[0] = /* implementation */ ;
 y[1] = /* etc */; 
} 
```

All of the linear algebra operations are implemented as for-loops. For readability, I separated steps into methods.  I've only tested this code on 2D systems, so it might not scale well to large problems.  Currently, we have use cases which are fairly low dimensional (like <= 5). And the first major use-case is solving leaf heat and CO2 balance simultaneously, and it is only 2D.

In principle, this code should work for large systems (say dimensions > 100) but it might be slow / memory inefficient.   

### Future plans 

The main additions to this code left to do are:

1. Documentation in the code. I haven't done this part yet because the overall design might change significantly.
2. Double checking the for-loops for optimizations. 

However, it would be nice to have the interface for multidimensional solving mirror the 1D solving library. And it would be nice to easily implement new methods. Those are future goals discussed in #172.  So it might be worth translating this code into a form that's compatible with uBLAS (which we already include in BioCro) so that the linear algebra operations can be expressed using vector notation (without losing much efficiency).  I decided not to do that at first because the documentation of uBLAS is a bit sparse, so it will take some trial and error. 

It would also be nice for the solver to infer the problem's dimension from the function. But I am not yet sure how best to do that.

Some other possible modifications specific to this algorithm are:

1. A limited memory implementation. This can be more efficient in higher dimensional problems where looping over the inverse jacobian becomes slow. 
2. Line searching techniques to avoid too large or small step size. This might require solving a linear system.

Let me know what you think.

### DIFF
--- a/src/module_library/broyden_method.h
+++ b/src/module_library/broyden_method.h
@@ -45,7 +45,7 @@ struct broyden {
     }
 
    private:
-    void set_identity(mat_t& A)
+    void inline set_identity(mat_t& A)
     {
         for (size_t i = 0; i < dim; ++i) {
             for (size_t j = 0; j < dim; ++j) {
@@ -56,7 +56,7 @@ struct broyden {
         }
     }
 
-    void update_delta_x(
+    void inline update_delta_x(
         vec_t& delta_x, const vec_t& y, const mat_t& inv_jac)
     {
         // delta_x = - inv_jac * y
@@ -68,7 +68,7 @@ struct broyden {
         }
     }
 
-    void update_x(
+    void inline update_x(
         vec_t& x, const vec_t& delta_x)
     {
         // x += delta_x
@@ -77,7 +77,7 @@ struct broyden {
         }
     }
 
-    void update_delta_y(vec_t& delta_y, vec_t& y)
+    void inline update_delta_y(vec_t& delta_y, vec_t& y)
     {
         for (size_t i = 0; i < dim; ++i) {
             std::swap(delta_y[i], y[i]);
@@ -85,7 +85,7 @@ struct broyden {
         }
     }
 
-    void update_inv_jac(mat_t& inv_jac, const vec_t& delta_y, const vec_t& delta_x)
+    void inline update_inv_jac(mat_t& inv_jac, const vec_t& delta_y, const vec_t& delta_x)
     {
         // out = inv_jac + (delta_x - inv_jac * delta_y) / ( t(delta_x) * inv_jac * delta_y ) * t(delta_x) * inv_jac
         vec_t a;
@@ -108,7 +108,7 @@ struct broyden {
         }
     }
 
-    bool is_zero(const vec_t& y)
+    bool inline is_zero(const vec_t& y)
     {
         T sq = 0;
         for (size_t i = 0; i < dim; ++i) {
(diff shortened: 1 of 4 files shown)


leavesandlemmas:
I plan to merge next week (~ wed) unless someone asks for more time to review.

PR #205, opened 2025-07-15
Fix canopy photosynthesis bug ebl

eloch216:
The major change in this PR is that I have reinstated the range errors in the FvCB and Ball-Berry model functions, instead of allowing negative `Ci` and `Cs`.

Of course, after doing this, there were lots of "negative Cs" errors. I addressed them by clamping the value of `Assim` passed to the Ball-Berry function in `c3photoC` and `c4photoC`. Nothing needed to be done about negative `Ci`, since the Dekker solver never passes such values. But, if another solver were used and this became a problem, `std::max(0.0, Ci)` could be passed to `FvCB_assim` instead of just `Ci`.

No tests needed to be changed, so this produces mostly the same behavior as on the `fix-canopy-photosynthesis-bug` branch. The main difference is that here the solver is responsible for ensuring that bad inputs are not passed to the Ball-Berry function, rather than the Ball-Berry function silently treating `Cs < 0` as if it were `Cs = 0`.

I think this approach is safer, since there may be contexts where treating `Cs < 0` as if it were `Cs = 0` in the Ball-Berry model could be problematic.

### DIFF
--- a/src/module_library/FvCB_assim.cpp
+++ b/src/module_library/FvCB_assim.cpp
@@ -1,6 +1,9 @@
-#include <algorithm>  // for std::min, std::max
-#include <limits>     // for std::numeric_limits
+#include <algorithm>                 // for std::min, std::max
+#include <limits>                    // for std::numeric_limits
+#include "../framework/constants.h"  // for eps_zero
 #include "FvCB_assim.h"
+
+using calculation_constants::eps_zero;
 
 double inf = std::numeric_limits<double>::infinity();
 
@@ -132,12 +135,14 @@ FvCB_outputs FvCB_assim(
     FvCB_outputs result;
 
     // Calculate rates
-    if (Ci <= 0.0) {
+    if (Ci < 0.0) {
+        throw std::range_error("Thrown in FvCB_assim: Ci is negative.");
+    } else if (Ci <= eps_zero) {
         // RuBP-saturated net assimilation rate when Ci is 0
         double Ac0 =
             -Gstar * Vcmax / (Kc * (1 + Oi / Ko)) - RL;  // micromol / m^2 / s
 
-        // RuBP-regeneration-limited net assimilation when C is 0
+        // RuBP-regeneration-limited net assimilation when Ci is 0
         double Aj0 =
             -J / (2.0 * electrons_per_oxygenation) - RL;  // micromol / m^2 / s
 
--- a/src/module_library/FvCB_assim.cpp
+++ b/src/module_library/FvCB_assim.cpp
@@ -5,8 +5,6 @@
 
 using calculation_constants::eps_zero;
 
-double inf = std::numeric_limits<double>::infinity();
-
 /**
  *  @brief Computes the net CO2 assimilation rate (and other values) using the
  *         Farquhar-von-Caemmerer-Berry model for C3 photosynthesis.
@@ -131,6 +129,9 @@ FvCB_outputs FvCB_assim(
     double electrons_per_oxygenation     // self-explanatory units
 )
 {
+    // Define infinity
+    double const inf = std::numeric_limits<double>::infinity();
+    
     // Initialize
     FvCB_outputs result;
 
--- a/src/module_library/FvCB_assim.cpp
+++ b/src/module_library/FvCB_assim.cpp
@@ -131,12 +131,12 @@ FvCB_outputs FvCB_assim(
 {
     // Define infinity
     double const inf = std::numeric_limits<double>::infinity();
-    
+
     // Initialize
     FvCB_outputs result;
 
     // Calculate rates
-    if (Ci < 0.0) {
+    if (Ci < -eps_zero) {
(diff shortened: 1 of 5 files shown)


eloch216:
Note that the use of the `conductance_limited_assim` function is nearly the same here as on `develop`, except that stomatal conductance is set to infinity. With this change `Assim` is capped at the value that produces `Cs = 0`, rather than the (smaller) value that produces `Ci = 0` (as is done on `develop`). This seems to be essential for getting the solver to converge, although I am not totally sure why.

src/module_library/ball_berry_gs.cpp
```diff
@@ -116,10 +116,13 @@ stomata_outputs ball_berry_gs(
...
                       (dr_boundary / gbw) * assimilation;  // mol / mol.
 
-    // Stomatal conductance becomes infinite as Cs approaches zero from the
-    // right. In this case, there is no water vapor drawndown across the
-    // stomata, so hs becomes 1.
-    if (Cs <= eps_zero) {
+    // Check for error conditions (Cs = 0 or Cs < 0)
+    if (Cs < -eps_zero) {
+        throw std::range_error("Thrown in ball_berry_gs: Cs is negative.");
+    } else if (Cs <= eps_zero) {
```
yufenghe22, line 122: Same as the Ci, this allows Cs to pass between `-eps_zero` and 0, a small negative value.

src/module_library/FvCB_assim.cpp
```diff
@@ -128,16 +129,21 @@ FvCB_outputs FvCB_assim(
...
+    double const inf = std::numeric_limits<double>::infinity();
+
     // Initialize
     FvCB_outputs result;
 
     // Calculate rates
-    if (Ci <= 0.0) {
+    if (Ci < -eps_zero) {
+        throw std::range_error("Thrown in FvCB_assim: Ci is negative.");
+    } else if (Ci <= eps_zero) {
```
yufenghe22, line 141: This allows Ci to pass between `-eps_zero` and 0, a small negative value. Why not just using `Ci < 0` to indicate negatives?

leavesandlemmas:
>When I run this, Assim_check is -1.865175e-14 for Catm = 1000 but is -5.489159e-01 for Catm = 1600. I thought this would trigger a convergence failure message, but it doesn't.

> But, when I use std::min(Assim, conductance_limited_assim(Ca, gbw, inf)) * 1e-6, the solver finds a solution in both cases, and Assim_check is always below the threshold.

Running the script you provided, I get the same output and convergence correctly identified for both PRs. Do you have a script for the A-CI curve? 

```
# using this branch
  Catm   Assim_check iterations
1 1000  1.687539e-14          8
2 1600 -5.329071e-14          8

# using the target marge branch
  Catm   Assim_check iterations
1 1000  1.687539e-14          8
2 1600 -5.329071e-14          8
```
I will do some further tests to see if I can reproduce it. I want to make sure there's no bug in the root-solving code, especially since that's my responsibility. What you describe would be a serious error. I might modify the `c3_assimilation` module and `c4_assimulation` modules so that they also print out the `flag` as termination code for testing purposes.

src/module_library/FvCB_assim.cpp
eloch216: It's due to round-off errors. Imagine it was written this way:

```c++
if (std::abs(Ci) <= eps_zero) {
    // Ci is zero
} else if (Ci < 0) {
    // Ci is negative
}
```

Here, `-eps_zero <= Ci <= eps_zero` is considered to be zero, and `Ci < -eps_zero` is considered to be negative. The code I wrote accomplishes the same thing. I could reverse the `if` conditionals if that would help.

I encountered some convergence errors when just using `if (Ci <= 0.0)`.

eloch216:
> > When I run this, Assim_check is -1.865175e-14 for Catm = 1000 but is -5.489159e-01 for Catm = 1600. I thought this would trigger a convergence failure message, but it doesn't.
> 
> > But, when I use std::min(Assim, conductance_limited_assim(Ca, gbw, inf)) * 1e-6, the solver finds a solution in both cases, and Assim_check is always below the threshold.
> 
> Running the script you provided, I get the same output and convergence correctly identified for both PRs. Do you have a script for the A-CI curve?

To reproduce the issue, you will need to locally change line 119 on this branch. See my earlier comment above. It changes the `gsw` input argument of `assim_limited_conductance` from `inf` to `Gs`.

leavesandlemmas:
> By construction (and assuming no bugs), there is at least one point at which check_assim_rate switches sign inside the bracket. That would means that there's some sort of singularity where the function changes sign but never gets close to zero (to within tolerances).

Oh I tried to reproduce the singularity, and I wasn't able to. That would point the finger at the impurity of `check_assim_rate`.  I can modify that in some future PR.

leavesandlemmas (approved):
I am satisfied. I think you're good to merge these changes into my PR.

PR #208, opened 2025-07-21
Remove energy balance fixed point

eloch216:
This PR changes the `leaf_energy_balance()` function (implemented in `leaf_energy_balance.cpp`) so that it uses Scott's new secant root finder instead of the fixed point iteration method. I also added a few new helper functions (to reduce duplicated code across functions) and some new test cases for the `leaf_evapotranspiration` module.

It was necessary to update some stored regression tests, but the changes were generally small, especially for the saved crop model outputs. By eye, the new results are indistinguishable from the old ones.

The secant method has definitely improved the evapotranspiration calculations. There are a few approaches to illustrating its new behavior. I will show them in a few comments below.

After this, it should be possible to stop using the Penman-Monteith linearization, which I think would also be an improvement. But I want to do this step-by-step so the PRs are simpler.


eloch216:
One issue was that the fixed-point solver sometimes wanted to choose leaf temperatures that were very different from the ambient temperature, forcing us to add a hard constraint on $\DeltaT = T_{leaf} - T_{air}$, requiring that $|\DeltaT| < 10$. See this comment for more info: https://github.com/biocro/biocro/issues/110#issue-2336142057

With the secant method, we no longer seem to have this issue. See the histogram below, showing that the little clumps near $\pm$10 have disappeared on this branch. The largest difference is ~11 degrees, but this is likely not due to the root finding algorithm.

Picture1

Code I used:

```r
library(BioCro)
library(lattice)

  initial_values,
  parameters,
  soybean_weather[['2002']],
  direct_modules,
  differential_modules,
  ode_solver
)})

leaf_temperature_names <- grep('leaf_temperature', colnames(res), value = TRUE)

delta_T <- unlist(lapply(leaf_temperature_names, function(lt) {
  res[[lt]] - res$temp
}))

histogram(
  delta_T,
  freq = FALSE,
  breaks = seq(-30, 30, by = 0.5),
  ylim = c(0, 100),
  grid = TRUE
)

print(max(delta_T))

print(min(delta_T))
```

eloch216:
A third issue is that the leaf temperature did not vary smoothly across the canopy layers. (This is related to the second issue discussed above.) This can be seen in one of our vignettes, available online here: https://biocro.github.io/BioCro-documentation/latest/pkgdown/articles/web_only/canopy_photosynthesis.html#biocro-examples

With the new changes, the variation of the leaf temperature across the layers is now smooth.

Picture3

Code I used:

```r
library(BioCro)
library(lattice)

# Set inputs
zenith_angle  <- 40   # degrees
LAI           <- 3    # dimensionless
total_ppfd    <- 2000 # micromol / m^2 / s
beam_fraction <- 0.8  # dimensionless
windspeed     <- 2.0  # m / s

# Calculate total PAR energy just above the canopy
total_par <- total_ppfd * soybean$parameters$par_energy_content

# Run the ten-layer canopy properties module
canopy_properties <- evaluate_module(
    'BioCro:ten_layer_canopy_properties',
    within(soybean$parameters, {
        cosine_zenith_angle = cos(zenith_angle * pi / 180)
        lai = LAI
        par_incident_diffuse = total_par * (1 - beam_fraction)
        par_incident_direct = total_par * beam_fraction
        windspeed = windspeed
    })
)

# Make a helping function for extracting subsets of the results
canopy_profile <- function(basename, reslist) {
    vnames <- grep(
        paste0(basename, '_layer_[0-9]'),
        names(reslist),
        value = TRUE
    )
    as.numeric(reslist[vnames])
}

# Set the layer numbers
nlayer <- seq(0, 9)

# Get boundary layer conductance
gbw_canopy <- evaluate_module(
    'BioCro:canopy_gbw_thornley',
    c(soybean$parameters, list(
        canopy_height = LAI / soybean$parameters$heightf,
        windspeed = windspeed
    ))
)$gbw_canopy

# Get absorbed longwave energy
air_temperature <- 25

absorbed_longwave <- evaluate_module(
    'BioCro:stefan_boltzmann_longwave',
    list(temp = air_temperature, emissivity_sky = soybean$parameters$emissivity_sky)
)$absorbed_longwave

# Calculate canopy assimilation rates
canopy_assimilation <- evaluate_module(
    'BioCro:ten_layer_c3_canopy',
    within(c(soybean$parameters, canopy_properties), {
        StomataWS = 1
        rh = 0.6
        temp = air_temperature
        gbw_canopy = gbw_canopy
        absorbed_longwave = absorbed_longwave
    })
)

# Plot canopy leaf temperatures
Tleaf_sunlit <- canopy_profile('sunlit_leaf_temperature', canopy_assimilation)
Tleaf_shaded <- canopy_profile('shaded_leaf_temperature', canopy_assimilation)

dev.new()
print(
  xyplot(
    Tleaf_sunlit + Tleaf_shaded ~ nlayer,
    type = 'b',
    pch = 16,
    xlab = 'Canopy layer',
    ylab = 'Leaf temperature (degrees C)'
  )
)
```

eloch216:
The tests are currently failing because a convergence error is thrown when building the `quantitative_model_comparison` vignette:

```
✖ ! Caught exception in R_run_biocro: leaf_temperature solver reports failed
  convergence with termination flag:
✖  Change in guess is zero. Slow improvement
✖ Backtrace:
✖  x
✖  1. \-base::sapply(alpha_rue_sequence, rue_fvcb_square_difference)
✖  2.  \-base::lapply(X = X, FUN = FUN, ...)
✖  3.  \-global FUN(X[[i]], ...)
✖  4.  +-global final_biomass(rue_2002(alpha_rue))
✖  5.  \-BioCro (local) rue_2002(alpha_rue)
✖  6.  +-base::do.call(run_biocro, temp_arg_list)
✖  7.  \-BioCro (local) `<fn>`(...)
✖  8.  \-base::as.data.frame(...)
✖ Quitting from
  /home/runner/work/biocro/biocro/vignettes/web_only/quantitative_model_comparison.Rnw:383-407
```

Unfortunately this will take a little longer to debug since the error doesn't occur on my (Windows) machine. (The documentation is built on Ubuntu.) Stay tuned.

eloch216:
> The results looks much nicer. Does this have much impact one growing season results?

No, it seems to have very little impact on simulated biomass for one growing season.

> Have you been able to trigger the bug? I can load the PR code and run some tests later today, if that would be helpful.

I've identified a test case that passes on Windows but not on Ubuntu. I just added it; we'll soon see whether it passes on Mac.

I've also identified a test case where there is a convergence error on Windows (exceeding max iterations):

```r
evaluate_module(
  list(
    absorbed_shortwave = 27.62563,
    atmospheric_pressure = 101325,
    canopy_height = 10,
    Gs = 0.052272,
    leafwidth = 0.1,
    min_gbw_canopy = 0.005578,
    rh = 0.3675,
    temp = 30.525,
    windspeed = 0.013253,
    wind_speed_height = 10
  )
)
```

eloch216:
The Nikolov boundary layer conductance model itself is odd. Part of it is currently solved using fixed-point. I have manually investigated this part of the model before and found that it does not have a unique solution; because the model uses the absolute value of a temperature difference, there are often two possible solutions (at positive and negative values of the temperature difference). This could also be part of the issue.

eloch216:
I updated `leaf_boundary_layer_conductance_nikolov` to use a better root finding algorithm, and this has addressed some, but not all, of the previous convergence issues.

I first tried the secant method, but had convergence issues in some "weird" cases where there is almost, but not quite, a root near the true root. Then I tried the Dekker method, which is able to deal with these cases. See the figure below for an example where the secant method failed to converge, but the Dekker method works. The line color shows whether the function is `< 0` or `>= 0`, and the black circle shows the root found by the Dekker method.

image

With the Dekker method, the number of iterations for the case discussed above is much smaller.

However, there are still some cases where the Dekker method is failing. These are difficult to understand, since the function looks pretty much linear near the root. See the figure below for an example where the Dekker method fails, producing a "Bracket width is zero. Could be singularity" error. Unfortunately, this type of error is preventing one of the vignettes from being built on Windows, even though all the tests pass. So although it is rare, it still does sometimes occur. (I am not yet sure about other OSes, will need to wait for the checks to complete.)

image

I will try out some other solvers to see if any of them work better than the secant or dekker methods. Any other suggestions? It would be nice to just use the Nikolov model, which has been in BioCro for a long time, instead of switching to a different one.

Here's the code I used to make these figures:

```r
# Load libraries
library(BioCro)
library(lattice)
library(PhotoGEA)

# Define constants
celsius_to_kelvin <- 273.15             # K or degrees C
ideal_gas_constant <- 8.31446261815324  # J / K / mol

# A copy of the C++ function with the same name
saturation_vapor_pressure <- function(
    air_temperature  # degrees C
)
{
    a <- (18.678 - air_temperature / 234.5) * air_temperature;
    b <- 257.14 + air_temperature;
    611.21 * exp(a / b)  # Pa
}

# A copy of the C++ function with the same name
molar_volume <- function(
  temperature,  # degrees C
  pressure      # Pa
)
{
    TK = temperature + celsius_to_kelvin  # K
    ideal_gas_constant * TK / pressure    # m^3 / mol
}

# Creates a copy of the C++ lambda called "check_leaf_gbv_free"
create_gbv_free_check_func <- function(
  air_temperature,  # degrees C
  delta_t,          # degrees C
  ea,               # Pa
  gsv,              # m / s
  lw,               # m
  windspeed,        # m / s
  p                 # Pa
)
{
  # Temperature calculations
  leaftemp <- air_temperature + delta_t            # degrees C
  Tak      <- air_temperature + celsius_to_kelvin  # K
  Tlk      <- leaftemp + celsius_to_kelvin         # K

  esTl <- saturation_vapor_pressure(leaftemp)  # Pa

  # Set constants
  cf <- 4.322e-3   # for broad leaves
  ce <- 1.6361e-3  # for broad leaves
  ct <- 0.378
  
  # Equation 29
  gbv_forced <- cf * Tak^0.56 * sqrt((Tak + 120) * windspeed / (lw * p))  # m / s
  
  # Create the function
  gbv_check <- function(gbv_free) {
    # Equation 35
    eb <- (gsv * esTl + gbv_free * ea) / (gsv + gbv_free)  # Pa
    
    # Equation 34
    Tvdiff <- (Tlk / (1.0 - ct * eb / p)) - (Tak / (1.0 - ct * ea / p))  # K
    
    # Equation 33
    new_gbv_free <- ce * Tlk^0.56 * sqrt((Tlk + 120) / p) * (abs(Tvdiff) / lw)^0.25  # m / s
    
    # Return the difference between the new and old values
    gbv_free - new_gbv_free  # m / s
  }
  
  # Return the function and the value of gbv_forced
  list(
    func = gbv_check,
    gbv_forced = gbv_forced
  )
}

# Runs the C++ Nikolov function via its module
solve_equations <- function(
  air_temperature,  # degrees C
  delta_t,          # degrees C
  ea,               # Pa
  gsv,              # m / s
  lw,               # m
  windspeed,        # m / s
  p,                # Pa
  ADD_CASES = FALSE
)
{
  mv  <- molar_volume(air_temperature, p)            # m^3 / mol
  svp <- saturation_vapor_pressure(air_temperature)  # Pa
  
  inputs <- list(
    windspeed = 0,                                 # m / s (this ensures gbw_leaf = gbw_free)
    leafwidth = lw,                                # m
    temp = air_temperature,                        # degrees C
    leaf_temperature = air_temperature + delta_t,  # degrees C
    Gs = gsv / mv,                                 # mol / m^2 / s
    rh = ea / svp,                                 # dimensionless
    air_pressure = p                               # mol / m^2 / s
  )
  
  if (ADD_CASES) {
    add_csv_row('BioCro:leaf_gbw_nikolov', 'biocro/tests/module_test_cases', inputs, 'Potentially tricky case (windspeed = 0)')
    
    new_inputs <- inputs
    new_inputs$windspeed <- windspeed
    
    add_csv_row('BioCro:leaf_gbw_nikolov', 'biocro/tests/module_test_cases', new_inputs, 'Potentially tricky case (windspeed > 0)')
  }
  
  evaluate_module('BioCro:leaf_gbw_nikolov', inputs)$gbw_leaf
}

# Investigate a tricky case where the secant method gets confused
case_secant <- list(
  air_temperature = 18.200000, # degrees C
  delta_t = -0.821938,         # degrees C
  ea = 292.602534,             # Pa
  gsv = 0.002032,              # m / s
  lw = 0.040000,               # m
  windspeed = 1.227339,        # m / s
  p = 101325.000000            # Pa
)

check_func_info <- do.call(create_gbv_free_check_func, case_secant)

gbv_free_check_func <- check_func_info$func

gbv_free_seq <- seq(0.002, 0.004, length.out = 1001)

gbv_check_seq <- gbv_free_check_func(gbv_free_seq)

solution <- tryCatch(
  do.call(solve_equations, case_secant),
  error = function(e) {
    print(e)
    NA
  }
)

PhotoGEA::pdf_print(
  xyplot(
    gbv_check_seq ~ gbv_free_seq,
    group = gbv_check_seq < 0,
    type = 'l',
    grid = TRUE,
    xlab = 'gbv_free (m / s)',
    ylab = 'check_leaf_gbv_free (m / s)',
    panel = function(x, y, ...) {
      panel.xyplot(x, y, ...)
      panel.points(solution, gbv_free_check_func(solution), pch = 16, col = 'black')
    }
  )
)

# Investigate a tricky case where the Dekker method gets confused
case_dekker <- list(
  air_temperature = 16.39999999999999857891452847979963,
  delta_t = -0.00000051257671529469917004462332,
  ea = 1865.12043054997297986119519919157028,
  gsv = 0.00019007768245344941384494807934,
  lw = 0.10000000000000000555111512312578,
  windspeed = 0.18352098710326952835103497818636,
  p = 101325.00000000000000000000000000000000
)

check_func_info <- do.call(create_gbv_free_check_func, case_dekker)

gbv_free_check_func <- check_func_info$func

gbv_free_seq <- seq(0, 1e-3, length.out = 1001)

gbv_check_seq <- gbv_free_check_func(gbv_free_seq)

solution <- tryCatch(
  do.call(solve_equations, case_dekker),
  error = function(e) {
    print(e)
    NA
  }
)

PhotoGEA::pdf_print(
  xyplot(
    gbv_check_seq ~ gbv_free_seq,
    group = gbv_check_seq < 0,
    type = 'l',
    grid = TRUE,
    xlab = 'gbv_free (m / s)',
    ylab = 'check_leaf_gbv_free (m / s)',
    panel = function(x, y, ...) {
      panel.xyplot(x, y, ...)
      panel.points(solution, gbv_free_check_func(solution), pch = 16, col = 'black')
    }
  )
)
```

eloch216:
> This uses root_onedim.h, which is in the module_library directory, but it is not a module. 

Ultimately, the root solver code will be moved to the `biocro/framework` repository and included in the `biocro/biocro` repository as a Git sumodule. But this code is still in a stage of fairly rapid development, and it's much easier to make changes with it in the same repository as the modules. Once it is more settled, we can move it to a more appropriate place.

> The number of iterations is quite high on the failing test case. Currently, it shows an expected iteration count of `69` and `261` iterations, which is very high

These are at 6 and 10 iterations now that a better solver is being used inside the Nikolov model.

> Most of the other bracketing methods...

It's not always easy to find a way to specify a bracket that's _always_ valid. I haven't tried finding one for the energy balance solver, but it seems like it would be difficult. That's why I went for the secant method there.

> They also seem to be talking about conifer needles and I am not sure these approximations are applicable to flatter leaves like soybean.

It's been a long time since I've carefully looked at the paper, but I believe they consider both needles and broad leaves. The only difference is in the values of some coefficients. If you look at our code, it indicates when we have chosen a broadleaf value.

leavesandlemmas:
> I think the "bracket zero width" errors are spurious in this case, since there is no singularity. I am investigating what happens if I don't treat them as convergence failures.

@eloch216  Yes, they are spurious. Dekker's method *was* identifying the root correctly, but the residual was just slightly outside the 1e-12 tolerance when the `bracket width is zero` check is triggered. I need to modify that. But you can also add an exception to throwing an exception for this termination state. `Flag::bracket_width_zero`

leavesandlemmas:
> Update: the bisection solver generates lots of "leaf_temperature solver reports failed convergence with termination flag: Change in guess is zero. Slow improvement" errors, causing a lot of test failures. Maybe I will revisit this later.)

Oh right, I forgot that this is the main way the bisection method terminates. Basically, Dekker's method looks like its working so allowing `bracket_width_zero` termination state, tightening the relative tolerance, or weakening the absolute tolerance should make the spurious failures pass. 

I think convergence checks need to be reworked. That part makes the code messier than I'd like. And the relative tolerance isn't really used for many checks, while the absolute tolerance is. I am going to grab a coffee to think it over.

leavesandlemmas (approved):
I think you're good to merge. From my perspective anyways.

eloch216:
> Even though this convergence issue occurs in the energy balance calculations, I suspect it may be related to the leaf boundary layer conductance model. I will look into this more later.

As a simple test, I locally set `gbw_free` to 0.005 m / s (based on the final figure here: https://biocro.github.io/BioCro-documentation/latest/pkgdown/articles/web_only/energy_balance.html) instead of solving the model iteratively, since the free conductance tends to be around that value when the wind speed is low. With this change, I'm not getting any convergence issues with the energy balance calculations. I will look into using the Campbell model as previously suggested by @leavesandlemmas, since it is much simpler.

I was hoping we could get the Nikolov model to work, but it seems to be difficult to reliably solve it, and the added complexity might not provide much benefit to us. In fact, fixing `gbw_free` to 0.005 m / s seems to have had no discernible impact on the soybean biomass and transpiration outputs.

leavesandlemmas:
> As a simple test, I locally set `gbw_free` to 0.005 m / s (based on the final figure here: https://biocro.github.io/BioCro-documentation/latest/pkgdown/articles/web_only/energy_balance.html) instead of solving the model iteratively,

I tried something like that too, but I test out this test tomorrow. I have been wondering if the layers of implicit functions / root solving is causing problems or making the leaf balance function less smooth.  I have also tried using Dekker's method rather than the secant method. 

I also think there might be issues with accepting `bracket_width_zero` as a success state because Dekker's method could identify a singularity or some other jump discontinuity and then report success rather than failure. Especially if that's occurring in the Nikolov model, then the secant method might get thrown way off. But on the other hand, I haven't been able to think of any reliable test or heuristic that distinguishes a singularity / jump discontinuity, especially since continuous functions can have steep derivatives which "look" like discontinuities but aren't. We could try to adapt the root-solver from boost to see  those perform any better, but I am not sure they will.

The other branch changes the way the tolerances are applied; basically creating a tolerance for the `x` value (the root) which is used to test whether the bracket has shrunk to zero or whether iterates aren't changing. and a `y` tolerance for the residual, which is used to test for zero.

eloch216:
> I have been wondering if the layers of implicit functions / root solving is causing problems or making the leaf balance function less smooth.
> 
> I also think there might be issues with accepting `bracket_width_zero` as a success state because Dekker's method could identify a singularity or some other jump discontinuity and then report success rather than failure. Especially if that's occurring in the Nikolov model, then the secant method might get thrown way off.

I made `check_leaf_temp` available as a module so it's possible to look at the function whose root is being found. Taking a look at the problematic case identified in my last comment, it seems that there's no discontinuity, but there is a sharp cusp.

image

Code I used:

```r
library(BioCro)
library(lattice)
library(PhotoGEA)

failing_case <- list(
  absorbed_shortwave = 12.93626795299657317173114279285073,
  atmospheric_pressure = 101433.27333333333081100136041641235352,
  canopy_height = 10,
  Gs = 0.20883488976272232551245622289571,
  rh = 0.76666666666666660745477201999165,
  temp = 27.96296296296295835759337933268398,
  windspeed = 0.01680555095806029325400388074740,
  wind_speed_height = 10
)

print(
  evaluate_module(
    failing_case
  )$error_msg
)

rc <- module_response_curve(
  failing_case,
  data.frame(
    leaf_temperature = seq(27.15, 27.35, length.out = 501)
    #leaf_temperature = seq(27.15, 27.20, length.out = 1001)
  )
)

pdf_print(
  xyplot(
    data = rc,
    type = 'b',
    grid = TRUE
  )
)
```

leavesandlemmas:
Hmm, do you want to adapt one of the formula's from Campbell and Norman?  Using Dekker's method (or the bisection method) should be relatively robust,

eloch216:
With the new module, I have confirmed that there are discontinuities. I would bet these are related the the Nikolov model switching between its potential roots. Here's an example:

image

Code I used:

```r
 library(BioCro)
library(lattice)
library(PhotoGEA)

failing_case <- list(
  absorbed_shortwave = 1.66433632521198360265657356649172,
  atmospheric_pressure = 101325.00000000000000000000000000000000,
  canopy_height = 10,
  Gs = 0.00800000000000000016653345369377,
  min_gbw_canopy = 0.01634955039594711626027567774599,
  rh = 0.53000000000000002664535259100376,
  temp = 30.84000000000000696331881044898182,
  windspeed = 0.00810823552960171847758807928130,
  wind_speed_height = 10
)

print(
  evaluate_module(
    failing_case
  )$error_msg
)

rc <- module_response_curve(
  failing_case,
  data.frame(
    leaf_temperature = seq(30.3, 30.33, length.out = 1001)
  )
)

pdf_print(
  xyplot(
    group = leaf_temp_check > 0,
    data = rc,
    type = 'b',
    grid = TRUE
  )
)
```

> Hmm, do you want to adapt one of the formula's from Campbell and Norman? I am willing to help.
>
> Using Dekker's method (or the bisection method) should be relatively robust against those cusps but they would absolutely trash the secant method or Newton's method. I wonder if that's why I can't get the update to the root-solving library to pass the tests.

I will try the Campbell approach. I can do it myself, but will appreciate feedback when they are ready. Looking at the equations, they won't remove the cusps, but they should remove the discontinuities. I think a combo of the Campbell equations + Dekker method (I think I figured out some bounds that work) will hopefully solve these convergence issues. Once it's sorted out here, we can merge into your branch with the root solver changes.

leavesandlemmas:
> With the new module, I have confirmed that there are discontinuities. I would bet these are related the the Nikolov model switching between its potential roots.

At one point, I created a module for the check assim function in the c3photoC.cpp file. But I deleted after fixing whatever because it was just a carbon copy of the code, but any future changes would require updating the module separately, which means it would easily become out of sync.  I think it would be nice to refactor some of that code so that a module could plot the check_assim. Something for the enhancements list

>  I think a combo of the Campbell equations + Dekker method (I think I figured out some bounds that work) will hopefully solve these convergence issues.

For bounds, I didn't run into any invalid brackets while running the test suite with the leaf energy balance using air temperature plus and minus 50 C (and that's probably overkill) if that helps refine your bounds. 

We do actually already include boosts' 1D root solving library. And I was looking at its code to figure out how they tested for convergence. But it seems only slightly more sophisticated than what I was doing. The bigger issue to me is it's not super clear how their code handles discontinuities in bracketing methods.

eloch216:
I will merge this once the checks have finished. The soybean model needs to be reparameterized since the leaf boundary layer conductance model was changed, and a few vignettes need to be updated, but this can happen via other PRs.

PR #211, opened 2025-07-23
Minor change to root solving

leavesandlemmas:
@eloch216  This code passes the convergence check. I made a PR to make it easy to see the differences. Feel free to reject this and modify #208 directly. 

I modified the error message so that any non successful flag prints out the `residual` and `root`. As you suspected, this is an issue with the convergence check. This makes it easier to judge whether errors are spurious or not. And what to investigate.

THe bracket width's tolerance check needs to change but that's going to take a few more minutes to think through. 

By the way, if you use the 3 point start to Dekker's method, I would use three different points. I suspect this just gives a better start for the secant method.

### DIFF
--- a/src/module_library/boundary_layer_conductance.cpp
+++ b/src/module_library/boundary_layer_conductance.cpp
@@ -3,7 +3,7 @@
 #include "root_onedim.h"               // for root_finder
 #include "water_and_air_properties.h"  // for saturation_vapor_pressure
 #include "boundary_layer_conductance.h"
-#include <iomanip>
+
 /**
  *  @brief Calculates the conductance for water vapor flow from the leaf across
  *  its boundary layer using a model described in Nikolov, Massman, and
--- a/src/module_library/boundary_layer_conductance.cpp
+++ b/src/module_library/boundary_layer_conductance.cpp
@@ -3,7 +3,7 @@
 #include "root_onedim.h"               // for root_finder
 #include "water_and_air_properties.h"  // for saturation_vapor_pressure
 #include "boundary_layer_conductance.h"
-
+#include <iomanip>
 /**
  *  @brief Calculates the conductance for water vapor flow from the leaf across
  *  its boundary layer using a model described in Nikolov, Massman, and
@@ -108,7 +108,7 @@ double leaf_boundary_layer_conductance_nikolov(
 
     root_algorithm::result_t result = solver.solve(
         check_leaf_gbv_free,
-        0,   // first guess
+        0.0001,   // first guess
         0,   // lower bound of initial bracket
         0.5  // upper bound of initial bracket
     );
@@ -116,8 +116,8 @@ double leaf_boundary_layer_conductance_nikolov(
     // Throw exception if not converged
     if (!root_algorithm::is_successful(result.flag)) {
         throw std::runtime_error(
-            "gbv_free solver reports failed convergence with termination flag:\n    " +
-            root_algorithm::flag_message(result.flag));
+           root_algorithm::error_message(result,  "gbv_free solver reports failed convergence with termination flag")
+        );
     }
 
     // Get final value
--- a/src/module_library/root_onedim.h
+++ b/src/module_library/root_onedim.h
@@ -4,6 +4,8 @@
 #include <cmath>
 #include <limits>
 #include <string>
+#include <sstream>
+#include <iomanip>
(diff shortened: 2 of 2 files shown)


leavesandlemmas:
@eloch216  I will probably close this PR in the future. I am working on a larger update to root_onedim to improve some of the issues raised by your efforts to remove the fixed point iteration from the energy balance calculations.

I could use your opinion on convergence checking, if you have one. In PR #202, it became clear that `root_finder` would fail to find a root but wouldn't flag the failure. In PR #202, I fixed this by making the `root_finder` more conservative in deciding successful termination. Basically anything other than `residual = 0` was treated as a failure. However, that created issues with PR #208 where a bracket would correctly shrink to zero on a true root, but `root_finder` would report a failure. The spurious convergence failures come from the fact that the tolerance for a bracket width can end up being laxer than the tolerance for `residual = 0`. So the "residual is zero" check isn't triggered but the "bracket width is zero" is triggered.

I am working on making that check less tolerant.

eloch216:
One thing I meant to say here -- I think there is a reason we have not used the `sstream` and `iomanip` libraries anywhere else in BioCro, although I cannot remember for sure. CRAN might flag them since they look like they could be print statements, which are not allowed? Might be good to see if Justin remembers anything about this.

eloch216:
> I could use your opinion on convergence checking, if you have one.

I don't think I have any strong opinions here -- I think you know a lot more about this than I do.

Something that could be helpful is a little more explanation about the meaning of `rel_tol`. Naively, I assumed that the absolute and relative tolerances were both used to set thresholds on the residual to determine when a root has been found, but it seems that's not the case. I tried changing `rel_tol` to see if I could eliminate the "zero width" errors but couldn't get it to help.

> The spurious convergence failures come from the fact that the tolerance for a bracket width can end up being laxer than the tolerance for `residual = 0`. So the "residual is zero" check isn't triggered but the "bracket width is zero" is triggered.
> 
> I am working on making that check less tolerant.

It would be nice if that issue could be directly solved. But it also seems fine to just have different options for `is_successful` that can be used for different problems / solvers. If you can determine by manual inspection that a particular error flag is erroneous for a particular problem, it seems okay to consider it to be successful.

eloch216:
> > One thing I meant to say here -- I think there is a reason we have not used the `sstream` and `iomanip` libraries anywhere else in BioCro, although I cannot remember for sure. CRAN might flag them since they look like they could be print statements, which are not allowed? Might be good to see if Justin remembers anything about this.
> 
> I can probably work around the `sstream` although, but I don't know how else to set the precision when converting to a string. I suppose this could be dropped if need be

I think precision can be specified when using `sprintf`. E.g. `sprintf("%.12f", x)` makes a string containing the value of a floating point variable `x` to 12 decimal places, IIRC.

leavesandlemmas:
When checking whether floating point numbers `x` and `y` are equal, you can either check their absolute difference `abs(x - y) < =abs_tol` or the relative difference `abs(x - y) < rel_tol * abs(x)` but there's several ways to do relative difference. So in general, its something like `abs(x-y) < rel_tol * norm` where `norm` scales for the magnitude. So these can be combined to `abs(x - y) < abs_tol + rel_tol * norm` 

When testing if `x==0` , only the absolute tolerance matters though because `x` will be small. So relative tolerance would only matter for the bracket width, but even then it couldn't be lower than the absolute tolerance, and I think it needs to be much lower. Like `2 * machine_eps` maybe.  I haven't found a paper that talks about this at length. Brent's paper for his method using `tol = 2 * abs(x) * machine_eps  + abs_tol` to find a root to within `tol`.  He says that the algorithm won't converge for a tighter relative tolerance (though I don't entirely understand the reasons why).

I have been rewriting the root solving library and I have the bracket methods use `2 * machine_eps * abs(x)` to decide if the bracket width is zero. So if the bracket width gets to zero without triggering `residual is zero` then its probably a singularity or jump discontinuity, or something else that should result in an exception.  

I have a test suite that kicks the tires much harder than the unit test in biocro. And this seems to work better at distinguishing singularities from valid roots. I have a test case where the "root" is the jump discontinuity in a step function (goes from -1 to 1) and I have another where `f(x) = 1/(x - 1)` which is a singularity, although some of the methods do correctly identify `inf` or `-inf` as a root (that probably also needs to be flagged correctly).

PR #212, opened 2025-07-24
Simplify RUE model

eloch216:
Here I have simplified the radiation use efficiency (RUE) model.

In the original version, the gross assimilation rate was calculated as being proportional to incident PPFD. Then, stomatal conductance was estimated using the Ball-Berry model, leaf temperature was estimated using energy balance, and a leaf-temperature-dependent respiratory cost was subtracted to get the net CO2 assimilation rate.

In this new version, the _net_ assimilation rate is calculated as being proportional to incident PPFD. Internal leaf state variables (e.g. Ci and temperature) are no longer calculated, nor is the leaf boundary layer conductance or the transpiration rate.

There are two reasons why this simpler version might be better.

One is that the old one is now frequently causing convergence errors when determining the leaf temperature. (See this comment: https://github.com/biocro/biocro/pull/208#issuecomment-3110034570) I would suspect that the fixed point solver was also failing in the past, but simply not reporting its failures, since it was silent when it did not converge. These errors no longer occur in the simplified version because no convergence loop is needed.

The other reason is that the new version is probably closer to how RUE models are actually used. I don't think people typically combine RUE with a mechanistic model for leaf temperature, and a temperature-dependent respiratory cost.

What do you think?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -132,6 +132,9 @@ be directly added to this file to describe the related changes.
   growth and maintenance respiration:
   `BioCro:total_growth_and_maintenance_respiration`.
 
+- Simplified the radiation use efficiency (RUE) model such that net CO2
+  assimilation is directly proportional to the incident quantum flux of photons.
+
 - Altered `test_module` (and hence `test_module_library`) so that new module
   outputs produce a warning but don't cause an error on their own.
 


src/module_library/rue_leaf_photosynthesis.cpp
```diff
@@ -1,189 +1,37 @@
...
-    // of the net assimilation rate in some cases.
-    double const an_conductance =
-        conductance_limited_assim(Ca * 1e6, gbw, 1e3) * 1e-6;  // mol / m^2 / s
-
-    an = std::min(an_conductance, an);  // mol / m^2 / s
-
-    // Determine stomatal conductance using the Ball-Berry model, equating the
-    // leaf and air temperatures (mol / m^2 / s)
-    stomata_outputs const BB_res =
-        ball_berry_gs(an, Ca, RH, bb0, bb1, gbw, Tleaf, Tleaf);
```
leavesandlemmas, line 66: Was it intentional to have the ball berry model get Tleaf twice?

leavesandlemmas (approved):
I am okay with the changes. Part of me wants to figure out why exactly the model wasn't converging.

I would say that RUE-type models tend to eschew anything that would require root solving. It's not uncommon to see the RUE parameter adjusted for environmental conditions (like temperature, water, nitrogen availability in the soil, etc). I sorta wonder if `Ag - RL` creates negative `An` which then causes problems later.  

Of course, it's pretty rare to use RUE for a leaf and then do the canopy integral that we do.

eloch216:
On this branch, the alpha_rue optimization plot in the "Quantitative Comparison" vignette now shows as a nice smooth parabola:

image

Compare this to the screenshot in my earlier comment (https://github.com/biocro/biocro/pull/208#issuecomment-3110034570).

Some other results in that vignette have also changed, which I find to be interesting.

Because the previous RUE model included a temperature-dependent respiratory cost, the temperature response of the two models was actually quite similar. This can be seen in Figure 6 of the vignette as built on the `main` branch (https://biocro.github.io/BioCro-documentation/latest/pkgdown/articles/web_only/quantitative_model_comparison.pdf). Because of this, the largest differences between the FvCB and RUE models was their different response to atmospheric CO2 (Figure 7). As CO2 levels rise, the two models diverge in their yield predictions, with the FvCB version predicting higher yields when CO2 is higher (Figure 10).

In the simplified RUE model, the temperature response is essentially flat. In this case, elevated temperature tends to decrease yield in the FvCB model relative to the RUE model, while increased atmospheric CO2 tends to increase yield. It seems that the temperature effect is more significant, and we actually see that the FvCB version predicts lower yields than the simplified RUE model when atmospheric CO2 increases, although the trend with CO2 is less clear due to the influence of temperature, which does not always increase with CO2 every year. Figure 10 from this branch is shown below.

image

src/module_library/rue_leaf_photosynthesis.cpp
eloch216: I think this is probably a "historical accident." At some point in the past, the Ball-Berry function only had one temperature input, which was historically the leaf temperature. Then I added another one, because it really should have two. At first I set both temperatures to the leaf temperature everywhere it was used, to confirm that the outputs remained the same. Then I changed one to air temperature, but I guess I forgot to do it for the RUE model.

I noticed this too and tried fixing it, in case it had any influence on the convergence issue, but it didn't seem to matter.

eloch216:
> I sorta wonder if `Ag - RL` creates negative `An` which then causes problems later.

I considered this too, but I don't think that should cause a problem, since the FvCB model predicts An < 0 as well.

> This version allows allow modules to pass a different `RUE` right ? So in principle, one could have those effects through the RUE parameter.

Correct. This is also noted in the module documentation.

> Of course, it's pretty rare to use RUE for a leaf and then do the canopy integral that we do. So if anything, it's already more mechanistic than the typical version.

Agreed. I think the layer-dependent canopy is not actually even necessary for this simplified version, since it is now completely linear and could be solved analytically. But I like keeping it this way to make it easier to see the difference compared to the FvCB version. And we don't use the RUE model often enough to really care about its computational efficiency.

PR #213, opened 2025-07-28
Update to the root solving library -- tests not passing.

leavesandlemmas:
@eloch216  I have updated the `root_onedim` library and although its basically ready to do a PR.  but I am having trouble getting the tests to pass. But I am not sure if it's a test issue. Would you take a look at the tests? In particular, 

+ the cumulative carbon
+ model simulation stored results

The stored model results seem fine (but I seem to have done something wrong when updating the model).


eloch216:
> It's also weird that the cumulative carbon modules and the FvCB consistency check was failing but the solvers aren't signaling any errors.

I am seeing lots of error messages from the solvers, both in the GitHub Actions check results and locally. For example, the GitHub check for Windows + release version of R reports:

```
Error: Error: processing vignette 'BioCro.Rmd' failed with diagnostics:
Caught exception in R_run_biocro: leaf_temperature solver reports failed convergence with termination flag:
    :
  No change in the sequence of iterates but `residual` is not zero. Either method's iteration is stuck at a fixed point, or improvement is very small slow. If `residual` is small, try increasing `ytol` or decreasing `xtol`. Otherwise, try a different method. 
    Flag = 2
    Root = 23.234908296601059163
    Residual = -1.7763568394002504647e-14
    Iteration = 4
--- failed re-building 'BioCro.Rmd'
```

Almost all of the error messages are similar to this one, with the leaf temperature solver reporting a small number of iterations (often 4) and a very small residual (<2e-14). Maybe the tolerances are not right?

PR #214, opened 2025-07-29
Create a daylength direct module

leavesandlemmas:
This PR adds a single direct module which computes the `day_length` from the solar declination (deg) `solar_dec` and latitude `lat` (degrees). I adapted the module from Equation (3) this paper: [William C. Forsythe, Edward J. Rykiel, Randal S. Stahl, Hsin-i Wu, Robert M. Schoolfield. "A model comparison for daylength as a function of latitude and day of year." Ecological Modelling, Volume 80, Issue 1, 1995, 87-95.](https://doi.org/10.1016/0304-3800(94)00034-F). Equation (3) is the CBM model and as described in the paper has relatively low error compared to the astronomical almanac.  The parameter `p` in the paper is called here the `sunrise_sunset_elevation` which defines the elevation angle at which the sun rises/sets. 

Here's a comparison of the `soybean_weather` `day_length` which is computed from the `oscillation_clock_calculator`:

day_length_comp

The primary reason for this module is that I can't run the default soybean model for any years other than 2004, 2005, 2006, 2007 without providing `day_length`. This module allows the default model to be run for any year.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: BioCro
-Version: 3.2.0-0
+Version: 3.2.0-2
 Date: 2025-02-25
 Title: Modular Crop Growth Simulations
 Description: A cross-platform representation of models as sets of equations
--- a/NEWS.md
+++ b/NEWS.md
@@ -35,6 +35,8 @@ be directly added to this file to describe the related changes.
 
 ## Minor User-Facing Changes
 
+- Added a module `daylength_calculator` which compute day length from solar position.
+
 - Added custom C++ library for the numerical approximation of the zeros of real
   valued 1D functions. See the header file `roots_onedim.h` for example usage,
   and list of available methods.


eloch216:
> This PR adds a single direct module which computes the `day_length` from the solar declination (deg) `solar_dec` and latitude `lat` (degrees). I adapted the module from Equation (3) this paper: [William C. Forsythe, Edward J. Rykiel, Randal S. Stahl, Hsin-i Wu, Robert M. Schoolfield. "A model comparison for daylength as a function of latitude and day of year." Ecological Modelling, Volume 80, Issue 1, 1995, 87-95.](https://doi.org/10.1016/0304-3800(94)00034-F). Equation (3) is the CBM model and as described in the paper has relatively low error compared to the astronomical almanac. The parameter `p` in the paper is called here the `sunrise_sunset_elevation` which defines the elevation angle at which the sun rises/sets.

This will probably be useful to have. I will check the reference and the code in a bit.

> The primary reason for this module is that I can't run the default soybean model for any years other than 2004, 2005, 2006, 2007 without providing `day_length`. This module allows the default model to be run for any year.

The default soybean model can be run for other years. To do this, it's necessary to first run the soybean clock model to get the photoperiod length. For example:

```r
library(BioCro)

# Helper function for adding day length via the clock model
add_day_length <- function(weather_df) {
  clock_res <- with(soybean_clock, {run_biocro(
    initial_values,
    parameters,
    weather_df,
    direct_modules,
    differential_modules,
    ode_solver
  )})
  
  within(weather_df, {day_length = clock_res$day_length})
}

# Now we can run the soybean model for any year
  initial_values,
  parameters,
  get_growing_season_climate(add_day_length(weather[['2019']])),
  direct_modules,
  differential_modules,
  ode_solver
)})
```

Something similar to this is done in the "Quantitative Comparison" vignette.

The clock is separated from the rest of the model because it only really needs to be run once for each year, and including it in the soybean model makes it run slower. There are only a few years in the `soybean_weather` data set because these were the only ones used in the original Soybean-BioCro publication.

src/module_library/daylength_calculator.h
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * @class daylength_calculator
+ *
+ * @brief Computes day_length using a model described by Forsythe et al. (1995).
+ *
+ * Using equation (3) from the paper, the day length \f$D\f$ in hours is
+ *
+ * \f[ D = 24 - \frac{24}{\pi} \cos^{-1} \left( \frac{\sin \rho + \sin \lambda \sin \phi}{ \cos \lambda \cos \phi} \right) \f]
+ *
+ * where \f$\phi = \delta \pi / 180\f$ is the solar declination in radians (computed elsewhere)
```
eloch216, line 21: Is this value typically computed by the `BioCro:solar_position_michalsky` module? If so, it would be helpful to mention, so people know how (one way) to calculate it.

src/module_library/daylength_calculator.h
```diff
@@ -0,0 +1,117 @@
...
+       };
+}
+
+void daylength_calculator::do_operation() const
+{
+    constexpr double pi = 3.14159265359;
+    constexpr double to_rad = pi/180;
+    double offset = std::sin(p * to_rad);
+    double lam = lat * to_rad;
+    double phi = solar_dec  * to_rad;
```
eloch216, line 105: From the weird spacing in this line, I can tell that you have not run `clang-format`. Please make sure the file follows `clang-format`. There are a few other places where the formatting also deviates from `clang-format`.

tests/module_test_cases/BioCro_daylength_calculator.csv
```diff
@@ -0,0 +1,3 @@
+input,input,input,output,"description"
+lat,solar_dec,sunrise_sunset_elevation,day_length,NA
+40,1,0.8333,12.2569962155787,"automatically-generated test case"
```
eloch216, line 3: I don't think this is the automatically generated test case, which uses 1 for all inputs.

src/module_library/daylength_calculator.h
```diff
@@ -0,0 +1,117 @@
...
+    double num  = offset  + std::sin(lam) * std::sin(phi);
+    double denom = std::cos(lam) * std::cos(phi);
+    double u = std::max(std::min(num/denom, 1.0), -1.0);
+    double D = 24 - 24/pi * std::acos(u);
```
eloch216, line 109: Please add comments showing the units of these calculated variables. It seems that `D` is in hours, `lam` is in radians, `denom` is dimensionless, etc.

src/module_library/daylength_calculator.h
leavesandlemmas: I don't have it on my work computer, and I don't know if I can install it on my work laptop as I don't have admin privileges. Does `clang-format` come bundled with rtools44? Or I need to download LLVM? I think on my desktop at home I installed the clang compiler which came with clang-format

eloch216: We do have some info here, with a link to the official `clang-format` page, which includes some instructions for using it: https://biocro.github.io/BioCro-documentation/latest/bookdown/contributing-to-biocro.html#formatting-tools

There is a slippery slope with this documentation. We cannot possibly cover all the ways `clang-format` can be installed or used, and there is no point in duplicating instructions that can be found elsewhere, especially since it is not our tool. Googling for installation/usage help turns up a variety of StackExchange articles. But if you have something that could be helpful, feel free to add.

yufenghe22 (approved):

PR #216, opened 2025-07-30
Simpler gbw leaf

eloch216:
This PR adds the leaf boundary layer conductance model from Campbell & Norman (1998) and uses it in the energy balance calculations.

There are some small changes to soybean biomass calculations (see below). I think this is a worthwhile change though, since the leaf energy balance calculations are now faster and more reliable.

Picture1


eloch216:
I did some more "stress testing" on this branch, and looks like things are still not quite perfect. I was able to find a case that fails with a "bracket zero width" error. Looking at it in more detail, I can see that there are two solutions, with a sharp cusp (or maybe singularity) between them.

image

Code I used:

```r
library(BioCro)
library(lattice)
library(PhotoGEA)

failing_case <- list(
  absorbed_shortwave = 2.81356819416969017311203060671687,
  atmospheric_pressure = 100902.80000000000291038304567337036133,
  canopy_height = 10,
  Gs = 0.20140721371028921793389088179538,
  rh = 0.89000000000000001332267629550188,
  temp = 27.22222222222222143273029359988868,
  windspeed = 0.00000000000000000000000000000000,
  wind_speed_height = 10
)

print(
  evaluate_module(
    failing_case
  )$error_msg
)

rc <- module_response_curve(
  failing_case,
  data.frame(
    leaf_temperature = seq(27.20, 27.25, length.out = 1001)
  )
)

pdf_print(
  xyplot(
    data = rc,
    type = 'b',
    grid = TRUE
  )
)

leavesandlemmas:
That's quite odd to me. Maybe there's a bug in the leaf energy balance?

PR #217, opened 2025-07-31
Reformulate the energy balance equation to actual solve a heat balance equation. 

leavesandlemmas:
@eloch216  So a source of convergence problems has to do with the Penman Monteith equation itself. I had forgot that BioCro was using the Penman Monteith equation to do this calculation, but it's not surprising that it's creating problems. This PR reformulates the code to just solve the original heat balance equation from which the PM equation is derived.  

Note I haven't updated the tests, so that you can judge how much it changes the stored values. There are some non-negligible changes, but nothing so large as to suggest a huge error. I only got two convergence errors running the test suite, and I believe both are spurious ones, I am going to try this with the udpate root solving library so that I can fine tune that aspect.  The code needs some clean up probably. But I at least I can run clang-format at home. 

Although it doesn't eliminate the cusp in the last example you shared in #216 , the heat balance equation is globally fairly linear (so the secant method or dekker method should rapidly converge).  The cusps definitely come from the heat and vapor conductances, and I am not entirely convinced that the cusp is a real physical phenomenon.  

Here's a plot of the failing case from before 

leafcheck

Code: 
```r
library(BioCro)
library(lattice)

failing_case <- list(
  absorbed_shortwave = 2.81356819416969017311203060671687,
  atmospheric_pressure = 100902.80000000000291038304567337036133,
  canopy_height = 10,
  Gs = 0.20140721371028921793389088179538,
  rh = 0.89000000000000001332267629550188,
  temp = 27.22222222222222143273029359988868,
  windspeed = 0.00000000000000000000000000000000,
  wind_speed_height = 10
)

print(
  evaluate_module(
    failing_case
  )$error_msg
)

rc <- module_response_curve(
  failing_case,
  data.frame(
    leaf_temperature = seq(26, 28, length.out = 1001)
  )
)

png('leafcheck.png')
  xyplot(
    data = rc,
    type = 'b',
    grid = TRUE
  )

dev.off()
``` 

So this function evidently has three roots but the two from the pointy cusp I think are not physically possible. I suspect the free heat conductance shouldn't go to zero when the difference in air and leaf temperature is zero. The forced conductance is also zero. 

Here's why the PM equation is causing problems. Campbell and Norman give a derivation of the PM equation (eg. chp 14 or chp 13).  Let $T_L$ be leaf temperature and let $T_A$ be the air temperature, then at steady state, the following heat fluxes should be equal:

$$
R - B(T_L) - H(T_L - T_A) - L(T_L) = 0
$$

where $R$ is the absorbed radiation, $B$ is the emitted blackbody radiation from the leaf, $H$ is the sensible heat flux (conduction/convection with the air) and $L$ is the latent heat flux (due to evaporation / transpiration). This equation is non linear in $T_L$ due to $B$ and $L$.  Famously the power of blackbody radiation varies with the fourth power of temperature, and transpiration depends on the vapor pressure of the leaf, which depends on leaf temperature (it is assumed that the leaf air filled spaces are always saturated). Moreover, the sensible heat loss $H$ is an odd function, it's only linear if the conductance doesn't depend on leaf temperature. 

However, the Penman Monteith equation is derived by linearization at $T_L=T_A$. So that $B$ and $L$ are effectively calculated using $T_A$ rather the $T_L$ and the heat balance equation becomes linear in $T_L$. For instance, this introduces the $s$ parameter as the derivative the saturation vapor pressure with respect to temperature,  

Linearization is a good approximation technique when nonlinear root solving isn't easy to due. Moreover the PM equation would be a good starting guess for the nonlinear equation, but the linearization is likely to produce poor results whenever the assumption $T_L  \sim T_A$ isn't true. 

I am not 100% sure I did the reformulation correctly, so a second pair of eyes would help.


leavesandlemmas:
I think I can prove that this equation has (or should have) only one root. And if so, then the cusp is either a bug or a model error somewhere.

eloch216:
I would like to hold off on this for now.

For one, it seems there are still convergence issues. So directly solving the energy balance equations didn't address the main problem. I'm sure it will be possible to sort this out, but doing so will take more time. Removing the fixed point solver from the energy balance calculations has already gone on for longer than I wanted it to.

Second, in the original PR, I explicitly mentioned that I didn't want to do this (https://github.com/biocro/biocro/pull/208#issue-3249714705):

> After this, it should be possible to stop using the Penman-Monteith linearization, which I think would also be an improvement. But I want to do this step-by-step so the PRs are simpler.

I am happy with the current state of the `simpler-gbw-leaf` branch. Yes, there are a few "zero width bracket" convergence failures that are being intentionally ignored. Yet, this is still a big improvement over the previous version, where the fixed-point method was failing much more often.

We have to balance out the desire for perfect convergence with the necessity of being able to run the model to completion before getting stopped by convergence errors. On this branch, it's not possible to run a parameter optimization for the cowpea or soybean models because of convergence errors.

My plans for the near future are to (1) merge `simpler-gbw-leaf` into `remove-energy-balance-fixed-point`, (2) update `root-convergence-checks-update` from `remove-energy-balance-fixed-point` and address any conflicts, (3) merge `root-convergence-checks-update` back into `remove-energy-balance-fixed-point`, and (4) merge the finalized `remove-energy-balance-fixed-point` into `develop`. Unless there are big objections. I won't be able to start on it until this afternoon, or perhaps tomorrow, though.

leavesandlemmas:
Closing so that any additional commits won't trigger the github actions.

PR #219, opened 2025-08-15
Reformat soybean model

eloch216:
The soybean model needs to be reparameterized again following some changes to the boundary layer conductance model in PR https://github.com/biocro/biocro/pull/216. To do this, I am planning to write an optimization script using tools from the new `BioCroValidation` package (https://github.com/biocro/BioCroValidation).

As a first step, it will be helpful to rewrite `data/soybean.R` using `BioCroValidation::write_model`. That's what I've done here. More specifically:

1. I moved all the documentation comments from `data/soybean.R` to `man/soybean.Rd` so they do not get overwritten by `BioCroValidation::write_model`. Citing sources in the `man` file also makes it easier to write longer comments, and makes it easier to access this information from R (without needing to look at a source file).

2. I rewrote `data/soybean.R` using `BioCroValidation::write_model`.

3. I also noticed that the direct modules were not in a suitable order for evaluation (according to `validate_dynamical_system_inputs`), so I re-ordered them a bit.

Note that this PR does not change any behavior.


leavesandlemmas (approved):
Seems like a good idea. I don't know that I have much to add.

I thought the order of the direct modules didn't matter (in the sense that the framework code would find a suitable evaluation order)?

PR #220, opened 2025-08-25
Reparam soybean

eloch216:
This PR adds a new script for reparameterizing the soybean model (in `inst/extdata/parameterize_soybean.R`) and uses it. I believe that the new script takes the same approach that @yufenghe22 has used in the past, but it would be good to confirm this. Ideally, we would use this script for all future reparameterizations of this soybean model, so it would be helpful to take a close look at it now, and I would appreciate any feedback on it.

While working on this, I discovered a few mistakes in the vignettes, so I fixed them as well.

The figure below compares the new version (on this branch) to the parameterization on `main` for the year 2002. The largest differences are in `Stem` and `StemLitter`, but they are not very large changes. Besides, these litter values are an underestimate for 2002 (see further below), so an increase in litter is not a problem.

Picture1

The script itself also creates plots comparing the soybean model against the measured biomass values (before and after optimization). This is done for both years used in the optimization (2002 and 2005):

image

image


inst/extdata/parameterize_soybean.R
```diff
@@ -0,0 +1,423 @@
...
+# Load required libraries
+library(BioCro)
+library(BioCroValidation)
+library(DEoptim)
+library(lattice)
+library(parallel)
+library(PhotoGEA)
+
+# Specify some settings
+NCORES  <- 64   # the number of cores to use will depend on your computer
```
yufenghe22, line 22: This is a lot of cores by default, but I'm okay with it!

inst/extdata/parameterize_soybean.R
```diff
@@ -0,0 +1,423 @@
+# To use this script, run it using `source`. Several output files will be
+# created in the current working directory. If the results are satisfactory,
+# copy the resulting `soybean.R` file to the `data` directory of the BioCro
+# repository.
```
yufenghe22, line 4: This parameterization is hard-coded for 2002-2006 datasets. We could note this here or somewhere else in case someone want to implement parameterization for any datasets they have.
It could be generalized, but would need more work to make sure data formats are consistent.

inst/extdata/parameterize_soybean.R
justinmcgrath: parallel has a detectCores function that returns the number of cores. `detectCores() - 1` is a safe default to keep one core free for user interaction.

eloch216: That's a good idea. I even used `detectCores` to figure out how many I had, but didn't think to use it that way.

inst/extdata/parameterize_soybean.R
eloch216: I just added some additional description comments to the script in a new comment

yufenghe22 (approved):

PR #233, opened 2025-10-27
Use a heat balance equation to compute leaf temperature (replacing the PM based equation)

leavesandlemmas:
This PR is a draft to replace the current used to compute leaf temperature with a new one based on a heat balance equation. The replacement works, but I am creating this as a draft so that it can be tested by other people. I suspect the biocro simulation tests will need to be updated and maybe a new parameter optimization because the change affects on photosynthesis.

When you have time, you would try out the code here, especially the module `leaf_heat_balance` and the module `ten_layer_c3_canopy_test` ?

## Overview of PR

This PR adds three modules  `leaf_heat_balance`, `ten_layer_c3_canopy_test`, and `c3_leaf_photosynthesis_test`. The latter two are basically just duplicates of `ten_layer_c3_canopy`, and `c3_leaf_photosynthesis` so any model using those modules will run with the test versions. The test versions use a different method to compute leaf temperature.

Th other module `leaf_heat_balance` (note its called `leaf_heat_balance` in the R module library, but `leaf_heat_balance_module` is the c++ type) does basically the same thing as the `leaf_evapotranspiration` module so it's useful for testing.  Just as the `leaf_evapotranspiration` is simply a module wrapper around the `leaf_energy_balance` function, the `leaf_heat_balance_module` is a wrapper around a class `leaf_heat_balance` defined in `leaf_heat_balance.h` . 

The `leaf_heat_balance` implements  the actual heat equation, the code for solving, and the extraction of the results. Like `leaf_energy_balance`, I didn't put it into a namespace (although maybe it should be inside `standardBML`?)  and I separated the declarations and definitions into a header and source file. 

## Theory 

The approach here is exactly the basic principle used in Campbell & Norman to compute leaf temperature (e.g. eq 14.1 in chp 14). They also describe the derivation of the Penman Monteith equation from the heat balance equation.  Basically the leaf absorbs heat from radiation, the leaf irradiance (absorbed radiation) $R$, loses heat via blackbody radiation $B$, sensible heat flux (conduction / convection with the air) $H$, and latent heat flux (from leaf transpiration) $L$.   These losses depend on leaf temperature so at steady state, we get an equation for leaf temperature $T_l$:

$$
R - B(T_l) - H(T_l) - L(T_l) =0 
$$

The blackbody radation term is just the stefan boltzmann law (with leaf emissivity set to 1).  The sensible heat flux is 

$$
H = c_p g_H (T_l- T_a)
$$

where $c_p$ is the specific heat capacity of the air, $T_l - T_a$ is the difference of leaf temperature and air temperature, and $g_H$ is the conductance for heat transfer between the leaf and the air. 

The latent heat flux is 

$$
L = \lambda E 
$$

where $\lambda$ is the latent heat of vaporization for water (which depends on leaf temperature I believe) and the rate of transpiration $E$. The rate of transpiration is the difference in vapor pressure between the air and the leaf (where the leaf is assumed to have 100% relative humidity). Critically this depends on the total conductance of water vapor $g_v$. 

$$
E = g_v \frac{(W_l - W_a)}{P_a}   
$$

The total conductance of water vapor  is computed exactly the same way as the our code currently does.  However, I couldn't figure out exactly what the old code used for the heat conductance, so I used the formula provided in Campbell & Norman (which is what we use for water vapor too). 

## Reasoning for this change

I have two main reasons for this PR.  The first is that the current model is based on the Penman-Monteith, but it isn't exactly the Penman-Monteith equation.  The Penman-Monteith equation is an *approximate* solution to a heat balance equation and the  derivation involves linearizing the blackbody radiation and leaf transpiration.  And the main reason for the linearization is that you get a closed form expression for leaf temperature (technically the original PM equation is for evapotranspiration, not leaf temperature but the derivation of leaf temperature from it is straightforward).

But we don't actually use the linearization exactly since our code doesn't linearize the blackbody radiation term (and maybe some other changes). If we're using a numerical method to compute roots of an equation, then we should probably solve the equation that the PM equation is an approximate solution for. Moreover, I find the logic of the leaf_energy_balance quite difficult to match up with the terms in the PM equation and its difficult to discern what the scientific model is .  For example, I couldn't tell exactly what `leaf_energy_balance` uses for the heat conductance, even after a lot of back and forth between Campbell & Norman and the code. 

Moreover, a followup goal is to solve for leaf temperature and Ci  simultaneously. PR #197 already adds code to do that, but the leaf_energy_balance code would have to be refactored anyway to make that happen, so this is a step in that direction. And a step in ensuring that a simultaneous equation solving code is solving the correct equations. 

## Note of the design of `leaf_heat_balance`

I wrote `leaf_heat_balance` using a `struct` rather than as a set of functions because a struct/class with methods and member fields allows me to break the logic of defining and solving the leaf heat balance equation into several methods without having to pass a lot of info around which  I think is  easier to understand. But I am open to suggestions on the design.  

Also, I have the goal of allowing `leaf_heat_balance` store references to the `state_map` in the same way that the modules do. That way, `leaf_heat_balance` could be a member field of a module, so that during a call of `run_biocro`, `leaf_heat_balance` is only instantiated once. This would allow `leaf_heat_balance` to remember its most recent root to accelerate finding the next root.  Something which is even more important for solving systems of equations.


yufenghe22:
Since you work on a branch, I suggest to remove all `_test`, and treat all scripts as if they are going to replace the original codes. This way, it's easier to spot the changes you made to the original codes on Github.

leavesandlemmas:
> > (1) get Yufeng + Umar's new soil water modules into the `develop` branch
> 
> @leavesandlemmas Regarding this, here is where the new soil water module, and you can play with it like this:
> 
>     1. Install BioCroWater from [here](https://github.com/Matthews-Research-Group/soil-tipping-bucket).
> 
>     2. Use “/run/minimal_case/run_water.R” to run a simple case.
> 
> 
> Any feedback is welcome for making it a PR.

Thanks @yufenghe22 , Is there some aspect you would like help with? Or just sharing?

PR #240, opened 2025-11-22
Improved error messages for time checks.

leavesandlemmas:
While helping Ximin fix some code, I decided that the error messages for the "time is spaced by timestep" could be better.  This PR tries to  make  the error messages more informative and point the user toward the source of the problem and suggest potential solutions. 

I also added some tests and shortened the check that `testthat.R` uses to decide if the correct error message was raised.

### DIFF
--- a/R/input_checking_functions.R
+++ b/R/input_checking_functions.R
@@ -283,6 +283,13 @@ check_time_is_sequential <- function(
     rtol = sqrt(.Machine$double.eps)
 )
 {
+    formatter <- function(err, bdy, idx){
+        rows <- paste0(idx, collapse=', ')
+        msg <- paste(err, bdy, rows, collapse=' ')
+        msg <- paste0(strwrap(msg, width = 80, exdent=4), collapse = '\n')
+        msg
+    }
+    newline_indent <- "\n    "
     
     # only checked if differential modules are present
     if (length(differential_modules) == 0) {
@@ -291,12 +298,7 @@ check_time_is_sequential <- function(
 
     no_time_variable <- !('time' %in% names(drivers))
     if (no_time_variable) {
-        msg <- paste0(c(
-            "No `time` variable found in the `drivers` dataframe.\nSuggested fixes:", 
-            "+ Add a `time` variable spaced by `timestep`:",
-            "    drivers$time <- seq(1, nrow(drivers), timestep)",
-            "+ Add the direct module `BioCro:format_time` if drivers contain `doy` and `hour` variables."), 
-            collapse = "\n    ")
+        msg <- "No `time` variable found in the `drivers` dataframe."
         return(msg)
     }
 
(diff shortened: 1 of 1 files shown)


leavesandlemmas:
Notes I made changes to the way test.that checks if the error message is correct. Instead of checking if the entire error message matches, I have each error message start with a short "error name" so that test.that can use a regexp to check if the error message begins with the correct name. That way, you can modify the error message without necessarily updating the test (as long as the error name is the same). 

`test.that` apparently also allows for checking for S3 classes which serve as custom error types. Although that's probably something to consider doing in the future (that way, errors are more like a value of a certain type; client code can detect the error by type and handle it), but I didn't want to figure out how to make that work. Low priority change.

yufenghe22 (approved):

R/input_checking_functions.R
```diff
@@ -283,19 +283,64 @@ check_time_is_sequential <- function(
...
+    }
+
+    if (is.unsorted(time, na.rm=TRUE, strictly=TRUE)) {
+        # duplicates ? 
+        duplicates <- which(duplicated(time))
+        if (length(duplicates) > 0) {
+            msg <- paste0(c(
+               sprintf("Duplicates found. The `time` variable in the drivers contains %1s duplicate entries.", length(duplicates)),
+                "To find duplicates, call: `which(duplicated(time))`",
+                "+ Check rows:",
```
eloch216: The meaning of "+ Check rows:" in the message is not immediately clear when reading it. I prefer the message sent for unsorted times, where it says "Unsorted times found at rows:" Here, I would suggest saying "Duplicated times found at rows:"

R/input_checking_functions.R
```diff
@@ -304,27 +349,46 @@ check_time_is_sequential <- function(
...
+        return(msg)
+    }
 
-    if (!is_evenly_spaced(time, timestep, rtol)) {
-        return("The `time` variable is not spaced by `timestep`.")
+    if (!all(is_spaced_by_timestep)) {
+        idx <-which(!is_spaced_by_timestep)
+        msg <- paste0(
+            c("Timestep mismatch.", "The `time` variable in the drivers is linear but not spaced by `timestep`. Drivers may contain gaps. Drivers should look like:",
+            "`drivers$time <- seq(1, nrow(drivers), timestep)`",
```
eloch216: The comment that "Drivers should look like: `drivers$time <- seq(1, nrow(drivers), timestep)`" seems a little odd. People are not usually specifying time this way. I would avoid putting R code in the message since people may think they are supposed to run it. It's probably enough to say that `time` is not spaced by `timestep`.

R/input_checking_functions.R
```diff
@@ -304,27 +349,46 @@ check_time_is_sequential <- function(
...
-        second_diff = diff(x, differences = 2)
-        is_zero = abs(second_diff) < rtol
-    } else {
-        first_diff = diff(x, differences = 1) - by
-        is_zero = abs(first_diff) < rtol
-    }
-
-    return(all(is_zero))
-}
+#is_evenly_spaced <- function(x, by = NULL, rtol = sqrt(.Machine$double.eps)){
```
eloch216: If this function is no longer needed, it should be deleted rather than commented out.

R/input_checking_functions.R
```diff
@@ -283,19 +283,64 @@ check_time_is_sequential <- function(
...
+        }
+        # otherwise
+        time_sorted <- sort(time)
+        not_sorted <-time != time_sorted
+        count <- sum(not_sorted) 
+        idx <- which(not_sorted)        
+        msg <- paste0(
+            c("`time` not sorted.", 
+            sprintf("The `time` variable in the drivers should be sorted in increasing order, but %1s elements are not in order.", count), 
+            "Call `sort(drivers, by = time, ascending=TRUE)` to sort the `drivers` data.frame.",
```
eloch216: I'm not sure this suggested fix should be in here. First, it will not work if the drivers have a different name. Second, simply re-sorting the drivers may just "sweep away" some underlying issue with them and may not be the appropriate solution.

PR #244, opened 2025-11-26
Soil albedo

eloch216:
A lot of the new functions and modules related to soil water are poorly documented and a bit confusing. They have many "magic numbers," don't cite references, etc.

It's possible to fix them by doing some detective work, but it can be time consuming. Here I have looked into a relatively simple function, `soil_albedo`, and tried to make it easier to understand.

First, I was able to find the original fortran code in a GitHub repository (https://github.com/DSSAT/dssat-csm-os). It had a reference to a useful paper about soil albedo: Idso et al. (1975) (https://journals.ametsoc.org/view/journals/apme/14/1/1520-0450_1975_014_0109_tdobsa_2_0_co_2.xml). I also figured out some things by checking edge cases in the code. Here's are some of the things I learned:

1. The function actually calculates the albedo of wet bare soil and then modifies it to account for crop canopy coverage. So, more accurately, it returns a surface albedo rather than just a soil albedo. With this in mind, I renamed the function to `surface_albedo`.
2. In the original C++ code, when `soil_water_content[0]` equals 0.03, the `wet_soil_albedo` evaluates to `bare_soil_albedo`. From this, it seems that the input argument `bare_soil_albedo` actually refers to the dry soil albedo. As shown in the paper, this is also the maximum soil albedo. I renamed this input to `bare_soil_albedo_max` to better reflect its meaning.
3. In the original C++ code, when `soil_water_content[0]` equals `soil_field_capacity[0]`, the `wet_soil_albedo` evaluates to `(1 - 0.45) * bare_soil_albedo`. This allowed me to replace the magic number 0.45 with a new constant called `albedo_fraction`. It is equal to 0.55 and represents the soil albedo at field capacity relative to its albedo when dry.
4. In the original C++ code, `ff` is clamped to a maximum value of 2. Working backwards, this corresponds to a maximum soil water content of `0.03 + 2.0 * (field_capacity - 0.03)`. This allowed me to identify the upper and lower limits of the soil water content.
5. In the original Fortran code, the magic number 0.75 is actually a canopy extinction coefficient that is not hard-coded.
6. In the original C++ code, LAI was explicitly allowed to be negative, but I think this should be considered an error instead. So I added some input checks.

Based on these observations and a few others, I wrote some documentation for the function and tried to make the code easier to follow. I also removed an unused input argument (`temp`) and changed two others from arrays to single values (since the entire array is not needed for these calculations). Following these changes, all the tests pass, so I don't think I changed the function's behavior.

Let me know if the new documentation makes sense, and it would be great if someone could double-check that I didn't change the function's behavior.

The original fortran code had absolutely no documentation explaining the modifications due to canopy coverage, so I had to figure that out for myself. Let me know if you think it's not correct.

### DIFF
--- a/src/module_library/soil_evaporation_functions.h
+++ b/src/module_library/soil_evaporation_functions.h
@@ -56,6 +56,9 @@ const double par_energy_content = 0.219;
  *  total canopy leaf area index (`LAI`) and an exponential decay parameter
  *  (`k_canopy`), where `canopy_transmittance = exp(-k_canopy * LAI)`.
  *
+ *  A typical value of canopy albedo for crops is 0.23; see the text following
+ *  Equation 2 of Ritchie (1972).
+ *
  *  Note: This function was originally based on the `ALBEDO` subroutine of
  *  `SOILDYN.for` from DSSAT (https://github.com/DSSAT/dssat-csm-os).
  *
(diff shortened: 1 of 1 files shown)


eloch216:
> I have noticed the same issues with the code I have looked so far. Rather than do the detective work, I am putting comments asking about magic numbers.

😂 I don't mind doing the detective work. I think it's satisfying to figure it out. BTW, my next target will be `potential_soil_evaporation` from `soil_evaporation_functions.h`.

eloch216:
I will merge this later today, unless there are objections

PR #246, opened 2025-11-27
Update partitioning parameters with new water

yufenghe22:
I updated the partitioning coefficients with the new soil water. See comparison with observed biomass in the PDF file below. The fitting looks fine.
There's a small issue regarding the parameterization. I used Gray et al's weather data instead of the BioCro's. Their difference is below:

- Gray’s precipitation: These are daily sum values, measured at Willard Airport (Champaign, IL) for 2004-2008

- BioCro’s precipitation: the [WARM Data](https://github.com/cropsinsilico/Soybean-BioCro/tree/main/WeatherDataProcessing), Champaign, IL (40.0839 N, 88.2403 W; ~3 miles north of SoyFACE Farm)

We can see in 2005 (figure below), Gray's precipitation is higher than the BioCro's (more difference in 2006). Apparently this can significantly impact soil water and yield. I've been using Gray's data to better match other data of theirs, including soil moisture, LAI, gs.

I can update the test file based on this results. Will that work? I can add a second weather data based on the what I use for reproducibility. I could also re-opt the model with BioCro's weather data, which would mean that the partitioning coefficients are different between my current study and the BioCro's.

image

[fig_new_water_v127_aCO2.pdf](https://github.com/user-attachments/files/23809366/fig_new_water_v127_aCO2.pdf)

### DIFF
--- a/data/soybean2.R
+++ b/data/soybean2.R
@@ -44,6 +44,7 @@ soybean2 <- list(
         adaptive_max_steps     = 200
     ),
     initial_values = list(
+        days_stage2                 = 0,
         DVI                         = -1,
         Grain                       = 1e-05,
         Leaf                        = 0.06312,
@@ -64,7 +65,6 @@ soybean2 <- list(
         StemLitter                  = 0,
         sumes1                      = 0.125,
         sumes2                      = 0,
-        days_stage2                 = 0,
         TTc                         = 0
     ),
     parameters = list(
--- a/src/module_library/soil_evaporation2.h
+++ b/src/module_library/soil_evaporation2.h
@@ -208,7 +208,7 @@ string_vector soil_evaporation2::get_inputs()
         "lat",                    // latitude of the location
         "elevation",              // altitude in meters
         "lai",                    // Healthy leaf area index (m2[leaf] / m2[ground])
-        "bare_soil_albedo_max",   // Bare soil albedo (fraction) - dimensionless
+        "bare_soil_albedo_max",   // Maximum bare soil albedo - dimensionless
         "k_diffuse",              // light extinction coefficient
         "windspeed",              // m/s
         "rh",                     // fraction. dimensionless


eloch216:
Thanks for unifying `kd` with `k_diffuse`, and un-hardcoding `par_energy_content`!

> I can update the test file based on this results. Will that work?

I think this is the simplest approach, so I would recommend doing this.

> I can add a second weather data based on the what I use for reproducibility.

I don't think the Gray weather data needs to be part of the BioCro package.

> I could also re-opt the model with BioCro's weather data, which would mean that the partitioning coefficients are different between my current study and the BioCro's.

I don't think it's necessary to do that right now. The purpose for the BioCro regression test is to let us know if any module/model behavior changes. It will still be able to do that, even if the model is not parameterized for the BioCro weather data.

A few other comments:

1. The `direct_modules` part of `soybean2` has changed a lot. It's hard to tell if there are changes to the included modules, or if they were just reordered. If they were reordered, what was the reason for reordering them?

2. Why was `wsFun` added to `soybean2$parameters`? It is not required for any of the `soybean2` modules.

BTW, you can identify unused parameters by running `validate_dynamical_system_inputs` with `verbose = TRUE`. It will also let you know if the direct modules are in a suitable order for evaluation. Here's the command:

```r
with(soybean2, {validate_dynamical_system_inputs(
  initial_values,
  parameters,
  soybean_weather[['2002']],
  direct_modules,
  differential_modules,
  verbose = TRUE
)})
```

On this branch, the output includes:

```
The following parameters were not used as inputs to any module:
 wsFun
You may want to consider removing them for clarity
```

and

```
The direct modules need to be re-ordered before evaluation.
(This will be done automatically during dynamical_system construction.)
Here is a suitable ordering:
 format_time
 solar_position_michalsky
 shortwave_atmospheric_scattering
 incident_shortwave_from_ground_par
 soybean_development_rate_calculator
 soil_type_selector
 multilayer_soil_profile_avg
 stomata_water_stress_linear
 soil_surface_runoff
 soil_water_downflow
 soil_water_tiledrain
 soil_water_upflow
 sla_linear
 parameter_calculator
 height_from_lai
 canopy_gbw_thornley
 ten_layer_canopy_properties
 partitioning_coefficient_logistic
 stefan_boltzmann_longwave
 ten_layer_c3_canopy
 ten_layer_canopy_integrator
 soil_water_uptake
 carbon_assimilation_to_biomass
 partitioning_growth_calculator
 senescence_coefficient_logistic
 maintenance_respiration_calculator
```

On the `add-new-soil-water-modules` branch, the output instead includes:

```
Each parameter was used as an input to one or more modules
```

and

```
The direct modules are in a suitable order for evaluation.
```

eloch216:
> @eloch216 the tests should be good now. Once the other PR #249 is merged into this, I think this PR is ready to merge.

Oh, I thought you wanted to merge this one before #249. Whoops! I will look at that one next, then. (In a few hours.)

Issue #247, opened 2025-12-02
Potential mistake in new soil evaporation module


eloch216:
I have been taking a look at the new soil water modules, trying to understand them better and update the documentation to include references. Unfortunately I think there may be a mistake in the new soil evaporation module (https://github.com/biocro/biocro/blob/add-new-soil-water-modules/src/module_library/soil_evaporation2.h).

This module is based on the `SOILEV` subroutine from DSSAT, which calculates a daily soil evaporation rate. To create the BioCro module, the code was translated from Fortran to C++, but was otherwise not altered. So it still assumes a daily time step and calculates a daily rate, which is not appropriate for BioCro models.

Below I will try to explain my reasoning. Ideally it would be great if someone else could look it over and double check it. If we agree that the module is not correct, then we can try to fix it. Sorry for the long message, but I don't think there's a simple way to explain this.

----

To start, the DSSAT code can be found here: https://github.com/DSSAT/dssat-csm-os/blob/develop/SPAM/SOILEV.for

When I started looking at it, I was puzzled by the lines including `T` and set out to understand them. For example, line 109:

```fortran
ES = 3.5 * T**0.5 - SUMES2
```

The bottom of the file includes definitions for the variables used. `ES` is "Actual soil evaporation rate (mm/d)," `SUMES2` is "Cumulative soil evaporation in stage 2 (mm)," and `T` is "Number of days into Stage 2 evaporation (WATBAL); or time factor for hourly temperature calculations."

Why does `T` have two definitions? I don't know, but providing two definitions without citing a reference for the equation sure makes this code difficult to understand. Just looking at the code and associated comments, it's tough to tell what's going on.

Through some Googling, I was able to find a paper that gives a high-level overview of the DSSAT soil water calculations: da Silva et al. (2022) (https://doi.org/10.1016/j.agrformet.2022.109075). This paper also included a reference to the original description of the "two stage soil water evaporation" model used in DSSAT: Ritchie (1972) (https://doi.org/10.1029/WR008i005p01204).

Equation 6 from Ritchie (1972) is $\Sigma E_{s2} = \alpha t^{1/2}$, where $\Sigma E_{s2}$ is the "cumulative evaporation from the soil surface during stage 2 (mm)," $t$ is "time (days)," and $\alpha$ is an empirical factor that depends on soil type. Table 1 lists values of $\alpha$ for four soil types. Houston black clay has $\alpha = 3.5$ mm / day^0.5.

[...]

PR #249, opened 2025-12-02
Fix soil evaporation

yufenghe22:
This PR addresses the [issue #247](https://github.com/biocro/biocro/issues/247).

In addition to the major problem as mentioned in the issue #247 about the incorrect time factor increment by days, I also found an issue with the constant `evap_limit`. The literature suggests a value of 6 mm. Note that this is not a rate, but a total amount of water. This can also be seen in the Table 1 of Ritchie (1972). It means: Stage 1 ends when a total of 6 mm of water has evaporated. The original c++ code treated it as a rate by dividing it by 24, which is incorrect.

FYI, This PR has the base of another PR #246, so I want to merge there first. Hope this is okay.

### DIFF
--- a/src/module_library/soil_evaporation2.h
+++ b/src/module_library/soil_evaporation2.h
@@ -280,6 +280,9 @@ void soil_evaporation2::do_operation() const
 
     int constexpr nlayers = 6;
     double constexpr canopyHeight = 1.0;  // m
+    // soil hydraulic properties. Houston black clay
+    // See Table 1 in Ritchie (1972), https://doi.org/10.1029/WR008i005p01204
+    double constexpr soil_evaporation_alpha = 3.5;  // mm/day^(0.5)
 
     // Upper Limit of Stage 1 Cumulative Evaporation. Houston black clay.
     // Table 1. Ritchie (1972)
@@ -403,6 +406,7 @@ void soil_evaporation2::do_operation() const
                 sumes1_temp,
                 sumes2_temp,
                 evap_limit,
+                soil_evaporation_alpha,
                 days_stage2_temp);
             sumes1_temp = evap_comp.sumes1;
             sumes2_temp = evap_comp.sumes2;
@@ -431,6 +435,7 @@ void soil_evaporation2::do_operation() const
                 sumes1_temp,
                 sumes2_temp,
                 evap_limit,
+                soil_evaporation_alpha,
                 days_stage2_temp);
             sumes1_temp = evap_comp.sumes1;
             sumes2_temp = evap_comp.sumes2;
@@ -445,6 +450,7 @@ void soil_evaporation2::do_operation() const
                 sumes1_temp,
                 sumes2_temp,
                 evap_limit,
+                soil_evaporation_alpha,
                 days_stage2_temp);
             sumes1_temp = evap_comp.sumes1;
             sumes2_temp = evap_comp.sumes2;
--- a/src/module_library/soil_evaporation_functions.h
+++ b/src/module_library/soil_evaporation_functions.h
@@ -9,10 +9,6 @@
  * for multilayer soil profile.
  */
 
-// soil hydraulic properties. Houston black clay
-// See Table 1 in Ritchie (1972), https://doi.org/10.1029/WR008i005p01204
-double constexpr soil_evaporation_alpha = 3.5;  // mm/day^(0.5)
-
 double soil_albedo(
     double temp,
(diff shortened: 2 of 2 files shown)


src/module_library/soil_evaporation2.h
```diff
@@ -8,6 +8,18 @@
 
 namespace standardBML
 {
+/**
+ *  @class soil_evaporation2
+ *
+ *  @brief This module is based on the SOILEV subroutine from DSSAT, which calculates a daily soil evaporation rate. Here, we have made several changes to allow hourly calculation to match BioCro; Only compatible with a fixed-step Euler solver.
```
eloch216: Can you make sure the lines in the doxygen comment are 80 characters or less?

src/module_library/soil_evaporation2.h
```diff
@@ -249,7 +263,7 @@ string_vector soil_evaporation2::get_outputs()
     return {
         "sumes1",       // Cumulative soil evaporation in stage 1 (mm)
         "sumes2",       // Cumulative soil evaporation in stage 2 (mm)
-        "time_factor",  //time factor for hourly temperature calculations
+        "days_stage2",  // time factor for hourly temperature calculations
```
eloch216: This definition is no longer correct

yufenghe22: Had that in mind, somehow it slipped...

yufenghe22:
I did a quick comparison between the original and fixed versions. The Soil Evaporation Rate now is lower in early and late growing season, leading to slightly higher soil water content. The mid-season is little affected because Soil Evaporation Rate  is very low due to high LAI. See figures below.
image

image

src/module_library/soil_evaporation2.h
```diff
@@ -258,14 +272,17 @@ void soil_evaporation2::do_operation() const
...
-    int nlayers = 6;
-    double canopyHeight = 1.0;
-    double evap_limit = 0.25;  // 6/24 cm/hr. Default Evaporation
-                               // limit (cm) from SOILDYN.for, line 468 (variable U)
+    int constexpr nlayers = 6;
+    double constexpr canopyHeight = 1.0;  // m
+
+    // Upper Limit of Stage 1 Cumulative Evaporation. Houston black clay.
+    // Table 1. Ritchie (1972)
+    double constexpr evap_limit = 6.0;  // mm
```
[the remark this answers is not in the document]
eloch216: > I agree. However, shall we hold on adding them to the main parameter list? I worry that there will be other major changes needed to make the new soil water module as solid as the other modules of BioCro. More constant parameters may be identified.

I guess we can wait until later to move these to the main parameter list.

However, `soil_evaporation_alpha` is currently defined at global scope, which is usually something we try to avoid. Could you move its definition to `soil_evaporation2::do_operation` (like `evap_limit`) and make `alpha` an input to the functions that need its value? I think this might only be `supplemetal_evap_computation` in `soil_evaporation_functions.h`.

> Also there are already 140 parameters in the list.

There are a lot, but I think this is just inherent to modeling crop growth. Regardless of whether `soil_evaporation_alpha` and `evap_limit` are in the main parameter list, they are still parameters of the model. When they have hard-coded values, they are just "inconvenient" parameters whose values can only be changed by recompiling the BioCro package.

eloch216:
> > In this constructor for this module class, a value of `requires_euler = true` should be passed to `differential_module`
> 
> Wow. Very cool. This is new to me. It's pretty elegant.

Fortunately this situation doesn't come up very often. I was hoping that one day we could remove `thermal_time_senescence` and then we would no longer have a need to specify `requires_euler = true`. But I guess we will have to keep living with it for a while still.

BTW, have you ever tried running the new soil water modules with another solver? Or the Euler solver with a time step that isn't 1? I suspect the results might be weird.

> I did a quick comparison between the original and fixed versions.

It's nice that there isn't a huge change. But I suspect there might be other mistakes we will need to fix. Hopefully the simulations don't change too much.

eloch216:
The `boost_rkck54` solver is used for the default because it's required when running the full soybean + circadian clock model. The clock model needs smaller time steps and doesn't work well with the Euler solver. But the clock is completely independent from the rest of the model and can be run separately, so that's what we usually do.

I think this PR is almost ready to go (except for the failed tests), but I want to check out a few things, so please don't merge yet.

yufenghe22:
You may ignore the test failure here. Once it's merged with the other PR, it should good.

PR #250, opened 2025-12-04
New boost and c++17

eloch216:
CRAN is requiring us to drop our C++11 specification very soon. Here are the latest CRAN check results for BioCro, showing some warnings on several flavors of linux: https://cran.r-project.org/web/checks/check_results_BioCro.html

We have needed C++11 in order to support R version 3.6.0, because the associated version of Rtools uses gcc version 4.9.3, which only supports up to C++11.

So, to remove the C++11 requirement, we will need to stop supporting R version 3.6.0. Rtools has used newer versions of gcc that support C++17 since R version 4.0.0, so that will be our new minimum R version.

We will also need to move to a newer version of the boost libraries that does not use deprecated C++11 code.

Here I have tried doing both of these things. Let's see how it works out!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -35,6 +35,14 @@ be directly added to this file to describe the related changes.
 
 ## Minor User-Facing Changes
 
+- Increased the minimum supported R version to 4.1.0.
+
+  - This was necessary because the C++ standard was switched from C++11 to
+    C++17, as required by CRAN.
+
+  - The version of boost bundled with BioCro (1.71.0) was also not compatible
+    with C++17, requiring an update to a new version (1.89.0).
+
 - A simple multidimensional root solving library has been added. See header
   file `src/math/roots/multidim/zeros.h` for example usage. Currently, only one
   method is available: Broyden's method. This library is intended to support


justinmcgrath:
CRAN doesn't want any specification in the package, and they'll ask for justification for why we've said it needs 17. It would be in our best interest to specify it, but I don't know what we say here. We could say Boost needs it, but I don't know if they'll look that up.

eloch216:
> CRAN doesn't want any specification in the package

Are you sure? `R CMD check` doesn't produce a `NOTE` about requiring C++17. Although I am also not sure that the C++17 requirement is actually used for anything. E.g. the online test results for windows-latest indicate that a C++14 compiler was used (unless I am not interpreting this correctly): https://github.com/biocro/biocro/actions/runs/19939800929/job/57174444457#step:10:79

Update: I am interpreting that incorrectly. 14.2.0 is the version of gcc. It's just a coincidence that 14 is also a version of C++.

leavesandlemmas:
> CRAN doesn't want any specification in the package, and they'll ask for justification for why we've said it needs 17. It would be in our best interest to specify it, but I don't know what we say here. We could say Boost needs it, but I don't know if they'll look that up.

Is it too cheeky to simply say that BioCro the C++ library is upgrading to C++17 as the minimum standard, so the R package which wraps BioCro has to upgrade too?

justinmcgrath:
> Are you sure? `R CMD check` doesn't produce a `NOTE` about requiring C++17.

I saw a comment somewhere about it, but if there's no note, then let's not worry about it.

eloch216:
> > Are you sure? `R CMD check` doesn't produce a `NOTE` about requiring C++17.
> 
> I saw a comment somewhere about it, but if there's no note, then let's not worry about it.

Sounds good.

I also just removed `makevars.win`. The version of `make` used in Rtools40 is 4.3. So, according to the note in `makevars.win`, we no longer need it since we are always using make 3.8 or greater.

PR #252, opened 2025-12-07
introduce a new water stress function

yufenghe22:
For this PR, I want to introduce a new and simple water stress function that incorporates both the supply (soil water) and demand (ET) sides. It's more realistic despite being a bit "crude". It's probably not that "crude" because it's based on the same mechanistic approach that you would use with water potentials. The only difference is that I use proxies (soil water and ET) instead of actual water potentials in the soil and leaf. It's possible to derive soil water potential with the new soil water module, but that's not what the goal is for this PR.
We have some water stress functions in our model to estimate a scaling factor (called stomatal water stress) that is applied to scale the ball-berry coefficients. They all have the same issue, but just to take the linear version as an example. 
$f_{ws}^{linear} = \frac{\theta - \theta_{wp}}{\theta_{fc} - \theta_{wp}}$. This function only sees the changes in the soil water ($\theta$) side, completely ignoring the demand side (Evapotranspiration). The soil water is a rather slow process and this is why the $f_{ws}$ shows very mild response to VPD. In fact, it has no diurnal dynamic at all.
A linear $\theta$ function like this treats a 1% drop in water the same way at field capacity as it does near wilting point. This is not what we want to represent the water stress felt by leaves. In reality, the water stress should be tied to both the soil and leaf water potentials. In dry soil, as soon as the soil dries past a threshold, $\Psi_{soil}$ drops exponentially, making it much harder for leaf to withdraw water than in wet soil.

### How the new water stress function works
Following the [Van den Honert Equation](https://doi.org/10.5194/hess-2-31-1998) for plant hydrology, $\Psi_{leaf} = \Psi_{soil} - R_{hydro} * E$, where $\Psi$ is water potential, $E$ is transpiration rate, $R_{hydro}$ is total hydraulic resistance (soil + root + xylem), with units of Pressure per Flux (e.g., $MPa \cdot m^2 \cdot s \cdot mol^{-1}$). Our goal is to estimate the water potential ($\Psi_{leaf}$) and stress in the leaf. To map the physical quantity to the 0-1 stress factor ($f_{ws}$), we assume a critical water potential $\Psi_{crit}$ and use it to scale the above equation, so we have:

$$\frac{\Psi_{leaf}}{\Psi_{crit}} = \frac{\Psi_{soil}}{\Psi_{crit}} -  E * \frac{R_{hydro}}{\Psi_{crit}}$$

which equals,

$$f_{ws} = f_{ws}^{linear} - E *R_{norm}$$

Where $R_{norm}$ is the normalized hydraulic resistance (unit: 1/E), representing the fraction cost of moving 1 unit of water. 
Now our new $f_{ws}$ is nonlinearly responding to changes in both supply (soil water) and demand (E). Since E has a clear diurnal signal, $f_{ws}$ also has a diurnal signal. This makes more sense because a plant at noon is likely under more hydraulic tension than a plant at dawn, even if the soil water content hasn't changed.

### Avoid the Module Cyclic issue

One issue I encountered was creating a cyclic module issue with the $E$ in the equation. Note that since the only soil water uptake is through ET, the `total uptake = ET ` in our current soil water module. So when I say uptake, it's the same as ET or E. The cyclic issue happens like this if we directly use the ET for the water stress calculation:
<p align="center">
water_stress → gs → ET (uptake) → water_stress
</p>
Since soil water potential changes slowly (it doesn't jump wildly in 1 hour), the error introduced by using "1-hour-old" uptake data is negligible. Therefore, I use the uptake from the previous hour to calculate Stress for the current hour. For this, I added a state variable named `uptake_laststep`, simply tracking the last step's value total uptake (E).  

----
@leighmatth Tag you here for review. For some reason, I cannot make you a reviewer from the list on the right.


leavesandlemmas:
To follow up from my comment, I think we have something like this: 

$$
E = g_t ( w_a  - w_i)
$$

where $E$ is transpiration rate, $g_t$ is total conductance to water vapor, and $w_a - w_i$ is the difference in vapor molar fraction.  But we also have something like $E = k (\psi_s - \psi_l)$ for a difference in leaf and soil water potential $\psi_l - \psi_s$, and conductance $k$. Is that correct?

So we should have a differential equation that's like  

$$
 \frac{d \psi_l }{dt} \propto k (\psi_s - \psi_l) - g_s (w_a - w_i)  
$$ 

The two sides can't be equal because the unit dimension aren't  compatible.  Presumably $\psi_l$ and $w_i$ are related so this isn't perfect.

yufenghe22:
> This was based on the soil being dryer, not the stomata being more sensitive. That's why I was asking what drives the higher StoamtaWS in elevated [CO2].

Yes. It is soil water driving, actually in both aCO2 and eCO2, but the amplification of eCO2 is assumed larger. The friction term $R=R_{base}/x^n$, where $x=f_{ws}^{linear}$ representing soil dryness. StomataWS is sensitive to the $n$ parameter because it's a power scale. In my calibration, $n=1$ for aCO2 (mild response to dry soil) and $n=5.7$ for eCO2 (strong response to dry soil). For wet soil (x close to 1), no matter how large $n$ is, it does not scale the friction much. Does this make sense?

yufenghe22:
> Does the difference in n here result in R terms that correspond to those K_leaf observations?

The paper's interesting, which seems to support our addition of the E*R term as they state:
`"transpiration demand exceeds the maximum capacity... high tension in the water column can cause cavitation"`. They also show "Midday Depression", which is nicely captured by the diurnal change of $E$.
 $K_{leaf}$ (conductance) in the paper is the inverse of $R$ (resistance). Our $R$ is normalized, but with the same concept. The paper shows $K_{leaf}$ decreases under eCO2, same direction as my assumption that $R$ increases under eCO2. However, they did not investigate soil water status' impacts on that. With only two days of data in July and August, I'm not sure how dry the soil were in their data. Our $n$ is tied to $x$, maybe their data are in "Wet/Mild" zone of our model where the curves for aCO2 and eCO2 are quite close.

@justinmcgrath In terms of calibration of $n$ and $R_{base}$, I used the Ci vs soil water content (swc) data for eCO2 and aCO2 from the Gray et al. You can see the significant difference in the gradient of Ci vs ABA in Fig. 5a, similar for Ci vs swc. Of course, Ci is related to stomata closure. I want to capture the response of Ci to swc, so my optimization goal was to match the `gradients` of them under aCO2 and eCO2, respectively. (It might be too premature for the model to match Ci and swc individually, especially Ci). In general I want to show how the CO2 fertilization effect changes with $n$, so I also vary $n$ to indicate such responses. This can explain at least one of the potential feedback mechanisms that cause reduced yield ratio with VPD.

PR #253, opened 2025-12-09
Release v3.3.0

eloch216:
CRAN is no longer allowing C++11, and will remove the BioCro package if we do not update to C++17. So, it's important to make a new release with a C++17 specification. Version 3.3.0 will meet that requirement, and also make several new updates available from CRAN as compared to the previous version, 3.2.0.

Here, to get ready for the new release, I have just updated the `DESCRIPTION`, the `cran-comments`, and the `NEWS`.  These changes will be easier to see on PR #254 .

Let me know if the `NEWS` is missing anything important.

### DIFF
--- a/.github/workflows/document_quantities.yml
+++ b/.github/workflows/document_quantities.yml
@@ -48,7 +48,7 @@ jobs:
         while read -r cmd
         do
           eval sudo $cmd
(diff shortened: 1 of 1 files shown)


eloch216:
The next step is that @justinmcgrath will submit the package (from this branch) to CRAN. Once it is accepted, I will merge these PRs.

justinmcgrath:
This is on CRAN now.

https://cran.r-project.org/web/packages/BioCro/index.html

PR #257, opened 2025-12-14
Broyden template error hotfix

leavesandlemmas:
To fix the template related compiler errors / warnings.

### DIFF
--- a/vignettes/BioCro.Rmd
+++ b/vignettes/BioCro.Rmd
@@ -8,8 +8,13 @@ vignette: >
 ---
 
 ```{r, include = FALSE}
+options(
+  crayon.enabled = FALSE, # crayon colors cause latex issues
+  cli.num_colors = 1
+)
+
 knitr::opts_chunk$set(
-  error = FALSE,
+  error = FALSE, # stop on errors to help with troubleshooting
   collapse = TRUE,
   comment = "#>",
   fig.width = 6,
--- a/vignettes/web_only/BioCro-II_Paper--Section-1.1-example.Rmd
+++ b/vignettes/web_only/BioCro-II_Paper--Section-1.1-example.Rmd
@@ -13,8 +13,13 @@ vignette: >
 bibliography: references/BioCro.bibtex
 ---
 ```{r, setup, include=FALSE}
+options(
+  crayon.enabled = FALSE, # crayon colors cause latex issues
+  cli.num_colors = 1
+)
+
 knitr::opts_chunk$set(
-  error = FALSE,
+  error = FALSE, # stop on errors to help with troubleshooting
   collapse = TRUE,
   autodep = TRUE
 )
--- a/vignettes/web_only/a_practical_guide_to_biocro.Rnw
+++ b/vignettes/web_only/a_practical_guide_to_biocro.Rnw
@@ -85,11 +85,16 @@ knitr::opts_knit$set(latex.options.xcolor = 'dvipsnames')
 
 % Set things up for R evaluation:
 <<preliminaries,echo=FALSE>>=
-options(crayon.enabled = FALSE, cli.num_colors = 1) # crayon colors cause latex
-                                                    # issues, so do not use them
-knitr::opts_chunk$set(error=FALSE) # do stop on errors so we can immediately
-                                   # identify problems with the vignette code
-knitr::opts_chunk$set(fig.width=5, fig.height=3)
+options(
+  crayon.enabled = FALSE, # crayon colors cause latex issues
+  cli.num_colors = 1
+)
+
+knitr::opts_chunk$set(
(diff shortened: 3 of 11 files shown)


eloch216:
> I am not sure how to test if it was successful.

The warnings do show up on the "Document Quantities" workflow. Here's a recent example: https://github.com/biocro/biocro/actions/runs/20243536773/job/58117523607#step:8:70

Yet, with the same OS and compiler, there are not any warnings on the "R CMD check" workflow: https://github.com/biocro/biocro/actions/runs/20110011928/job/57704606985#step:10:77

One difference between the workflows is that `R CMD INSTALL` is used on "Document Quantities" while `R CMD check` is used on "R CMD check." So right now my best guess is that perhaps `changes-meaning` warnings are ignored by `R CMD check` but they can still cause problems elsewhere. We don't specify any warning options in `Makevars` that would override the warning.

I will try adding `-Wchanges-meaning` to `Makevars` (on another branch) and see if it causes an `R CMD check` failure on Ubuntu.

leavesandlemmas:
> So right now my best guess is that perhaps `changes-meaning` warnings are ignored by `R CMD check` but they can still cause problems elsewhere. We don't specify any warning options in `Makevars` that would override the warning

So I have noticed that whether or not the compiler emits C++ compilation warnings seems to depend on the OS. Linux and macOS show warnings (both from boost and my code)  whereas windows shows bumpkiss. I thought it was the compiler that was different, but @justinmcgrath  did a test this morning using the same compiler version (gnu 13.2  I think) but on a linux and windows. The linux one emitted errors but the windows one did not.

That would suggest to me that it's not the compiler per se (the compiler can emit warnings) but something about R's build system which decides which warnings to emit.

leavesandlemmas:
> I just noticed that there are several places in the Broyden files where code has been commented out:
>
>     * A section in `broyden.h` that begins with `UPDATE FORMULA`
>     * A section in `broyden.h` that begins with `inv_jac` (this is not C++ code but it sorta looks like C++ code which is confusing)
>     * A constructor for `zero_finding_method` in `zeros.h` 
>     * Two inline implementations of an `invert` function in `zeros.h`.

All of these changes have already been made in another branch (at least in `solve-linear-systems`). The second comment isn't intended to be code, but mathematical notation (which in ascii looks like code). I would prefer to simply defer these changes to a PR for `solve-linear-systems`. Otherwise the hotfix isn't that much of a hotfix.

eloch216:
This has been annoying to figure out.

At first, I could not reproduce the issue with `a_practical_guide_to_biocro.Rnw` locally. Then I decided to update all my R packages using `update.packages()`. Afterwards, I was able to reproduce the error, although I am not sure which update in particular is causing the issue, since many packages were updated. The error only occurs when using `pkgdown::build_site()` or `pkgdown::build_articles()` to render the vignette. If I just use `tools::buildVignette()`, there are no issues.

Digging deeper, there is more information about the error in `a_practical_guide_to_biocro.log`:

```
! LaTeX Error: Unicode character ^^[ (U+001B)
               not set up for use with LaTeX.
```

Unicode 001B is the `ESC` character. Searching through `a_practical_guide_to_biocro.tex`, I found this unicode character in the output from R code blocks that cause errors, such as:

```
{\ttfamily\noindent\bfseries\color{errorcolor}{\#\# [1m[33mError[39m in `as.data.frame()`:[22m\\\#\# [33m![39m Caught exception in R\_run\_biocro: Thrown by dynamical\_system::dynamical\_system: the supplied inputs cannot form a valid dynamical system\\\#\# \\\#\# \\\#\# [pass] No quantities were defined multiple times in the inputs\\\#\# \\\#\# [fail] The following module inputs were not defined:\\\#\# \ Leaf from the 'maintenance\_respiration\_calculator' module\\\#\# \ Leaf from the 'parameter\_calculator' module\\\#\# \ Leaf from the 'senescence\_logistic' module\\\#\# \ Leaf from the 'partitioning\_growth' module\\\#\# \\\#\# [fail] The following differential module outputs were not part of the initial values:\\\#\# \ Leaf from the 'senescence\_logistic' module\\\#\# \ Leaf from the 'maintenance\_respiration' module\\\#\# \ Leaf from the 'partitioning\_growth' module\\\#\# \\\#\# [pass] There are no cyclic dependencies among the direct modules.}}\end{kframe}
```

Apparently these weird sequences like `[1m[33m` are color formatting surrounding parts of R error messages: https://github.com/r-lib/rlang/issues/1366

After some more Googling, I learned that these "crayon" colors can be disabled by adding this R command to one of the first code blocks in a document: `options(crayon.enabled = FALSE, cli.num_colors = 1)`.

Adding this to `a_practical_guide_to_biocro.Rnw` fixed the issue. It turns out the same code also needed to be added to `dynamical_systems.Rmd`. I figured I should just go ahead and add it to all the vignettes.

PR #258, opened 2025-12-15
Enable `changes-meaning` compiler warnings

eloch216:
A test to see if we can detect these warnings in our checking workflow

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -1,4 +1,4 @@
-PKG_CPPFLAGS+=-I../src/inc -DR_NO_REMAP
+PKG_CPPFLAGS+=-I../src/inc -DR_NO_REMAP -Wchanges-meaning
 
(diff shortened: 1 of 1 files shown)


justinmcgrath:
That flag isn't accepted by all compilers. It looks like CRAN flags it because of that, and because of how they handle it, you couldn't include it conditionally. If you did some check to see if it was available, and only included it for compilers that allowed it, CRAN would still flag it as a warning!

Maybe you could check when you're running on GitHub and only include it there.

```makefile
ifeq ($(GITHUB_ACTIONS), true)
  ifeq ($(shell $(CXX) -Wchanges-meaning -c -x c++ -o /dev/null /dev/null 2>/dev/null && echo flag_present), flag_present)
    PKG_CXXFLAGS += -Wchanges-meaning
  endif
endif
```

I'm not sure that works. We wouldn't get errors for local builds, just GitHub actions, but we should see a problem in pull requests.

PR #259, opened 2025-12-16
Try checking for `changes-meaning` compiler warnings

eloch216:
Trying a different technique for enabling `changes-meaning` warnings only on specific operating systems, based on Justin's comments on PR #258

### DIFF
--- a/.github/workflows/R-CMD-check-changes-meaning.yaml
+++ b/.github/workflows/R-CMD-check-changes-meaning.yaml
@@ -42,8 +42,8 @@ jobs:
           inputFile: '.github/workflows/strategy_matrix.json'
           filter:
             '[?configName == ''R 4.1.0 on Ubuntu''
-                || ?configName == ''R release version on Ubuntu''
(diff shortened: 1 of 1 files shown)


eloch216:
I made a new workflow called `R-CMD-check-changes-meaning` that adds the `-Wchanges-meaning` compiler flag on Ubuntu and runs `R CMD check` with `error-on` set to `error` (rather than `warning`). Together, these options mean the workflow has the following behavior:
- If the compiler reports a `changes-meaning` warning, package installation will fail and there will be an `ERROR` from `R CMD check`, causing the GitHub check to fail.
- If there is no issue with the compiler, `R CMD check` will report a `WARNING` due to a non-portable compiler flag, but this will not cause the GitHub check to fail.

This workflow seems to behave as expected. On this branch, there are three check failures from the new workflow, corresponding to three different versions of R running on Ubuntu. With Scott's fixes, all the checks pass (see PR #260 ).

PR #260, opened 2025-12-16
Testing hotfix and warnings

eloch216:
This branch includes the changes in PRs #257 and #259 .

### DIFF
--- a/.github/workflows/R-CMD-check-backend.yaml
+++ b/.github/workflows/R-CMD-check-backend.yaml
@@ -41,6 +41,8 @@ on:
         type: boolean
       error-on:
         type: string
+      changes-meaning:
+        type: boolean
 
 env:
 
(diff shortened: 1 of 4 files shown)


justinmcgrath:
> which suggests using warning flags to ensure C++ code is standard conformant

We have a way to make it portable, but the way CRAN performs its checks doesn't let us use it. It is one more questionable CRAN policy; if a platform didn't support the flag, then it would fail during compilation, so there's no point checking it. It's entirely redundant. It looks like there's discussion about standardizing this, so maybe in the future it will be a standard compiler option.

eloch216:
I think a workaround is to run `R CMD check` two different ways on Ubuntu:
1. Specify the `-Wchanges-meaning` compiler flag and ignore warnings from `R CMD check`; this will ignore the spurious warning about non-portable paths, but would also ignore any real warnings.
2. Do not specify the `-Wchanges-meaning` compiler flag and check for warnings from `R CMD check`; this will catch any real warnings ignored in the check above.

Together, these will catch any `changes-meaning` issues, as well as any other issues with the package. This strategy is now implemented on #259 . It's a bit clunky but it gets the job done. It's also tested here to show that it works.

PR #261, opened 2026-01-05
Basic linear algebra for solving linear systems of equations (for use in nonlinear equation solving). 

leavesandlemmas:
This PR introduces a simple linear algebra library. The main barrier to implementing nonlinear equation solving algorithms is the ability to solve systems of linear equations. This PR implements an algorithm for LU decomposition with partial pivoting. 

To simplify the implementation of the algorithm used in Broyden's method to update the approximation jacobian,  this PR also introduces two types for matrix - vector math:

1. `linalg::vector<double, Dim> v`  column or row vectors of dimension `Dim`. Entries are indexed by brackets `v[i]` or parentheses `v(i)`.
2. `linalg::matrix<double, Row, Col> A` dense matrices of dimension `Row * Col`.  Entries are indexed by parentheses `A(i, j)` as multiple argument indexing is a  C++23 feature.

Vector space operations and a couple of matrix-vector operations are provided, with operators overloaded.  Let `A,B` denote matrices , `u, v` vectors, and `s, t`  scalars (which are `double` or some other floating point type ).  

1. `s * v  + u * t`  scalar multiplication and vector addition (both are commutative).   
2. `s * A + B * t`  scalar multiplication and matrix addition (both are commutative) (i.e., matrices are themselves a vector space). 
3. ` u / s ` and  `A / s` are defined for scalar multiplication. 
4. `A * v` indicates a matrix-vector product; and `v * A` treats `v` as a covector or row-vector (the dual action).
5. The compound assignment operators are also provided.
6. `dot(u, v)` computes the dot-product and `outer(u,v)` computes the outer-product. 

The vector space operations are implemented using a technique called expression templates. The idea is operations should be automatically fused into a single for-loop. For instance, to compute the sum `u + v + w`, we want one for-loop evaluating `u[i] + v[i] + w[i]` rather than two loops, one for `tmp = u + v` then one for `tmp + w`. 

Note Boost's own BLAS library uses expression templates (as do other C++ linear algebra libraries), but this is a simple version. I would have used Boost's BLAS library but the documentation is shit (and the code is dense). If we want to use a pre-existing library, I would recommend `Eigen3` but implementing the basics here avoids the dependency issue of using `Eigen3` (although it is a header only library). 

Unfortunately, it seems I created the original branch from one of the branches with the new soil water modules, so they are included in this branch. Perhaps, the base branch should be one of those.


eloch216:
> Unfortunately, it seems I created the original branch from one of the branches with the new soil water modules, so they are included in this branch. Perhaps, the base branch should be one of those.

I think this is a real issue, even beyond the problem of distinguishing the linear algebra changes from the soil water changes in the `git diff`.

On one hand, this branch cannot be merged into `develop`, since the soil water modules are not yet ready to be included there. On the other hand, this branch probably should not be merged into the `add-new-soil-water-modules` branch either, since it has nothing to do with the soil water modules.

To fix this, I would suggest the following:

1. Copy `src/math/linalg` and `src/math/roots/multidim` to some other location on your computer that isn't tracked by `git`.
2. Make a new branch based on `develop`. (Make sure you have the latest version that includes the two PRs that were recently merged.)
3. Paste `src/math/linalg` and `src/math/roots/multidim` back into the repository. It's possible that simply pasting the files in `multidim` will not work, and some kind of "merge conflict resolution" will be required.

This should create a new branch that only contains the new linear algebra stuff. Some `git` history will be lost but that's probably not a big problem.

Issue #256, opened 2025-12-14
Build issues with older compilers?


eddelbuettel:
On Ubuntu 22.04 a standard package build (in the context binary package creation for [r2u](https://eddelbuettel.github.io/r2u/)) failed:

```
g++ -std=gnu++17 -MT module_library/module_library.o -MMD -MP -MF .deps/module_library/module_library.Td -I"/usr/share/R/include" -DNDEBUG -I../src/inc -DR_NO_REMAP      -fpic  -g -O2 -ffile-prefix-map=/build/r-base-xupQTd/r-base-4.5.2=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2   -c -o module_library/module_library.o module_library/module_library.cpp
In file included from module_library/../math/roots/multidim/broyden.h:7,
                 from module_library/broyden_test.h:6,
                 from module_library/module_library.cpp:14:
module_library/../math/roots/multidim/zeros.h:44:11: error: declaration of ‘using vec_t = root_multidim::vec_t<dim>’ changes meaning of ‘vec_t’ [-fpermissive]
   44 |     using vec_t = vec_t<dim>;
      |           ^~~~~
module_library/../math/roots/multidim/zeros.h:16:7: note: ‘vec_t’ declared here as ‘using vec_t = struct std::array<double, dim>’
   16 | using vec_t = std::array<double, dim>;
      |       ^~~~~
module_library/../math/roots/multidim/zeros.h:45:11: error: declaration of ‘using mat_t = root_multidim::mat_t<dim>’ changes meaning of ‘mat_t’ [-fpermissive]
   45 |     using mat_t = mat_t<dim>;
      |           ^~~~~
module_library/../math/roots/multidim/zeros.h:20:7: note: ‘mat_t’ declared here as ‘using mat_t = struct std::array<std::array<double, dim>, dim>’
   20 | using mat_t = std::array<vec_t<dim>, dim>;
      |       ^~~~~
module_library/../math/roots/multidim/zeros.h:46:11: error: declaration of ‘using result_t = struct root_multidim::result_t<dim>’ changes meaning of ‘result_t’ [-fpermissive]
   46 |     using result_t = result_t<dim>;
      |           ^~~~~~~~
module_library/../math/roots/multidim/zeros.h:34:8: note: ‘result_t’ declared here as ‘struct root_multidim::result_t<dim>’
   34 | struct result_t {
      |        ^~~~~~~~
[...]

PR #262, opened 2026-01-06
Release v3.3.1

eloch216:
This PR fixes the compiler warnings mentioned in issue #256 , and also adds new checks to our testing actions to make sure we catch any future `changes-meaning` warnings before they cause problems. It also fixes some vignette building issues that popped up since our last release.


eloch216:
@justinmcgrath We will need to submit this branch to CRAN (once it passes the tests)

justinmcgrath:
I received the notice that this is on its way to CRAN now. I had to push one change to update the date in case that affects the sister PR to this.

Issue #265, opened 2026-01-06
Mistake in `potential_soil_evaporation` function


eloch216:
I am currently working on understanding and documenting the `potential_soil_evaporation()` function from `soil_evaporation_functions.h` (in the new soil water modules) and noticed a potential issue.

When looking through the code, one thing that caught my attention was the presence of two LAI-based exponential decay functions:
- One is found in the calculation of `kcb`, where `kcb = max(0.0, kcbmin + (kcbmax - kcbmin) * (1.0 - exp(-1.0 * skc * lai)));`
- One is found in the calculation of `attenuation`, where `attenuation = exp(-ksevap * lai);`

Having compared `potential_soil_evaporation()` to the original DSSAT code, I believe this is a mistake, and the second "attenuation" coefficient should not be used here. I am already working on fixing this. But I wanted to explain my reasoning here to make sure others agree.

----

Overall, the soil water modules seem to follow the `SPAM` subroutine of `SPAM.for`. Here, "SPAM" is an acronym, where "SPA" stands for "Soil-Plant-Atmosphere" and I am not 100% sure what the "M" is for; maybe "Model?" This subroutine is a bit difficult to follow because it includes calls many other subroutines depending on user inputs. Many of these subroutines don't have return values; instead, they modify the value of pre-existing variables, which also makes everything difficult to follow. So, it's very easy to make mistakes when translating into C++.

To calculate the potential soil evaporation rate, the `SPAM` subroutine first calls the `PET` routine, which calculates potential evapotranspiration. Within `PET` itself, several options are present, and the particular one is determined by the value of the `MEEVP` user input. If this input is `S` or `T` ("short" or "tall" crop with a dual-coefficient approach), then the `PETASCE` subroutine is called. Much of the code in our `potential_soil_evaporation()` function is taken from `PETASCE`. There are a few `if` statements in `PETASCE` where the calculations are different for `S` and `T`, but our code only includes the `T` branches of the `if` statements. So it seems we are using the model for tall crops. The `PETASCE` subroutine calculates several things, among them the "soil evaporation coefficient" (`KE`).

[...]

Issue #243, opened 2025-11-26
Duplicated code and hard-coded values in new soil water modules


eloch216:
I have started taking a look at the new soil water modules, and I think there are some instances of code duplication / hard-coded values that should be addressed. Here are a few I have noticed so far:

1. The new `BioCro:soil_evaporation2` module has an input called `kd`. I believe this is identical to `k_diffuse` as used by other modules, so it should probably be renamed to `k_diffuse`.
2. There is a hard-coded value of `par_energy_content` in `soil_evaporation_functions.h`.
3. The `reference_evapotranspiration` function in `soil_evaporation_functions.h` calculates a value of the psychrometric parameter $\gamma$. This is also calculated in `leaf_energy_balance.cpp`. We should make sure they are calculated the same way.
4. It looks like the `reference_evapotranspiration` also has its own calculation of solar position, instead of using the outputs from the `BioCro:solar_position_michalsky` module.
5. There is a definition for $\pi$ called `pie` that differs from the one in `constants.h`.

I think there are probably a lot more cases like these. I assume they resulted from directly translating old fortran code. Nevertheless, we need to make sure calculations are done consistently across different BioCro modules.

I am particularly suspicious of the new `evapotranspiration` functions. They might be repeating what is already done in our canopy transpiration calculations.

yufenghe22:
> * The `reference_evapotranspiration` function in `soil_evaporation_functions.h` calculates a value of the psychrometric parameter 
>     γ
>   . This is also calculated in `leaf_energy_balance.cpp`. We should make sure they are calculated the same way.
> * It looks like the `reference_evapotranspiration` also has its own calculation of solar position, instead of using the outputs from the `BioCro:solar_position_michalsky` module.

Thanks for pointing out, I probably would not know about this. I will go through the `evapotranspiration` and try to remove repetitions, in addition to finishing running my experiments and complete the results.
FYI, the reason to use this `evapotranspiration` is that 1) Umar found that the BioCro's soil evaporation is unrealistic; and 2) I once compared the soil evaporation between them, and the new one does seem to show more realistic seasonal patterns, meaning that the soil evaporation is sensitive to LAI. When canopy is big, the soil evaporation is low.

eloch216:
Many of these have been addressed by our big updates to the soil water modules, so I will close this now. There are probably still improvements that can be made, but they can be done later.

PR #267, opened 2026-01-07
Potential soil evaporation

eloch216:
This PR addresses issue #265 and also adds new documentation (primarily to `potential_soil_evaporation` but also some to `reference_evapotranspiration`).

While working on this, I noticed that the `potential_et` input argument of `potential_soil_evaporation` was unused, so I removed it. With this change, there was no reason to calculate `potential_et` in the `BioCro:soil_evaporation2` module, so I removed that calculation as well. With this change, the `potential_evapotranspiration` function was not used for anything, so I also removed this function.

I also discovered a nice reference for these potential soil evaporation rate calculations: DeJong & Thorp (2017) (https://doi.org/10.13031/trans.12321). This paper discusses an improvement to DSSAT that occurred around 2017, where the "single crop coefficient" approach was replaced by the "dual crop coefficient" approach. In terms of potential soil evaporation, Equation 5 in the paper is replaced by Equation 8. See the paragraph just before "Methods and Materials" section for some key reasoning. Looking at the original code in `potential_soil_evaporation`, it is clear that the "extra" LAI calculations mentioned in the issue correspond to Equation 5 from the paper, and should not be used.

After removing these lines, the behavior of the model has changed a bit. I will share a few plots below in another comment.

### DIFF
--- a/src/module_library/soil_evaporation_functions.h
+++ b/src/module_library/soil_evaporation_functions.h
@@ -362,12 +362,12 @@ double reference_evapotranspiration(
  *  @return The potential soil evaporation rate `ES_0`; same units as `ET_0`
  */
 double potential_soil_evaporation(
-    double SK_c,      // dimensionless
-    double K_cb_max,  // dimensionless
-    double K_cb_min,  // dimensionless
-    double LAI,       // dimensionless
-    double height,    // m
-    double ET_0       // any transpiration rate units such as mm / hr
+    double const SK_c,      // dimensionless
+    double const K_cb_max,  // dimensionless
+    double const K_cb_min,  // dimensionless
+    double const LAI,       // dimensionless
+    double const height,    // m
+    double const ET_0       // any transpiration rate units such as mm / hr
 )
 {
     // Check for bad inputs
@@ -388,8 +388,8 @@ double potential_soil_evaporation(
     }
 
     // Set constants
-    double constexpr K_r = 1.0;       // dimensionless
-    double constexpr f_w = 1.0;       // dimensionless
+    double constexpr K_r = 1.0;  // dimensionless
+    double constexpr f_w = 1.0;  // dimensionless
 
     // Equation 6 from DeJonge & Thorp (2017)
     double const K_cb = K_cb_min + (K_cb_max - K_cb_min) * (1.0 - exp(-1.0 * SK_c * LAI));  // dimensionless


eloch216:
Here are a few plots showing changes to biomass, soil evaporation, and soil water content on this branch. The soil evaporation rate has increased, and a weird jump in the rate has disappeared.

image

image

image

yufenghe22:
This is [the observed data on soybean field I was referring to](https://www.sciencedirect.com/science/article/pii/S0378377414001450#fig0020). Data in Fig. 4 shows seasonal pattern of soil evaporation. My understanding is that, in early season when canopy is small and soil is more exposed, the soil evaporation is mostly driven by precipitation and radiation. Once canopy get bigger, the soil evaporation becomes very small. This is what our previous result predicted. The observed data also suggests the soil evaporation is small till the end of season, but our model results suggest the evaporation will become large again at the end when the canopy senesced. I think this may be due to dead leaf cover that is not accounted for in our model, which I would not worry much for now.

eloch216:
> > Here are a few plots showing changes to biomass, soil evaporation
> 
> How did you compare new and stored results? Did you have to save the old result, check out to this PR, compile, get the new result, and then compare them? Just wonder if there's an easier way to do it.

There is an easier way, but it only works for one year (2002). The main idea is to use the stored regression test data as the "old" version, and results from the currently-compiled version of BioCro as the "new" version. Here's how I do it:

1. Use the online GitHub interface to navigate to the branch I want to use for the "old" data, and download the appropriate regression test data from `tests/test_data/`. For the plots above, I went to the `add-new-soil-water-modules` branch and downloaded `soybean2_simulation.csv`. I temporarily store this file in my local `tests/test_data` directory, overwriting whatever version was previously there. (Note: If you have made a change to a module but haven't yet updated the stored regression tests, you can skip this step, since the regression test will still represent the "old" version.)

2. Make sure I have installed BioCro from the local repository.

3. Set the R working directory to `tests/testthat/` and run `source('test.CropModels.R`).

4. Follow the instructions in `test.CropModels.R` to compare the new and old versions. For these graphs, I used:

```r
INDEX <- 4
compare_crop_output <- BioCro::compare_model_output(PLANT_TESTING_INFO[[INDEX]])
lattice::xyplot(Leaf + Stem + Root + Grain ~ time, group = version, data = compare_crop_output, type = 'l', auto = TRUE, grid = TRUE)
lattice::xyplot(soil_evaporation_rate ~ time, group = version, data = compare_crop_output, type = 'l', auto = TRUE, grid = TRUE, ylim = c(-0.0001, 0.001))
lattice::xyplot(soil_water_content_1 + soil_water_content_2 + soil_water_content_3 + soil_water_content_4 ~ time, group = version, data = compare_crop_output, type = 'l', auto = TRUE, grid = TRUE)
```

5. When finished, use git to discard the changes to the stored regression tests if they were altered.

I am not sure why your soil evaporation rate graph looks so different from mine.

eloch216:
I am also quite skeptical of the equation used to calculate the fraction of canopy coverage (`f_c`). This is Equation 76 in FAO-56, where several disclaimers are provided:

> This equation should be used with caution and validated from field observations. K_c_min is the minimum crop coefficient for dry bare soil when transpiration and evaporation from the soil are near baseline (diffusive) levels. K_c_min approximately 0.15 - 0.20 is recommended. The value of K_c_min is an integral part of all K_cb coefficients. K_c_min ordinarily has the same value as the K_cb_ini used for annual crops under nearly bare soil conditions (0.15 - 0.20). Equation 76 assumes that the value for K_cb is largely affected by the fraction of soil surface covered by vegetation. This is a good assumption for most vegetation and conditions. The '1+0.5h' exponent in the equation represents the effect of plant height on shading the soil
surface and in increasing the value for K_cb given a specific value for f_c. The user should limit the difference K_cb - K_c_min to ≥ 0.01 for numerical stability.

It's definitely strange that the DSSAT code (and hence our code) fixes `k_c_min` to 0 even though the reference describing the equation suggests a different value.

It's fairly easy to recalculate `f_c` after running a simulation. When `kcbmax` is at 0.25, `f_c` only reaches ~0.1, which is totally unreasonable for soybean. At maximum LAI, soybean almost completely covers the ground, so `f_c` should be close to 1. With `kcbmax = 1.0`, `f_c` reaches up to ~0.8. This is better but still not right. When `kcbmax` is 1.2, `f_c` reaches a weird plateau at ~0.9. The clipped shape here also seems unrealistic. (See below for figures and the code used to generate them.)

Ultimately I think it would be better to estimate the fraction of canopy coverage by taking a different approach. Currently there are (at least) three different approaches used in different modules.
1. In our multilayer canopy modules, the fraction of direct sunlight passing through the canopy (`canopy_direct_transmission_fraction`) is calculated. This enables an estimate of the fraction of canopy coverage as `1 - canopy_direct_transmission_fraction`.
2. In `surface_albedo()`, a similar approach is used, except that the extinction coefficient for direct light is fixed to 0.75.
3. In `potential_soil_evaporation()`, Equation 76 from FAO-56 is used, with height fixed to 1 m.

(This is another example of the issues discussed here: #243 )

In my opinion, version 1 seems to be the most realistic. It is a more sophisticated approach than 2 since it considers the leaf angle distribution. And version 3 has several self-admitted shortcomings. I would like to take a consistent approach across all parts of BioCro, and I think it would be best to use version 1. But this can be changed in another PR. For now we can just change `kcbmax` to 1.2 based on DeJonge & Thorp (2017).

image

image

image


Code used:

```r
# Load libraries
library(BioCro)
library(lattice)

# Clear workspace
rm(list=ls())

# Run a simulation
soybean2_temp <- within(soybean2, {
  parameters$kcbmax = 0.25 # change this to try different kcbmax
})

res <- with(soybean2_temp, {run_biocro(
  initial_values,
  parameters,
  soybean_weather[['2002']],
  direct_modules,
  differential_modules,
  ode_solver
)})

# Set parameter values needed to calculate f_c
height   <- 1 # Hard-coded in `soil_evaporation2`
K_cb_min <- 0 # Hard-coded in `potential_soil_evaporation`
K_cb_max <- soybean2_temp$parameters$kcbmax
SK_c     <- soybean2_temp$parameters$skc

# Extract columns needed to calculate f_c
LAI <- res$lai

# Calculate f_c as in `potential_soil_evaporation`
K_cb   <- K_cb_min + (K_cb_max - K_cb_min) * (1.0 - exp(-1.0 * SK_c * LAI))
K_cmax <- pmax(1.0, K_cb + 0.05)
f_c    <- ((K_cb - K_cb_min) / (K_cmax - K_cb_min))^(1.0 + 0.5 * height)
f_c    <- pmax(0.0, f_c)

# Plot f_c
xyplot(
  f_c ~ res$fractional_doy,
  type = 'l',
  ylim = c(0, 1),
  xlab = 'Day of year (2002)',
  main = paste('kcbmax =', soybean2_temp$parameters$kcbmax)
)
```

eloch216:
> > I am not sure why your soil evaporation rate graph looks so different from mine.
> 
> This is weird. Running `run_biocro` directly looks good. Something in `BioCro::compare_model_output` altered the model behavior, and it's only for `soil_evaporation_rate`!?

Oh, I think I know what happened... the regression tests only store one value per day. So `BioCro::compare_model_output` also only retains one value per day from the "new" results. I guess a lot of the dynamics of soil evaporation are missed when doing this. I guess I need to be careful when using the regression test method to compare versions!

leavesandlemmas:
> > > I am not sure why your soil evaporation rate graph looks so different from mine.
> > 
> > 
> > This is weird. Running `run_biocro` directly looks good. Something in `BioCro::compare_model_output` altered the model behavior, and it's only for `soil_evaporation_rate`!?
> 
> Oh, I think I know what happened... the regression tests only store one value per day. So `BioCro::compare_model_output` also only retains one value per day from the "new" results. I guess a lot of the dynamics of soil evaporation are missed when doing this. I guess I need to be careful when using the regression test method to compare versions!

Moreover, the regression tests store the nighttime simulation values. For biomass and other integrated quantities, the difference between daytime and nighttime values is small, but it's quite important for things like assimilation evaporation.

With that in mind, we might want to store a different subset of values.  Some possibilities.

1. Store the midday values `hour == 12`
2. Store a random subset of time values
3. Store every `k`-th value, but select a `k` that causes simulations to be sampled through the day.  e.g. every 23rd or 25-value. 

For hourly simulations, saving every 23rd or 25th simulation result produces roughly the same number of saved values, but a simulation result from each hour appears once per 23 days  or once per 25 days.

That doesn't really solve the model comparison. I think it might be worth having some code to compute daily average, daily max, day min, etc.

yufenghe22 (approved):

Issue #235, opened 2025-11-07
Too much carbon fixed under elevated CO2?


yufenghe22:
The current BioCro version on the develop branch predicts too much yield compared to the original [Soybean-BioCro](https://academic.oup.com/insilicoplants/article/4/1/diab032/6453473#329258254:~:text=Relative%20maximum%20biomasses%20(elev.%20%5BCO2%5D/amb.%20%5BCO2%5D)). The current predictions are ~28% increases due to eCO2 at 550 ppm (see bar plot results below), compared to ~12% from the original. Observations are more close to the original prediction, so the current version largely overestimates the effects of eCO2. You can easily check that the peak biomasses of leaf and stem are also much higher than the original ones, indicating the overall carbon fixed are overestimated. There have been so many substantial changes since the original Soybean-BioCro in 2022, such as changes made to leaf energy balance and added respiration modules. I don't think the yield overestimation indicates anything wrong with these changes, but it definitely changed the model sensitivity to eCO2. Some parameters need to be adjusted to lower yield and overall biomasses under eCO2.

In the future, when substantial changes are made to the model that impact yield, in addition to comparison to the existing data under ambient CO2, we should also show results under elevated CO2 to make sure the changes have not altered the sensitivity too much.
<img width="1470" height="940" alt="Image" src="https://github.com/user-attachments/assets/8d8532f4-b558-4f2b-a64a-db2c8fba4e4a" />

@leighmatth

leavesandlemmas:
I think that @justinmcgrath 's idea of fitting the Soybean-BioCro to the elevated CO2 and seeing which parameters change the most might reveal the culprit. It's possible that some parameters which are sensitive to the data might not have a big impact on the yield.

@yufenghe22 Do you have a branch / commit that I could look at for the original Soybean-BioCro? Or is that in the paper? I could look at the differences to see if there's anything that stands out.

yufenghe22:
> I think that [@justinmcgrath](https://github.com/justinmcgrath) 's idea of fitting the Soybean-BioCro to the elevated CO2 and seeing which parameters change the most might reveal the culprit. It's possible that some parameters which are sensitive to the data might not have a big impact on the yield.
> 
> [@yufenghe22](https://github.com/yufenghe22) Do you have a branch / commit that I could look at for the original Soybean-BioCro? Or is that in the paper? I could look at the differences to see if there's anything that stands out.

[...]

PR #273, opened 2026-02-26
More temp responses

eloch216:
This PR is a response to the recent discussion about temperature response functions (https://github.com/biocro/biocro/discussions/272). It adds two new responses (`eyring_response` and `peaked_arrhenius_response`), along with additional documentation that clarifies the differences between the `arrhenius_response`, `eyring_response`, `peaked_arrhenius_response`, and `johnson_eyring_williams_response`.

It doesn't alter the temperature response functions used for any BioCro modules or crop models, so there are no changes to any of the tests.


src/module_library/temperature_response_functions.h
```diff
@@ -72,23 +72,135 @@ inline double Q10_temperature_response(
...
- *  @brief A temperature response function originally defined in Johnson,
- *         Eyring, and Williams (1942), but used more recently to describe the
- *         temperature response of triose phosphate utilization limitations in
- *         Harley et al. (1992), Sharkey et al. (2007), and Yang et al. (2016).
+ *  @brief Calculates a relative reaction rate using the Eyring equation
+ *
+ *  The Eyring equation (Eyring 1935) gives the dependence of the rate constant
+ *  of a chemical reaction on the absolute temperature:
+ *
+ *  > `k = kappa * e^(S / R) * e^(-H / R / T) * (k * T / h)` (1)
```
justinmcgrath: Needs different symbols for rate constant and Boltzmann's constant.

In general I find the extra parentheses add some unnecessary mental parsing.

kappa * k_B * T / h * e^(S / R) * e^(-H / R / T)

is simpler.

src/module_library/temperature_response_functions.h
```diff
@@ -72,23 +72,135 @@ inline double Q10_temperature_response(
...
+/**
+ *  @brief A temperature response function originally defined in Hall (1979),
+ *  often used to represent peaked temperature responses in plant models.
+ *
+ *  The peaked Arrhenius response gives the dependence of the value of a
+ *  parameter on the absolute temperature:
+ *
+ *  > `p = k * exp(-H_a / R / T) / [1 + exp(S / R) * exp(-H_d / R / T)]` (1)
+ *
+ *  Where `p` is the parameter value, `k` is a scaling factor, `H_a` is the
```
justinmcgrath: Maybe don't reuse k this way. k is defined here again so there isn't any ambiguity, but it is a bit odd to reuse it in this context. And commonly the Arrhenius equation uses it in the same way as the Eyring equation described above, as the rate constant, so it's going against a common notation.

src/module_library/temperature_response_functions.h
leavesandlemmas: One can remove even more parentheses by writing

`kappa * k_B * T / h * exp(S/R - H/R/T )`

src/module_library/temperature_response_functions.h
```diff
@@ -72,23 +72,135 @@ inline double Q10_temperature_response(
...
+inline double eyring_response(
+    double gamma,         // K^-1
+    double H,             // J / mol
+    double S,             // J / K / mol
+    double temperature_k  // K
+)
+{
+    using physical_constants::ideal_gas_constant;  // J / k / mol
+
+    return gamma * temperature_k *
```
leavesandlemmas: Wouldn't it be simpler to express this function so that the user doesn't have to ensure gamma is calculated correctly? That seems like a potential source of bugs in client code.

src/module_library/temperature_response_functions.h
```diff
@@ -114,15 +226,91 @@ inline double johnson_eyring_williams_response(
 {
     using physical_constants::ideal_gas_constant;  // J / k / mol
 
-    double const top =
-        temperature_k * arrhenius_exponential(c, Ha, temperature_k);
+    double const top = arrhenius_exponential(c, Ha, temperature_k);
```
leavesandlemmas: Again, wouldn't it be simpler to reformulate so that the user doesn't have to ensure `c` meets some requirement?

src/module_library/temperature_response_functions.h
leavesandlemmas: E.g. 

```cpp
double eyring_response(T, Tref, H, S) {
     double gamma = 1/(Tref* arrhenius_exponential(S/ideal_gas_constant, H, Tref));
     return gamma * T * arrhenius_exponential(S/ideal_gas_constant, H, T);
}  
```

and if we are concerned about the needless duplicate calculation then we could use a function object to save the reference temperature:

```cpp
struct eyring_response {
    eyring_response(double Tref, double S, double H) :
    S{S},
    H{H},
    gamma{1/(Tref * arrhenius_exponential(S / ideal_gas_constant, H, Tref))} {}

    double operator()(double T)
    {
        return gamma * T * arrhenius_exponential(S / ideal_gas_constant, H, T);
    }

   private:
    double gamma;
    double S;
    double H;
};
```

yufenghe22 (approved):

src/module_library/temperature_response_functions.h
yufenghe22: There's probably some valid ranges of $c$ regarding each type of equation. A warning message can be sent if the user uses an out of range $c$ with the selected equation.

leavesandlemmas: I don't think there's any issues with invalid `c`, but if `c` is supposed to be the response value at some reference temperature, there's no guarantee that it is. 

I think we have had this issue before when varying parameters, e.g. issue #235 . Changing `c` is equivalent to changing the base response at the reference temperature and to changing the reference temperature without changing the response. 

In my opinion, it's simpler to just compute `c` from a reference value and a reference temperature, since those parameters are more meaningful than `c`. The units are simpler to track too as the value of `c` changes if one changes in the units of the reference value.

src/module_library/temperature_response_functions.h
justinmcgrath: That's fine too, but my point is about having parentheses that don't do anything, not about expressing equations in a form with the fewest parentheses.

eloch216: image

I wrote it that way to follow how it's written in the original publication. I guess I dropped the subscript `i` from the rate and forgot there was already another `k` involved. I can use `k_b` for Boltzmann's constant since that is more standard. I can also remove some parentheses.

src/module_library/temperature_response_functions.h
eloch216: image

I used `k` here to follow how the equation is written in the original publication. But I can change it to something else instead.

src/module_library/temperature_response_functions.h
eloch216: > Wouldn't it be simpler to express this function so that the user doesn't have to ensure gamma is calculated correctly? That seems like a potential source of bugs in client code.

I don't think this is the proper place to do that. Although it is common for `gamma` to be chosen such that the return value is 1 at the reference temperature, this is not actually a requirement for `gamma`. In principle, `gamma` could also be chosen such that the returned value is the actual rate at `Tref`. In my opinion, `c3_temperature_response()` would be a better place to calculate values of `c` and `gamma`. I don't want to do that right now though, since Sruthi is also making changes to that function at the moment.

I find function objects to be confusing in the context of modeling. They take what could be some very simple code that would be intelligible even to someone without much coding experience, and chop it up into various class initializers, members, and methods.

eloch216: I've reflected on this a bit more, and I guess it does seem weird to expect `c` or `gamma` to be calculated somewhere else. I suppose right now is as good a time as any to change the way this works. I still don't like function objects, though.

eloch216:
I made a few small changes to notation as suggested by Justin, and I also modified the the Arrhenius, Eyring, peaked Arrhenius, and Johnson-Eyring-Williams reposes so they return values normalized to a reference temperature. This eliminated a few parameters from the soybean and willow models, which is nice. I had to update a few tests, but the changes were very small. For example, the value of `Tp_norm` at 25 degrees C changed from 0.99999744773899 to 1.

leavesandlemmas (approved):
I think expressing the temperature responses in terms of a reference value and temperature is a good change for the long run. 

I am guessing making this change removed a couple of parameters because we were specifying both the c parameter and a reference value? Even though specifying the c parameter is equivalent to picking a reference value at some reference temperature.

leavesandlemmas:
> but the changes were very small. For example, the value of `Tp_norm` at 25 degrees C changed from 0.99999744773899 to 1.

Probably due to rounding of the C parameter I guess. Or possibly using 298 K as the reference temperature rather than 298.15 K

eloch216:
> I am guessing making this change removed a couple of parameters because we were specifying both the c parameter and a reference value?

Yep, that's exactly what happened.

> Probably due to rounding of the C parameter I guess. Or possibly using 298 K as the reference temperature rather than 298.15 K

For TPU it's definitely due to rounding. For some of the others, I believe we used the `c` values from published papers, and I can't say for sure if those papers assumed 298 or 298.15 for the reference temperature.

eloch216:
There is a spurious test failure when checking the package on MacOS-latest using R version 4.1.0. The check fails because the vignettes cannot be built:

```
Error in loadVignetteBuilder(pkgdir, TRUE) : 
  vignette builder 'knitr' not found
Execution halted
```

This is difficult to understand because the "Set Up R Dependencies" step reports that `knitr` was successfully installed.

Looking at the logs of our GitHub actions, this has been happening since January 19 (https://github.com/biocro/biocro/actions/runs/21147925198). It did not happen on January 15 (https://github.com/biocro/biocro/actions/runs/21050437782).

I will see if I can figure anything out about this.

PR #274, opened 2026-03-03
solve gs and Tleaf simultaneously

yufenghe22:
Without solving $g_s$ and $T_{leaf}$ simultaneously, we overestimate Tleaf by not using the updated $g_s$. There's also a clear change on the final yield. Here I implemented a simple fix-point method to solve  $g_s$ and $T_{leaf}$. The convergence is good. All layers can reach convergence within 5 iterations as I created a variable to track the number of steps.
image
image


yufenghe22:
@leavesandlemmas I wonder if this is duplicate with what you might be working on

leavesandlemmas:
> @leavesandlemmas I wonder if this is duplicate with what you might be working on

Yes and no. My branch for this objective is on https://github.com/biocro/biocro/tree/simultaneous-heat-ci-equation  

I like that your version would allow us to test whether or not we have problems with convergence and only requires minimal changes to the code.  Do you get convergence? How many iterations does it take?

My version is still very much a work-in-progress, and it doesn't work yet. I think the fixed point iteration probably fails to converge a significant number of time-points, and even when it does converge, it can converge slowly.  So long-term it seems better to me to use a multi-dimension version of the secant method (AKA broyden's method  ) or other multidimensional methods like the Levenberg-Marquadt algorithm.  There are two barriers to implementation of those methods. 

1. To implement any multi-dimensional root solving algorithm, we need to be able to solve linear systems of equations. Frankly I don't understand boost's BLAS and BLAS is also limited. So I wrote my implementation of the LU decomposition algorithm with partial pivoting, and the code for specifying vectors / matrices and what not.  I have an implementation of Broyden's method which should work alright, especially for low-dimensional problems. The LM algorithm is likely more robust, but it's a bit more sophisicated.
2. The photosynthesis code needs to be reworked so that it interfaces with the multidimensional root-solving library, 

I haven't worked on this since like Christmas. Basically, I have concentrated on nitrogen transport and some lab work, and have neglected improvements to BioCro. Similarly, I was working on refactoring the canopy modules so that it's easy to try different numerical integration methods. And I also stalled out on that.

src/module_library/c3_leaf_photosynthesis.cpp
```diff
@@ -117,31 +118,50 @@ void c3_leaf_photosynthesis::do_operation() const
...
-        initial_stomatal_conductance,
-        windspeed);
+    photosynthesis_outputs photo; 
+    energy_balance_outputs et;
+    // Initial guess
+    double current_gs = initial_stomatal_conductance;
 
-    double const leaf_temperature = ambient_temperature + et.Deltat;  // degrees C
+    // 1. Set convergence criteria
+    const int max_iterations = 10;
```
leavesandlemmas: I would test with a higher max_iterations. Just to diagnose if convergence is an issue.

src/module_library/c3_leaf_photosynthesis.cpp
```diff
@@ -117,31 +118,50 @@ void c3_leaf_photosynthesis::do_operation() const
...
+              et.gbw_molecular);
+
+      // 4. Check for convergence
+      if (std::abs(photo.Gs - current_gs) < tolerance) {
+          break;
+      }
+      // 5. Update guess for next iteration
+      current_gs = photo.Gs;
+      num_of_it += 1;
+    }
```
leavesandlemmas, line 154: Consider added a check for non-convergence here, then add a print statement or throw an exception to see how often we don't get convergence during typical simulations

eloch216:
@leavesandlemmas Did you include a fixed-point solver in your collection of 1D solvers? If so, it would probably be helpful to use that here instead of "manually" making a fixed-point loop. That would also make it easier to test whether another solver could be faster.

yufenghe22:
> Do you get convergence? How many iterations does it take?

Yes. I ran 2002. All layers can reach convergence within 5 iterations. I added a variable called `numit` to track this. I think solving for gs and Tleaf here is easier to reach convergence because $g_s$ has already gone through the FvCB convengency and the energy balance is a rather stable step. 
image

src/module_library/c3_leaf_photosynthesis.cpp
leavesandlemmas: Okay, I did some testing with max_iteration = 50 and tolerance = 0.001  and the fixed point iteration performed remarkably well. Better than I thought it would.

leavesandlemmas:
> @leavesandlemmas Did you include a fixed-point solver in your collection of 1D solvers? If so, it would probably be helpful to use that here instead of "manually" making a fixed-point loop. That would also make it easier to test whether another solver could be faster.

Yes, there's a fixed point method. It's the only one that doesn't check for a zero, essentially tests `f(x) == x` so it should be possible to rewrite this code to use the fixed_point solver.

leavesandlemmas:
Okay, I did some testing with max_iteration = 50 and tolerance = 0.001 and the fixed point iteration performed remarkably well. Better than I thought it would. 

I was able to get failed convergence, but only for very low RH. Convergence slows for `StomataWS ~ 0.5`  and `temp> 35` or low `Catm` (like 100-200). More or less the same conditions that cause pathology for the Ci equation. 

So I am surprised how robust it is. However, the efficiency is harder to judge. While it generally takes less than 10 iterations to achieve convergence within tolerance, `c3photoC` and `leaf_energy_balance` both have equation solving which takes around 5-20 iterations. Presumably, a multi-dimensional solver would take fewer iterations () but it's possible that the improvement would be negligible.   


It's possible to expand the coupled system into several equations (5-7 depending) and brodyen's method solves n-dimensional linear systems in 2n iterations, and presumably nonlinear-but-almost-linear equations take a few more iterations. So we should expect 15-30 iterations to solve the coupled Ci, heat balance, and stomatal model equations.

leavesandlemmas:
It should be possible to use the secant method or dekker's method to solve the fixed point loop. I am not sure this would be much faster however.  

> It's possible to expand the coupled system into several equations (5-7 depending) and brodyen's method solves n-dimensional linear systems in 2n iterations, and presumably nonlinear-but-almost-linear equations take a few more iterations. So we should expect 15-30 iterations to solve the coupled Ci, heat balance, and stomatal model equations.

To explain what I mean here. The leaf photosynthesis model is a set of nonlinear equations for heat balance, Ci balance, stomatal response.  For example, the Ci balance equation is:

$$
g_s (C_s - C_i)- F(C_i, T_L) =0
$$ 

where $F$ is the rate of carboxylation less the rate of respiration, suppressing all parameters that aren't unknowns to be solved for.  Temperature/ heat balance:

$$
R - B(T_L) - H(T_L) -  E(T_L, g_s) = 0
$$

and stomatal conductance

$$
b_0 + b_1 \max\left( 0,  \frac{F(C_i, T_L) h_s}{C_s} \right) - g_s = 0
$$

plus an equation for $C_s$ and $h_s$ which I won't bother to write. The number of equations can be reduced by variable substitution; e.g., $g_s$ can be computed from $C_i$ and $T_L$   as can $h_s$ although the expression for the latter is complicated.  But it isn't possible to reduce the number of equations to 1, the minimum is two because it isn't possible to isolate $T_L$ as a function of $C_i$, or $T_L$ as a function of $C_i$. These equations have to be solved numerically as 1D equations. 

It seems that BioCro's code was originally based on the Nikolov paper where fixed-point iteration by variable substitution was used to effectively reduce the number of equations from 7 to 2, with fixed-point iteration.  While there's nothing wrong with fixed-point iteration, most numerical methods are fixed-point iteration but with techniques to ensure robustness and to speed convergence, this process is potentially slow and non-convergent. 

The process of reducing the number of equations to solve from 5 to 2 (or 7 to 2) makes the smaller dimension equations more nonlinear than they were. Like the Ci balance equation is easier to solve numerically for $C_i$ when $g_s$ is independent constant, but harder when $g_s$ is a function of $C_i$.  Newton's method (and Broyden's method) work by approximating the system of nonlinear equations with a set of linear equations, then iterating that. And the more "non-linear" (i.e. the smaller the region where the linear approximation is good) the slower or less robust these methods are. 

So my intuition is that the variable substitution makes the numerical methods perform worse, but I haven't been able to confirm or reject this idea.

yufenghe22:
> So my intuition is that the variable substitution makes the numerical methods perform worse, but I haven't been able to confirm or reject this idea.

I think that's true. Fewer equations $\neq$ easier convergence. The 2x2 Jacobians become much more complicated, while the 7x7 may be easier to solve.

yufenghe22:
@eloch216 @leavesandlemmas I also apply the solver to the `c3CanAC` to make sure the tests pass.

yufenghe22:
All tests passed on my laptop. Is the failed check related to `knitr`? @eloch216

eloch216:
> All tests passed on my laptop. Is the failed check related to `knitr`? @eloch216

Yes, it's not a real test failure. Once I merge PR #273, we can update this branch from `develop`, and then the `knitr` issue will be fixed.

eloch216:
I still haven't been able to take a close look at this, but it seems very useful and a positive development. My main comment right now is that we should probably do the same thing for `c4_leaf_phototosynthesis` and `CanAC`.

eloch216:
I can update this from `develop` and address the merge conflicts

eloch216:
@yufenghe22 Thanks for updating the other tests... I started doing that, and then apparently forgot what I was doing, and committed before I was done 😂

yufenghe22:
All tests should be good now. I'll merge by the end of day if no further comments.

eloch216:
> All tests should be good now. I'll merge by the end of day if no further comments.

Please see this comment: https://github.com/biocro/biocro/pull/274#issuecomment-3998936452

leavesandlemmas:
@yufenghe22  If you would prefer, I might have time later today to implement the same thing for the C4 modules.

yufenghe22:
> @yufenghe22 If you would prefer, I might have time later today to implement the same thing for the C4 modules.

Thanks! Let me try it first.

yufenghe22:
It's nice the C4 canopy modules are very much like the C3. It's pretty easy to make the changes. @leavesandlemmas @eloch216

yufenghe22:
There's almost no difference for Miscanthus' biomasses. Small difference is seen in the ET:
image

eloch216:
I am taking a look at some other outputs to see how things have changed. One thing I noticed is that Scott used tighter tolerances when implementing the `root_finding::fixed_point solver` approach, and these seem to be important. The figure below shows one of the plots from the "Ball-Berry" vignette. There are some discontinuities when using the larger tolerance. Just bringing this up since I think it shows the current tolerances are good, and we shouldn't make them looser.

Picture1

src/module_library/c3_leaf_photosynthesis.cpp
```diff
@@ -68,7 +69,9 @@ string_vector c3_leaf_photosynthesis::get_outputs()
         "RH_canopy",         // dimensionless
         "RL",                // micromol / m^2 / s
         "Rp",                // micromol / m^2 / s
-        "TransR"             // mmol / m^2 / s
+        "TransR",            // mmol / m^2 / s
+        "numit",             // not a physical quantity
```
eloch216: We now have several modules that return a number of loop iterations. Most of them use the name `iterations` (`BioCro:c3_assimilation`, `BioCro:c4_assimilation`, and `BioCro:leaf_evapotranspiration`). `BioCro:c3_leaf_photosynthesis` now uses `numit`, and `BioCro:c4_leaf_photosynthesis` does not return the number of iterations.

I think it would be nice if all modules that use loops return the number of iterations. We should standardize the names, and also make sure each one is unique to avoid any confusion. Maybe we could do something like `iterations_modulename`? E.g. this one would become `iterations_c3_leaf_photosynthesis`. I'm open to other suggestions.

src/module_library/c3_leaf_photosynthesis.cpp
```diff
@@ -68,7 +69,9 @@ string_vector c3_leaf_photosynthesis::get_outputs()
...
+        "residual"           // mol / m^2 / s
```
eloch216: We now have several modules that return the residual at the end of a loop. Most of them use a name formatted like `q_check`, where `q` is the name of the quantity whose value is used to determine convergence. Specifically, the `BioCro:c3_assimilation` and `BioCro:c4_assimilation` modules return `Assim_check`, and the `BioCro:leaf_evapotranspiration` module returns `leaf_temp_check`. `BioCro:c3_leaf_photosynthesis` now uses `residual`, and `BioCro:c4_leaf_photosynthesis` does not return the residual.

I think we should return these residuals when possible and also standardize their names. Since the `q_check` format is established, we could continue using it. In that case, the C3/C4 leaf photosynthesis modules should both return `Gs_check`.

Alternatively, we could use `residual` rather than `check`, since that may be more informative. What do you think? I'm open to other suggestions.

eloch216:
@yufenghe22 Please make sure to run `clang-format`. I used it when addressing the merge conflicts. But I think it has not been applied to the C4 files.

eloch216:
Other than a few small "style" suggestions above, I think this is just about ready to go. It's nice to see that there aren't any huge changes to the simulation results. So our old approach was not super far off from the "true" solution.

One remaining question is whether there are multiple potential solutions. Right now it's not very easy to check. One option could be to make a module that calculates the value of `func`. Then we could use this module to check for multiple fixed points. I think the most straightforward way to do this would be to implement `func` as a free function that could be called by `c3_leaf_photosynthesis`, `c3CanAC`, and the new module. (And similar for a C4 version.)

This is the same approach used for the leaf energy balance calculations. There is a free function called `check_leaf_temp()` defined in `leaf_energy_balance.cpp` that is passed to a solver in the `leaf_energy_balance()` function via a lambda. It is also directly available from R via the `BioCro:leaf_evapotranspiration_check` module.

Here the free function could be called something like `check_c3_gs()`, and likewise for the associated module. It's not complicated, but it is a bit time consuming to do this.

Checking for multiple solutions would be interesting but it doesn't necessarily need to happen before merging this PR, though.

src/module_library/c3_leaf_photosynthesis.cpp
leavesandlemmas: > Maybe we could do something like `iterations_modulename`?

sounds good to me

src/module_library/c3_leaf_photosynthesis.cpp
leavesandlemmas: I would vote for `check_q` for consistency with `iterations_modulename`.  We could also just do `residual_modulename` for the same reason. That ought to work unless/until we start using multidimensional root solving. 

I think residual is better, even if I originally came up with `check`. I picked residual partly because that's what the quantity is called in numerical analysis (if named at all) and partly because I think it better suggests that `residual = 0` is correct whereas `check` sounds a bit like the result of a boolean test.  But that's very much vibes based. I think the name isn't that important (but maybe we need a comment to explain what it does? or better documentation for it?)

yufenghe22: residual sounds better to me.

src/module_library/c3_leaf_photosynthesis.cpp
yufenghe22: I agree the name change will be more informative. One thing to think about is that all these solver related quantities are often not used, but they occupy a lot of output space (basically multiply each variable by 20). It would be useful to have an option to turn off these diagnostic outputs.

yufenghe22:
> One remaining question is whether there are multiple potential solutions. Right now it's not very easy to check. One option could be to make a module that calculates the value of `func`.

How does the `func` help find a different solution? Is it via multiple initial values or estimation of Jacobian?

leavesandlemmas:
> > One remaining question is whether there are multiple potential solutions. Right now it's not very easy to check. One option could be to make a module that calculates the value of `func`.
> 
> How does the `func` help find a different solution? Is it via multiple initial values or estimation of Jacobian?

For a 1D equation, we can plot the graph to see if multiple solutions are probable.

leavesandlemmas:
> is is the same approach used for the leaf energy balance calculations. There is a free function called `check_leaf_temp()` defined in `leaf_energy_balance.cpp` that is passed to a solver in the `leaf_energy_balance()` function via a lambda. It is also directly available from R via the `BioCro:leaf_evapotranspiration_check` module.

I wish we had a more ergonomic way to do this kind of testing. I think we have a large number of modules which more or less wrap the same code and it becomes confusing for new modelers/developers/users. But I am not sure there's much we can do without other more significant changes to the code base.  

Similarly, I think free functions with more than 3-5 parameters is not particularly ergonomic since it makes fixing bugs like the one below more difficult (although TBH the upgrade to C++17 has made the compiler messages more informative than they were before).

```
* installing to library 'C:/Users/Scott/AppData/Local/R/win-library/4.3'
* installing *source* package 'BioCro' ...
** using staged installation
** libs
using C++ compiler: 'G__~1.EXE (GCC) 12.3.0'
using C++17
g++  -std=gnu++17 -MT module_library/c3_canopy.o -MMD -MP -MF .deps/module_library/c3_canopy.Td -I"C:/PROGRA~1/R/R-43~1.3/include" -DNDEBUG -I../src/inc -DR_NO_REMAP    -I"C:/rtools43/x86_64-w64-mingw32.static.posix/include"     -O3 -march=native -mtune=native -c -o module_library/c3_canopy.o module_library/c3_canopy.cpp
module_library/c3_canopy.cpp: In member function 'virtual void standardBML::c3_canopy::do_operation() const':
module_library/c3_canopy.cpp:107:61: error: too few arguments to function 'canopy_photosynthesis_outputs c3CanAC(c3_temperature_response_parameters, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, double, int, int)'
  107 |     const canopy_photosynthesis_outputs can_result = c3CanAC(
      |                                                      ~~~~~~~^
  108 |         tr_param,
      |         ~~~~~~~~~
  109 |         absorbed_longwave,
      |         ~~~~~~~~~~~~~~~~~~
  110 |         temp,
      |         ~~~~~
  111 |         atmospheric_pressure,
      |         ~~~~~~~~~~~~~~~~~~~~~
  112 |         atmospheric_scattering,
      |         ~~~~~~~~~~~~~~~~~~~~~~~
  113 |         atmospheric_transmittance,
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~
  114 |         b0,
      |         ~~~
  115 |         b1,
      |         ~~~
  116 |         beta_PSII,
      |         ~~~~~~~~~~
  117 |         Catm,
      |         ~~~~~
  118 |         chil,
      |         ~~~~~
  119 |         cosine_zenith_angle,
      |         ~~~~~~~~~~~~~~~~~~~~
  120 |         electrons_per_carboxylation,
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  121 |         electrons_per_oxygenation,
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~
  122 |         gbw_canopy,
      |         ~~~~~~~~~~~
  123 |         growth_respiration_fraction,
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  124 |         Gs_min,
      |         ~~~~~~~
  125 |         Gstar_at_25,
      |         ~~~~~~~~~~~~
  126 |         heightf,
      |         ~~~~~~~~
  127 |         Jmax_at_25,
      |         ~~~~~~~~~~~
  128 |         k_diffuse,
      |         ~~~~~~~~~~
  129 |         Kc_at_25,
      |         ~~~~~~~~~
  130 |         Ko_at_25,
      |         ~~~~~~~~~
  131 |         kpLN,
      |         ~~~~~
  132 |         lai,
      |         ~~~~
  133 |         leaf_reflectance_nir,
      |         ~~~~~~~~~~~~~~~~~~~~~
  134 |         leaf_reflectance_par,
      |         ~~~~~~~~~~~~~~~~~~~~~
  135 |         leaf_transmittance_nir,
      |         ~~~~~~~~~~~~~~~~~~~~~~~
  136 |         leaf_transmittance_par,
      |         ~~~~~~~~~~~~~~~~~~~~~~~
  137 |         leafwidth,
      |         ~~~~~~~~~~
  138 |         LeafN,
      |         ~~~~~~
  139 |         lnb0,
      |         ~~~~~
  140 |         lnb1,
      |         ~~~~~
  141 |         O2,
      |         ~~~
  142 |         par_energy_content,
      |         ~~~~~~~~~~~~~~~~~~~
  143 |         par_energy_fraction,
      |         ~~~~~~~~~~~~~~~~~~~~
  144 |         rh,
      |         ~~~
  145 |         RL_at_25,
      |         ~~~~~~~~~
  146 |         StomataWS,
      |         ~~~~~~~~~~
  147 |         Tp_at_25,
      |         ~~~~~~~~~
  148 |         Vcmax_at_25,
      |         ~~~~~~~~~~~~
  149 |         windspeed,
      |         ~~~~~~~~~~
  150 |         windspeed_height,
      |         ~~~~~~~~~~~~~~~~~
  151 |         lnfun,
      |         ~~~~~~
  152 |         nlayers);
      |         ~~~~~~~~
```

src/module_library/c3_leaf_photosynthesis.cpp
yufenghe22: Does the list of vars look good? Open to suggestions.
- residual_c3Assim
- residual_c4Assim
- iter_c3Gs, residual_c3Gs
- iter_LeafE, residual_LeafE
- iter_c4Gs, residual_c4Gs

leavesandlemmas (approved):
Looks good to me. 

I initially reduced the tolerance to see if it would cause convergence errors, so I might check whether reducing the tolerance a lot (e.g., `~ 1e-10`) causes convergence problems.  But I suspect it won't for the conditions we test for, although convergence might be quite slow in edge cases. 

However, no need to wait for me before merging, so feel free to merge whenever you're done

src/module_library/c3_leaf_photosynthesis.cpp
eloch216: Looks pretty good. I guess you're just indicating C3 or C4 instead of the full module name? At some point this might become ambiguous, but I think it should work for now.

I would suggest putting an underscore between C3/C4 and the quantity name to clarify that they are different parts. So `residual_c3_Assim` instead of `residual_c3Assim`.

I also have a slight preference for `iterations` rather than `iter`. Most of the time, I don't think there is much benefit to abbreviations. I know we do have many abbreviations in BioCro, such as "Assim." But quite a lot of these are relics from BioCro's early days that would be tedious to change now.

For the leaf energy balance outputs, I think `LeafE` is a bit misleading because the transpiration rate is not checked during the loop; it checks leaf temperature. Or maybe `LeafE` was intended as an abbreviation of the module name? If that's the case, then the name of the checked quantity is missing. I think something like `iterations_energy_balance_Tleaf` and `residual_energy_balance_Tleaf` would be clearer, and more consistent with the other names you suggested. Consistency is important here.

eloch216:
> @yufenghe22 Please make sure to run `clang-format`. I used it when addressing the merge conflicts. But I think it has not been applied to the C4 files.

The changes made by `clang-format` look strange to me. Specifically, it looks like `clang-format` shortened some lines in some odd ways that don't necessarily improve readability. Are you not using the style specifications in `biocro/.clang-format`? Our style file doesn't enforce a column width, so `clang-format` shouldn't be splitting things across lines like that.

yufenghe22:
> Specifically, it looks like `clang-format` shortened some lines in some odd ways that don't necessarily improve readability.

Ah. I see. It looked strange to me too. I did it in the `src/module_library` folder.

eloch216:
> > Specifically, it looks like `clang-format` shortened some lines in some odd ways that don't necessarily improve readability.
> 
> Ah. I see. It looked strange to me too. I did it in the `src/module_library` folder.

In principle, if you call `clang-format` using `clang-format -i -style=file <file_to_format>`, it should be able to find our `.clang-format` style file because it is in a parent directory of `src/module_library`. It's hard to say why this might not be working for you. But it's possible to manually specify the style file when calling `clang-format`: `clang-format -i -style=file:<path/to/your/style_file> <file_to_format>`

yufenghe22:
I've changed names to the following files:
- c3_leaf_photosynthesis.cpp
- c4_leaf_photosynthesis.cpp
- c3_assimilation.h
- c4_assimilation.h
- leaf_evapotranspiration.h

Both `c3_leaf_photosynthesis` and `c4_leaf_photosynthesis` output the residuals and iterations for `Gs`. I think it makes sense because this is the last call to the solver with everything coupled. These will be outputs if someone runs the full BioCro model.

The residuals and iterations for `Tleaf` are hidden in the `energy_balance_outputs` structure. They are output in the `leaf_evapotranspiration` module.

The residuals and iterations for `Assim` are hidden in the `photosynthesis_outputs` structure. They are output in the `c3_assimilation` and `c4_assimilation` modules.

eloch216:
Thanks, this looks good. There is one other small change I would like to make; I can just do it myself, and then this can be merged. (I just want to keep the outputs from `leaf_evapotranspiration_check` consistent with `leaf_evapotranspiration`.)

eloch216:
I am done making changes here and this can be merged after the GitHub checks are done running.

When running the tests locally, I noticed some warning messages about the module test cases. These are related to new outputs. For example:

```
Warning ('test.Modules.R:5:5'): All modules are functioning correctly
Module `BioCro:ten_layer_c3_canopy` test case `automatically-generated test case`: unexpected outputs were found: shaded_iteration_C3_Gs_layer_0, shaded_iteration_C3_Gs_layer_1, shaded_iteration_C3_Gs_layer_2, shaded_iteration_C3_Gs_layer_3, shaded_iteration_C3_Gs_layer_4, shaded_iteration_C3_Gs_layer_5, shaded_iteration_C3_Gs_layer_6, shaded_iteration_C3_Gs_layer_7, shaded_iteration_C3_Gs_layer_8, shaded_iteration_C3_Gs_layer_9, shaded_residual_C3_Gs_layer_0, shaded_residual_C3_Gs_layer_1, shaded_residual_C3_Gs_layer_2, shaded_residual_C3_Gs_layer_3, shaded_residual_C3_Gs_layer_4, shaded_residual_C3_Gs_layer_5, shaded_residual_C3_Gs_layer_6, shaded_residual_C3_Gs_layer_7, shaded_residual_C3_Gs_layer_8, shaded_residual_C3_Gs_layer_9, sunlit_iteration_C3_Gs_layer_0, sunlit_iteration_C3_Gs_layer_1, sunlit_iteration_C3_Gs_layer_2, sunlit_iteration_C3_Gs_layer_3, sunlit_iteration_C3_Gs_layer_4, sunlit_iteration_C3_Gs_layer_5, sunlit_iteration_C3_Gs_layer_6, sunlit_iteration_C3_Gs_layer_7, sunlit_iteration_C3_Gs_layer_8, sunlit_iteration_C3_Gs_layer_9, sunlit_residual_C3_Gs_layer_0, sunlit_residual_C3_Gs_layer_1, sunlit_residual_C3_Gs_layer_2, sunlit_residual_C3_Gs_layer_3, sunlit_residual_C3_Gs_layer_4, sunlit_residual_C3_Gs_layer_5, sunlit_residual_C3_Gs_layer_6, sunlit_residual_C3_Gs_layer_7, sunlit_residual_C3_Gs_layer_8, sunlit_residual_C3_Gs_layer_9
```

These are easy to fix using `update_csv_cases` so I also did that.

PR #275, opened 2026-03-03
Change oldest R version for MacOS testing

eloch216:
The CRAN MacOS binary repository (https://cran.r-project.org/bin/macosx/) no longer offers v4.2.0, and the earliest version is v4.2.3. Maybe changing to this version will solve the problem.

This is pretty much a wild guess, but we were able to solve another testing problem in the past by changing to v4.2.0. So maybe it will help again.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -59,7 +59,7 @@ on:
           - R 4.1.0 on Ubuntu
           - R release version on Ubuntu
           - R devel version on Ubuntu
-          - R 4.2.0 on macOS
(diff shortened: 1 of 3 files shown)


eloch216:
It seems that the spurious `knitr` issue does not occur when using R version 4.3.0. See this comment for more details: https://github.com/biocro/biocro/pull/273#issuecomment-3992690602

I assume that the BioCro package likely works fine for earlier versions of R when using Mac, but it seems the online testing setup is limiting us to only check 4.3.0. I think this is okay, considering there is no real alternative.

leavesandlemmas (approved):

PR #278, opened 2026-03-06
New version name

eloch216:
I have forgotten to change the development version number following recent changes. This is just a small change moving to version `3.3.1-0`, since some behavior has been altered.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: BioCro
-Version: 3.3.1
+Version: 3.3.1-0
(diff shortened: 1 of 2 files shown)


eloch216:
This is pretty straightforward so I will just merge when the checks are done, without waiting for explicit approval.

eloch216:
There is a check failure that occurred when building the package documentation. It's hard to believe that changing the package version number and adding a few lines to `NEWS.md` could really cause an issue like this. I will try re-running the job.

PR #280, opened 2026-03-09
Fix root uptake

yufenghe22:
- Add a dynamic root extension with DVI. 
- Add a simple exponential decay function for root distribution.
- Fix some other minor issues between hourly-to-daily conversions from DSSAT.

The root extension and distribution are similar to the idea in the old BioCro two-layer soil profile, where a Poisson function was used. I find it strange to convert root biomass to root depth in that function. Here I simply use a segment function (by DVI) to allow root to access deeper soil water as time passes. This is based on my analysis of the `gs` and soil water data from Gray et al 2016.
With these changes, we can simulate different depths of soil water content much better than before.


eloch216:
It looks like the package cannot be compiled on this branch in its current form because `soil_water_dynamic_rooting.h` is missing

eloch216:
There are some changes here that aren't explained and don't seem to be directly related to the dynamic rooting:
- Why was the initial soil water content changed from 0.32 to 0.4 for soil layers 2 through 6?
- Why was soil layer 1 changed from type 1 ("loamy sand") to type 2 ("sandy loam")?
- Why was soil layer 2 changed from type 4 ("silt loam") to type 3 ("loam")?
- Why were the hardcoded `soil_evaporation_alpha` and `evap_limit` parameter values changed from the "Houston black clay" to "Yolo loam" values in the `soil_evaporation2` module?
- Why were the units of `Ks` changed from `J s m^-3` to `kg s m-3` in `soil_type_selector`?
- Why were some parameter values for the "sandy loam," "loam," and "silt loam" soil types changed in `soil_type_selector`?

yufenghe22:
> There are some changes here that aren't explained and don't seem to be directly related to the dynamic rooting:

Sorry for the confusion. These changes are all related to the calibration I made against observed soil water data. Some of these are minor compared to the root uptake one in my mind. Let me clarify.

> Why was the initial soil water content changed from 0.32 to 0.4 for soil layers 2 through 6?

Because the observed mid-depth layers are close to 0.4 for their initial values. The deep layers are over 0.4. Starting with 0.32 is underestimating. However, the observed data do not have the same start DOY as our growing seasons, so it's more of a guess. The assumption is that the soil are close to saturation at the beginning, which is very reasonable for middle to deep layers (here I assumed 2 to 6 layers).

> Why was soil layer 1 changed from type 1 ("loamy sand") to type 2 ("sandy loam")?
> Why was soil layer 2 changed from type 4 ("silt loam") to type 3 ("loam")?
> Why were some parameter values for the "sandy loam," "loam," and "silt loam" soil types changed in soil_type_selector?

Illinois' soil is some type of loam. I don't think it makes sense to call them "sandy". More importantly, this type of lookup table to get the soil parameters is very crude. Actual soil has certain fractions of sand, silt and clay. When I compare the model results with observations, I found systematic over-estimation of high flow and under-estimation of low low. The low and high points we find in the observations can be used as a calibration criteria for changing the `wiltp` and `fieldc` parameters as what I did.

> Why were the hardcoded soil_evaporation_alpha and evap_limit parameter values changed from the "Houston black clay" to "Yolo loam" values in the soil_evaporation2 module?

Same as above. I think IL soil is closer to loamy soil. Also, you can see the loamy soil has higher evaporation limit, which I found better to match observed soil water at the surface.

> Why were the units of Ks changed from J s m^-3 to kg s m-3 in soil_type_selector?

It was a typo. See Table 9.1 on page 130 of Campbell and Norman.

yufenghe22:
We could use the Poisson function again for root distribution. But it's not clear to me why there's a constant conversion from root biomass to root depth in that function. Here I simply use the segment function and the processes are clearer to me. Until a more mechanistic model of the root is introduced, I consider the two equal in representation.

eloch216:
> > There are some changes here that aren't explained and don't seem to be directly related to the dynamic rooting:
> 
> Sorry for the confusion. These changes are all related to the calibration I made against observed soil water data. Some of these are minor compared to the root uptake one in my mind. Let me clarify.

Thanks for the explanation, this all makes sense. I guess there will be a lot of things to change when moving to a new location. At some point it would be nice to stop hard-coding the "Houston black clay" / "Yolo loam" values so they are easier to modify. But that doesn't need to happen right now.

> We could use the Poisson function again for root distribution. But it's not clear to me why there's a constant conversion from root biomass to root depth in that function. Here I simply use the segment function and the processes are clearer to me. Until a more mechanistic model of the root is introduced, I consider the two equal in representation.

It also seems strange to me that root depth would _only_ depend on DVI. With this approach, it's possible to have a weird scenario where a plant with zero root mass continues to reach deeper soil layers as it ages. Or a less extreme scenario where the root stops growing before DVI reaches 1.0, but nevertheless continues to reach deeper soil layers for DVI above 1.0. But sorting out a better way will take a long time, so this seems fine for now.

eloch216:
These changes have altered the biomass values quite a lot. Is this an expected result? (Just curious.)

image

eloch216:
I think there are a few remaining changes that need to be made here:

1. The tests are not passing. The stored `soybean2` regression data needs to be updated, a test case file for the `BioCro:soil_water_dynamic_rooting` needs to be added, and the test case file for `BioCro:soil_water_uptake` needs to be updated.

2. There are some new additions to the soil water code, but the source for these new equations is not documented. So it's not clear if these were part of the DSSAT code that was accidentally left out, if they are based on a (different) published model, or if they are new additions not based on anything else. A big challenge with understanding these new soil water modules has been tracking down sources, so I don't want to make this problem worse. I will point out a few specific cases where I think more explanation is needed.

src/module_library/multilayer_soil_profile_avg.h
```diff
@@ -274,6 +274,8 @@ void multilayer_soil_profile_avg::do_operation() const
     double tot_soil_field_capacity = 0.0;
     double tot_soil_wilting_point = 0.0;
 
+    //do not use the surface layer since plant would seek deep water
+    //when surface is dry.
```
eloch216: I don't understand this comment about not using the surface layer, since the for loop immediately after it begins at layer 0. Doesn't that mean the surface layer is being used?

src/module_library/soil_water_dynamic_rooting.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "../framework/module.h"
+#include "../framework/state_map.h"
+
+namespace standardBML
+{
+/**
+ *  @class soil_water_dynamic_rooting 
+ *
+ *  @brief Allows max_rooting_layer to change over time
```
eloch216: Can you clarify in the description that this is not based on any published model, but rather is a phenomenological way to roughly reproduce the expected changes in root depth with age?

src/module_library/soil_water_dynamic_rooting.h
```diff
@@ -0,0 +1,75 @@
...
+{
+    return {
+        "max_rooting_layer"  // 
+    };
+}
+
+void soil_water_dynamic_rooting::do_operation() const
+{
+//segment function for now
+//will replace it with continous function to avoid sudden jump
```
eloch216: I don't think I understand this comment. The soil layers are quantized, so I thought there cannot be a continuous change in the maximum rooting layer.

src/module_library/soil_water_uptake.h
```diff
@@ -141,18 +180,84 @@ void soil_water_uptake::do_operation() const
...
+
+    // Calculate the weight of roots in each layer
+    double root_weight[6] = {0.0};
+    double total_weight = 0.0;
+    double current_top_depth = 0.0; // top depth of the rooting layer 
 
     // get total root depth
     for (int i = 0; i < max_rooting_layer; i++) {
-        root_depth = root_depth + soil_depth[i];  // cm
+        // Find the middle depth of the current layer
```
eloch216, line 219: Is this exponential decay from a published model? Or is this a phenomenological addition? Please clarify in a code comment.

src/module_library/soil_water_uptake.h
```diff
@@ -141,18 +180,84 @@ void soil_water_uptake::do_operation() const
...
     }
 
     // uptake in layer = ET*(depth of layer/total depth of root zone)
     // add constraint if soil_water_content_layer < uptake
+    // Calculate uptake using the heavy surface weights
+    // Variable to hold the missing water
+    double unmet_demand = 0.0;
     for (int i = 0; i < max_rooting_layer; i++) {
-        uptake[i] = -canopy_transpiration_rate * (soil_depth[i] / root_depth);  // (Mg*hr/ha)*(cm/cm) = Mg/ha/hr
+       // Check if all dirt is totally dry
```
eloch216: There are a bunch of additions here related to a "safe limit" and "unmet demand." Is this part of a published model? Please clarify in a code comment.

src/module_library/soil_type_selector.h
```diff
@@ -51,14 +51,14 @@ soilText_str get_soil_properties(int soiltype);
...
+        //                             d     d     d     J kg^-1     d     kg s m-3     d     d      d     Mg m^-3
         //                             silt  clay  sand  air_entry   b     Ks          satur fieldc wiltp bulk_density
         {SoilType::sand, {0.05, 0.03, 0.92, -0.7, 1.7, 5.8e-3, 0.437, 0.09, 0.03, 1.60}},
         {SoilType::loamy_sand, {0.12, 0.07, 0.81, -0.9, 2.1, 1.7e-3, 0.437, 0.13, 0.06, 1.55}},
-        {SoilType::sandy_loam, {0.25, 0.10, 0.65, -1.5, 3.1, 7.2e-4, 0.453, 0.21, 0.10, 1.50}},
+        {SoilType::sandy_loam, {0.25, 0.10, 0.65, -1.5, 3.1, 7.2e-4, 0.42, 0.15, 0.10, 1.50}},
         // { SoilType::loam,            { 0.40, 0.18, 0.42,      -1.1,  4.5,  3.7e-4,     0.57, 0.27,  0.12, 1.43 } },
-        {SoilType::loam, {0.39, 0.21, 0.40, -1.1, 4.5, 3.7e-4, 0.463, 0.34, 0.13, 1.30}},  // used for comparison with DSSAR
-        {SoilType::silt_loam, {0.65, 0.15, 0.20, -2.1, 4.7, 1.9e-4, 0.501, 0.33, 0.13, 1.36}},
+        {SoilType::loam, {0.39, 0.21, 0.40, -1.1, 4.5, 3.7e-4, 0.42, 0.30, 0.22, 1.30}},  // used for comparison with DSSAT
```
eloch216: With the changes to `sandy_loam`, `loam`, and `silt_loam`, the values for these soil types no longer match the reference (Campbell & Norman 1998). Please add a code comment explaining where the new values came from.

eloch216: The current comment ("used for comparison with DSSAT") is not very clear.

src/module_library/soil_water_dynamic_rooting.h
yufenghe22: The latter. When I analyze the gs and soil water data, I notice the gs matches surface (near-surface) soil water data (water stress) during early season. I basically just need something to mimic that, i.e., during early season the root should take water from the upper layers, and then move to deeper layers during the mid season. Maybe let me use the Poisson function back? It would serve similar purpose, but I'd need to change the parameters a bit to get a reasonable agreement on soil water content.

src/module_library/multilayer_soil_profile_avg.h
yufenghe22: I'll remove this

src/module_library/soil_type_selector.h
yufenghe22: I'll remove this

src/module_library/soil_water_uptake.h
yufenghe22: The entire uptake module is a new addition (not from DSSAT). I made changes to the original one. I will clarify this

yufenghe22:
> These changes have altered the biomass values quite a lot. Is this an expected result? (Just curious.)

Possibly. We'd need to see how it goes after optimization.

src/module_library/soil_water_dynamic_rooting.h
yufenghe22: removed

justinmcgrath:
Is this something that needs to be done before we work on the stacking? If not, I saw that we postpone it.

yufenghe22:
> Is this something that needs to be done before we work on the stacking? If not, I saw that we postpone it.

Yes. Let's have this version in first. This is important the get the soil water right. As I see it, the only necessary change needed later is to use an alternative root distribution function.

justinmcgrath:
Ok, then let's go with the minimum needed for now, but here are some of my thoughts to potentially be incorporated at a later time.

It looks like the model specifies layers since it goes to 5 and I doubt that's meters. I find that confusing rather than using depth. It also makes the equations dependent on the number of layers, so if that changes, then someone needs to rewrite this and sort out the layers would correspond to each other.

Units are missing in many places.

Here are the soil types at SoyFACE from the Web Soil Survey.

image

The plots are all in the Flanagan and Drummer. The upper layers in those are silt loam and silty clay loam. The lower layers appears to have less silt and more clay.


Drummer
image


Flanagan
image

eloch216 (approved):

PR #281, opened 2026-03-17
New weather data

eloch216:
This PR extends the weather data to 2025.

There are a few minor changes to past values of the atmospheric CO2 concentration in `data/catm_data.R`. This is explained by a note in the original source for the values:

```
# NOTE: In general, the data presented for the last year are subject to change, 
# depending on recalibration of the reference gas mixtures used, and other quality
# control procedures. Occasionally, earlier years may also be changed for the same
# reasons.  Usually these changes are minor.
```


man/cmi_weather_data.Rd
```diff
@@ -9,7 +9,7 @@
 
 \description{
   Champaign, IL weather data specified at hourly intervals in the CST time zone
-  for the years 1995--2023. The data includes typical inputs required for BioCro
+  for the years 1995--2025. The data includes typical inputs required for BioCro
   imulations. Note: some values are missing near the start of 1995 since those
```
leavesandlemmas: Typo: `simulations` not `iìmulation`

data/catm_data.R
```diff
@@ -1,9 +1,9 @@
 catm_data <- utils::read.table(textConnection('
```
leavesandlemmas, line 1: Nothing to do with the PR per se, but isn't there CO2 data from the efflux towers for the area? I wonder how much CO2 concentration varies locally.

leavesandlemmas (approved):

eloch216:
While comparing this new weather data to the Energy Farm weather data, I noticed a weird part of 2024 where the daily average `solar` values were increasing linearly (see figure below). Looking into it more, it turns out there are no SURFRAD logs for days 213-234 of 2024, and day 235 is incomplete. So, the strange `solar` values were a result of the interpolation process we use to fill in gaps in the weather data. This process works well for small gaps, but not when the gap spans three weeks. To fix this, I will use `solar` values from the WARM database instead of SURFRAD during this time period.

image

eloch216:
Using the WARM data, the 2024 `solar` values look more reasonable:

image

data/catm_data.R
yufenghe22: There would be diurnal changes from the flux tower due to respiration and photosynthesis.

yufenghe22 (approved):

man/cmi_weather_data.Rd
eloch216: I checked the "blame" and this typo has been around for four years. Good catch!

PR #283, opened 2026-03-20
Lower maxlay -- refactor to make MAXLAY a local variable

leavesandlemmas:
So the `Light_profile` struct in `sunML.h`  allocates arrays of size 200 for 200 layers even though we only use 10 layers for default biocro.   I created this branch to test whether or not those extra memory allocations have a significant impact on speed. 

The answer seems to be no. Changing the number of layers seems to have a negligible effect. I got very different speeds running from command line vs. running from Rstudio. The test script is below. 

That said, I am still making a PR because I made a small change to the code which improves its readability IMO.   Because AFAIK, MAXLAY is not used anywhere other than inside light_profile, I replaced it with a static, constant  expression integer `max_layers` defined in the `Light_profile` struct.  As a result, the `AuxBio.h` is not needed (I think ultimately that file should be deleted and anything we still use should be moved elsewhere). This makes the `sunML.h` and `sunML.cpp` marginally more modular.  

Long term there are other changes I would make, but those aren't a priority. 

 
```r
library(BioCro)
library(microbenchmark)

runit <- function(ode_solver) {
  run_biocro(
    soybean$initial_values,
    soybean$parameters,
    soybean_weather[['2006']],
    soybean$direct_modules,
    soybean$differential_modules,
    default_ode_solvers[[ode_solver]]
  )
}

res <- microbenchmark(
  runit("boost_euler"),
  runit("boost_rkck54")
  times = 10L
)
```

### DIFF
--- a/src/module_library/sunML.h
+++ b/src/module_library/sunML.h
@@ -4,21 +4,48 @@
 #include <string>
 #include <cmath>
 #include <stdexcept>
+#include <vector>
 
-struct Light_profile {
-    static constexpr int max_layers = 25;
+struct LightProfile {
+    struct Layer {
+        double height;                       // m
+        double shaded_absorbed_ppfd;         // micromol / (m^2 leaf) / s
+        double shaded_absorbed_shortwave;    // J / (m^2 leaf) / s
+        double shaded_fraction;              // dimensionless
+        double shaded_incident_nir;          // J / (m^2 leaf) / s
+        double shaded_incident_ppfd;         // micromol / (m^2 leaf) / s
+        double sunlit_absorbed_ppfd;         // micromol / (m^2 leaf) / s
+        double sunlit_absorbed_shortwave;    // J / (m^2 leaf) / s
+        double sunlit_fraction;              // dimensionless
+        double sunlit_incident_nir;          // J / (m^2 leaf) / s
+        double sunlit_incident_ppfd;         // micromol / (m^2 leaf) / s
+    };
+
     double canopy_direct_transmission_fraction;  // dimensionless
-    double height[max_layers];                       // m
-    double shaded_absorbed_ppfd[max_layers];         // micromol / (m^2 leaf) / s
-    double shaded_absorbed_shortwave[max_layers];    // J / (m^2 leaf) / s
-    double shaded_fraction[max_layers];              // dimensionless
-    double shaded_incident_nir[max_layers];          // J / (m^2 leaf) / s
-    double shaded_incident_ppfd[max_layers];         // micromol / (m^2 leaf) / s
-    double sunlit_absorbed_ppfd[max_layers];         // micromol / (m^2 leaf) / s
-    double sunlit_absorbed_shortwave[max_layers];    // J / (m^2 leaf) / s
-    double sunlit_fraction[max_layers];              // dimensionless
-    double sunlit_incident_nir[max_layers];          // J / (m^2 leaf) / s
-    double sunlit_incident_ppfd[max_layers];         // micromol / (m^2 leaf) / s
+
+    std::vector<Layer> layers;
+
+    // default initialize to nlayers
+    explicit LightProfile(size_t nlayers) : layers(nlayers) {}
+
+    size_t nlayers() const { return layers.size(); }
+
+    Layer& operator[](size_t idx) {
+        return layers[idx];
+    }
+
+    const Layer& operator[](size_t idx) const {
+        return layers[idx];
+    }
+
+    // with bounds checking
+    Layer& at(size_t idx) {
+        return layers.at(idx);
+    }
+
+    const Layer& at(size_t idx) const {
+        return layers.at(idx);
+    }
 };
 
 double thin_layer_absorption(
@@ -74,7 +101,7 @@ double shaded_radiation(
     double ell         // dimensionless from m^2 leaf / m^2 ground
 );
 
-Light_profile sunML(
+LightProfile sunML(
     double ambient_ppfd_beam,       // micromol / (m^2 beam) / s
     double ambient_ppfd_diffuse,    // micromol / m^2 / s
     double chil,                    // dimensionless from m^2 / m^2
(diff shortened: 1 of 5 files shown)


yufenghe22 (approved):

yufenghe22:
> So the `Light_profile` struct in `sunML.h` allocates arrays of size 200 for 200 layers even though we only use 10 layers for default biocro.

Why 200 layers!? 🤣

leavesandlemmas:
@yufenghe22  no idea, you'd have to ask whoever created the `sunMLh` and `AuxBio.h` files, but I think those predate the version of the model for Megan's 2022 paper.   

The `Light_profile` uses a C array, and the size of the array has to be known at compile time. The code uses what I believe is a convention in C programming. You define a macro which corresponds to the size of the array, so you can change this and recompile. 

To use a C array where the number of layers is determined at run time (which is the case when called from R), then the C array's dimension sets the maximum number of layers. So I suspect someone set it to 200, thinking that was sufficiently high to never be an issue. 

But the issue is that memory is allocated for 200 layers even if the computation only uses 10. It's not a huge cost, but it would add up. And it is easy to change, and therefore easy to test if it made a big difference, hence the PR. 

Frankly, it would be more ergonomic to change

```cpp

struct Light_profile
   double height[max_layers];                       // m
    double shaded_absorbed_ppfd[max_layers];         // micromol / (m^2 leaf) / s
    double shaded_absorbed_shortwave[max_layers];    // J / (m^2 leaf) / s
    double shaded_fraction[max_layers];              // dimensionless
    double shaded_incident_nir[max_layers];          // J / (m^2 leaf) / s
    double shaded_incident_ppfd[max_layers];         // micromol / (m^2 leaf) / s
    double sunlit_absorbed_ppfd[max_layers];         // micromol / (m^2 leaf) / s
    double sunlit_absorbed_shortwave[max_layers];    // J / (m^2 leaf) / s
    double sunlit_fraction[max_layers];              // dimensionless
    double sunlit_incident_nir[max_layers];          // J / (m^2 leaf) / s
    double sunlit_incident_ppfd[max_layers];         // micromol / (m^2 leaf) / s
};
```

to something like this 

```cpp
struct LightProfile {
    struct Layer {
          double absorbed_ppfd;
          double incident_ppfd; 
          // etc
    };

// either 
    std::vector<Layer> layers;
// or 
     std::array<Layer, 10> layers;
double canopy_transmission;
// other variables not specific to each layer 

};

```

The `Layer` struct does not have to be defined inside the `LightProfile` struct. I like doing that because it acts the same as a namespace, and it makes the data structure `Layer` specific to the `LightProfile` so the identifier `Layer` could be used elsewhere in other contexts to mean something else. 

Given how this is used elsewhere, `std::vector` would be better so the number of layers is dynamic. 

Because written this way, iterating over the layers is simpler and changes to what is stored for each light layer is better.   I have a branch with changes like this, as part of trying to reexpress the canopy light integral as an explicit integral. And actually, it is not necessary to have a `LightProfile` store explicit light layers at all, so the memory does not need to be allocated at all. 

 Unfortunately, refactoring the code is a bit slow and I have more pressing priorities lately.

src/module_library/sunML.h
```diff
@@ -1,21 +1,24 @@
 #ifndef SUNML_H
 #define SUNML_H
 
-#include "AuxBioCro.h"  // for MAXLAY
+#include <string>
+#include <cmath>
+#include <stdexcept>
 
 struct Light_profile {
+    static constexpr int max_layers = 25;
```
eloch216: I would choose a higher limit. Personally, I have used at least 50 layers when checking for convergence. I think 100 would be a safer limit, especially since the limit doesn't seem to influence the run times.

eloch216:
> As a result, the `AuxBio.h` is not needed (I think ultimately that file should be deleted and anything we still use should be moved elsewhere). 

Removing `AuxBioCro.h` has been a long-term goal of mine for years. I think I even made a branch called `eliminate-aux-biocro` back in 2020 or so. But as I started working on it, I realized how complicated it would be. So I have settled for slowly moving things out of `AuxBioCro` and into their own dedicated header files as I come across them.

FYI, everything in `lightME.h`, `sunML.h`, and `canopy_photosynthesis_outputs.h` was originally in `AuxBioCro`. There was also a copy of the leaf transpiration calculations in there. (For some reason, there were separate transpiration functions for C3 and C4 leaves. One was in `AuxBioCro` and one was in its own header file.) I think there are others too but I don't remember them all -- possibly the respiration functions, temperature response functions, and water vapor property functions were in there too. Anyway, my point is that if you think `AuxBioCro` is bad now, you should have seen it five years ago!

In the past, `MAXLAY` was also used in other places, which is why it was still defined in `AuxBioCro`. But I guess those other references have since been removed. I am always happy to see another thing removed from `AuxBioCro` and another `#include` directive to it eliminated. One day we will succeed in slaying the beast.

eloch216:
I don't understand the check failure for macOS-latest using R version 4.3.0. The check is failing because of an error encountered while setting up R:

```
Downloading macOS utils
  Error: The operation was canceled.
```

Hopefully this will resolve itself. Otherwise we may have to switch to a newer version of R when testing on Mac (again).

src/module_library/sunML.h
eloch216: I approve of the PR pending this one change.

leavesandlemmas: What about the simple refactor so that the number of layers can be chosen dynamically?

eloch216: > What about the simple refactor so that the number of layers can be chosen dynamically?

I think I'm not sure what you mean here. The number of layers _can_ be chosen dynamically when using the `BioCro:c3_canopy` module. My request was just to change the limit from 25 to 100 layers in this line of code.

leavesandlemmas: I implemented what I was talking about.  See current commit.

leavesandlemmas:
Okay, I made a few more changes so that LightProfile now allocates the number of layers based on the `nlayers` passed to  the `sunML` function desires (well up to `2^32`  layers).   So that removes MAXLAY and removes any hard coded limited on  the number of layers. 

I structured  as I suggested above.

eloch216:
Looks good to me

eloch216 (approved):

PR #284, opened 2026-04-01
use a new ball-berry water stress function

yufenghe22:
In this PR, I replaced the default linear water stress function with a new version. This updated function is also a linear fit, so I named it `stomata_water_stress_linear_v2`. I'm open to suggestions to change the name.
This update addresses two main issues: 
- The default linear water stress severely underestimates stomatal conductance ($g_s$).
- The new linear function is a linear fit to soil water content, and can go above 1, i.e., in well-watered condition, the ball-berry parameters can increase beyond their standard values (See analysis below).

## Why $StomataWS$ can be larger than 1?
We apply a scaling factor, $StomataWS$ (0-1), to the ball-berry parameters $b0$ and $b1$ (I'll just refer to $b1$ after) to model the impact of soil water status on stomata conductance. If $StomataWS<=1$, the assumption is that $b1$ represents some kind of maximum value, but is it true though? Unless $b1$ is measured under perfectly water-stress-free condition, it could still be increased (i.e., $StomataWS>1$).
Our current $b1$ value comes from the Leakey's paper ([See Fig. 2]( https://doi.org/10.1111/j.1365-3040.2006.01556.x)), and they did not seem to deliberately choose days that are free of water stress, so I'd assume the $b1$ value represents an average environmental condition rather than a true maximum. Therefore, under well-watered condition, $b1$ should be allowed to go above this average value.
Related but for maize, there’s a significant difference in the $b1$ under water-stress and well-water conditions. See [Fig. 1 in this](https://onlinelibrary.wiley.com/doi/full/10.1111/pce.12990#:~:text=(a)%20Relationship%20of%20stomatal%20conductance%20(gs)%20with%20the%20Ball%20Index%20for%20sunlit%20leaves%20of%20well%2Dwatered%20greenhouse) paper.

## How was the new linear function created?
I use the $g_s$ data from [the Gray et al (2016)](https://doi.org/10.1038/nplants.2016.132) to create and calibrate the new water stress function. They did gas exchange measurements of $g_s$ on the **midday of several days** throughout the growing season during 2009-2011. To match their measurement conditions, I ran **c3_leaf_photosynthesis** with the following set-ups:

- Use a constant boundary layer conductance ($gbw_canopy$) value that is normally read from a LI-COR file: $gbw_{canopy} = 0.072\ m/s\ or\ 2.9\ mol\ m^{-2} s^{-1}$
- Use a large wind speed (100 m/s) to simulate well-mixed air environment to minimize the leaf boundary layer.
- Use measured dynamic Vcmax/Jmax that are different on different days to exclude the potential of enzyme change due to environmental change or aging.
- Because there could be a difference in how they derive their Vcmax25 and Jmax25 values (e.g., they use different temperature response parameters), I scale all their measured Vcmax25 and Jmax25 by factors of 1.22 and 1.06 respectively so that their **means** match our default values of Vcmax25 (110) and Jmax25 (195).
- Use a $Chil$ value of $1e8$ to simulate a flat leaf to match what LI-COR measurement does to clap on a leaf. And use the **absorbed PPFD** from the sunlit top leaf.

#### $g_s$ underestimation with the default linear water stress 
The figure on the left below shows a simulation without water stress, compared to the figure on the right with the default linear water stress. It's clear that the default water stress is too strong. Another key point from the figure on the left is that there's a bit of underestimation of the high $g_s$ part and overestimation of the low $g_s$ part. Since this is under water stress free condition ($StomataWS =1$), I assume the errors should be solely attributed to $StomataWS$. Now we ask, can we find the $StomataWS$ for each point to make a perfect prediction?
image

#### Reverse engineering of the $StomataWS$
So I ran a optimization on the $StomataWS$ values to fit the $g_s$ data. It turns out we can make a perfect prediction from this. The next step is to correlated the **fitted** $StomataWS$ with some independent variables, like the soil water content (SWC).
image

#### Fit the $StomataWS$ function
So i plot the $StomataWS$ against two independent variables: SWC (Leaf figure below) or REW (Right figure below). The linear fitting performance is similar between the two since the points are scattered. The range of the fitted $StomataWS$ is from 0.7 to 1.4, suggesting an increase of $b1$ under well-watered contidions.
image

In the commit, I used the linear function with SWC, and the $g_s$ prediction looks much better (see figures below) than using the default function. Using a linear function with REW gives a similar result. I could also try non-linear fitting, but given that these points are rather scattered, the fitness wouldn't change much. A non-linear function like an asymptotic one may have better physical meanings as some people have used before in their models.
image

@leighmatth


justinmcgrath:
Was there supposed to be more information in the post? You asked how the new function was created it, but didn't you create this?

To explain the old one, the thinking is that the Ball-Berry curve gives the stomatal response in an unstressed condition. Based on that, you would never have a situation that made gs higher than what the BB response gave, so StomataWS has a maximum value of 1.

On the other end, it says that when the SWC is at the wilting point, by definition turgor is 0, and based on that there's the assumption that conductance should be at it's lowest value. It then says that everything in between is scaled linearly.

I don't know why a linear relationship is assumed, and that's very likely not right. Using something other than linear could fix a lot of the problem. I'm not entirely sure that assuming gs = b0 at the wilting point is a good assumption either. It definitely seems reasonable, but I don't feel like it's something that's necessarily true.

In this new version, it's only SWC, and although that may work well for the data you've fit it too, I suspect it will cause big problems in other situations since now there's nothing specific to the soil type. Using the difference between WP and FC is a rough way to incorporate water potential, which is what really drives this, and in this version that's completely gone.

Conceptually, it doesn't make sense that this would increase conductance. I don't know that that is terribly important, but this is losing a connection to the physiological mechanism that's being modeled, which possibly reduces generality and definitely makes things difficult to interpret from a physiological standpoint.

Rather than reworking this as a linear model, can you keep the scaling between 0 and 1 between WP and FC, but make it nonlinear? Something of this sort maybe: `y= C−R*e^(−k*x)`, where x is SWC, with parameters so y = 0 at x = WP, and y is 1 at x = FC? That has an asymptote, so you'd need to do a little something to get y = 1 at x = FC where the asymptote itself isn't 1, but it shouldn't be difficult to work out.

yufenghe22:
> Was there supposed to be more information in the post? You asked how the new function was created it, but didn't you create this?

Sorry. I'm drafting this. I will request your review after it's done. Yes. I can use the difference between WP and FC. I had a version like in [the initial commit](https://github.com/biocro/biocro/pull/284/changes/ff917d88f1bd6cfe73121a800fc8d45a7c4ba3ec#diff-886290967c4ec216c3e262f4dfd5389ac6945f512880f738472616342ff5db28R62), but I was thinking maybe it could be further simplified because in terms of fitting performance, using either the SWC directly or the nomalized term does not make much of a difference.

justinmcgrath:
How does REW go above 1? Field capacity is supposed to be the largest SWC the soil can hold, so theta - theta_wp > theta_fc - theta_wp is unexpected. Field capacity isn't a perfectly well defined thing, so seeing SWC somewhat higher than that could happen, but 1.5 and higher for REW seems like something is wrong with the soil type being used.

For a single location, where the soil type is the same for every point, I wouldn't expect using SWC vs REW to matter here because you're parameterizing this in a way that forces the model to match observations. One you go to a different soil type though, I suspect this isn't going to work at all using SWC, and there won't be a set of parameters that works across different soil types. At the very least I think this approach needs to use REW, or that there needs to be some evidence to indicate that SWC only makes sense across soil types.

Overall, this doesn't make too much sense to me either. There are two major changes here. First, b0 and b1 are allowed to vary somewhat independently. That I think can make sense in general, and quite likely that is what happens. There isn't much reason to think that if b1 is halved, then b0 should be halved.

Second, is the part of using a different scale and approach from StomWS, where it is making the data fit better, but at a substantial cost to the mechanistic representation, when it seems like almost the same thing can be done in a different way. Based on what you described, the problem appears to lie in b0 and b1 not representing water-replete parameters. So we're treating them as the maximum, but they're more like a middle value. If you use the old linear model and fit b0 and b1 to fit predict gs to observed gs, does it work as well as using the new model?

If it works, it indicates that we're using inappropriate parameters. We wouldn't want to fit parameters like that, but it makes a case for giving them a new reference point on the SWC scale, rather than treating them as values at FC.

justinmcgrath:
And maybe this is because of the very large REW values. Maybe the soil parameters are not appropriate and it's making the SWC more stressful than it should.

yufenghe22:
> How does REW go above 1? Field capacity is supposed to be the largest SWC the soil can hold,

Saturation is the largest SWC soil can hold. Saturation is between 0.4-0.5 for all soil types as you can see from our table in soil_type_selector.h. It may take a couple of days for soil to drain from fully-saturated to the field capacity. The average values for field capacity is 0.33 and the wilting point is 0.25, which line up with the Table 9.1 in the Campbell & Norman book reasonably well for silty-loamy type. For example, if you just have a SWC of 0.4, this would give you a REW of 1.875. Observed SWC also show points with SWC over 0.4 during the season.

> First, b0 and b1 are allowed to vary somewhat independently.

I'm not sure I understand this. The $StomataWS$ is applied to b0 and b1 at the same time right?

> If you use the old linear model and fit b0 and b1 to fit predict gs to observed gs, does it work as well as using the new model?

Do you mean using the old linear water stress function (with lots of low $StomataWS$ values), but increase the b0 and b1 to achieve a good fit? I think that would work too. However, I think there are two ways of doing this: 1. we fit b0 and b1 for individual points to get a perfect fit and then do an average of them. 2. in the optimization, we assume the same pair of b0 and b1 for all points.

src/module_library/stomata_water_stress_linear_v2.h
```diff
@@ -0,0 +1,74 @@
...
+namespace standardBML
+{
+class stomata_water_stress_linear_v2 : public direct_module
+{
+   public:
+    stomata_water_stress_linear_v2(state_map const& input_quantities, state_map* output_quantities)
+        : direct_module{},
+
+          // Get pointers to input quantities
+          soil_field_capacity_ip{get_ip(input_quantities, "soil_field_capacity")},
```
leavesandlemmas: use get_input and const& instead of input pointer?

leavesandlemmas:
Missing module test cases..

yufenghe22:
> Missing module test cases..

Yes. Will add that and also formatting. I want to first make sure we agree on the form of modification we need. I think Justin had a good point (assuming I interpret his suggestion correctly): instead of making a stomataWS function that go over 1, which might cause troubles with reviewers, we could simply increase the default values of b0 and b1. So the rationale is that the b0 and b1 are now treated as fitted parameters.

leavesandlemmas:
@yufenghe22  Overall I think an easy-to-fit function for water stress is a good idea, even if its not particularly realistic.  I don't think there's anything particular special about the logic of the old version. It seems to me that the logic of the old one is basically this: The stomata open at high SWC and close a low SWC, so interpolate between the max and min values. However, if the stomata respond mostly to leaf water potential (and abscissic acid, AKA ABA, transported from the roots), and leaf water potential depends on both soil water potential and the vapor pressure of the air (along with stomatal conductance), then the respond should be pretty nonlinear when expressed in SWC. Although it depends on the soil type, SWC dropping from 35% to 25%  has basically no effect on soil water potential (like < 0.1 MPa), but the drop from 25% to 15%  is bigger, and below that water potential shoots off toward negative infinity. 

E.g.  see the plot in the wikipedia page: [Water retention curve](https://en.wikipedia.org/wiki/Water_retention_curve). Note the logarithmic scale on the y-axis. I also measured the relationship between VWC and soil water potential in my phd, so here's an example (though I think this curve is particularly sharp because of the soil):  

vwc_water_potential1


 

> using the difference between WP and FC is a rough way to incorporate water potential, which is what really drives this, and in this version that's completely gone.

Why don't we just use a stomata model with water potential then?  It would be straightforward to create a module for the moisture retention curves, and the curves are easy enough to parameterize.  The moisture retention curve would capture the different responses by soil type.  The Tardieu and Davies one uses leaf water potential directly. Maybe we combine it with the Ball-Berry model  ?

https://onlinelibrary.wiley.com/doi/full/10.1046/j.1365-3040.2002.00909.x

data/soybean2.R
```diff
@@ -8,7 +8,7 @@ soybean2 <- list(
         "BioCro:soybean_development_rate_calculator",
         "BioCro:soil_type_selector",
         "BioCro:multilayer_soil_profile_avg",
-        stomata_water_stress = "BioCro:stomata_water_stress_linear",
+        stomata_water_stress = "BioCro:stomata_water_stress_linear_v2",
```
leavesandlemmas: For names, how about these

1. `stomata_stress_by_soil_water_content` for the new one
2. `stomata_stress_by_relative_soil_water_content` for the old one.

src/module_library/stomata_water_stress_linear_v2.h
```diff
@@ -0,0 +1,74 @@
...
+    // Collect inputs and make calculations
+    double soil_wilting_point = *soil_wilting_point_ip;
+    double soil_field_capacity = *soil_field_capacity_ip;
+    double soil_water_content = *soil_water_content_ip;
+
+    // This linear fitted equation is from a reverse engineering of fitting observed gs data
+    // Ref: Gray, S., Dermody, O., Klein, S. et al. Intensifying drought eliminates the expected benefits of elevated carbon dioxide for soybean. Nature Plants 2, 16132 (2016). https://doi.org/10.1038/nplants.2016.132
+    double linear_fit = 1.88 * soil_water_content + 0.379;
+    // From my data fitting, I do not see any StomataWS values beyond this range
+    double StomataWS_min = 0.5;
```
leavesandlemmas: Minor suggestion, but make the minimum and maximum values `constexpr double`

src/module_library/c3photo.cpp
```diff
@@ -92,7 +92,9 @@ photosynthesis_outputs c3photoC(
     double const alpha_TPU = 0.0;  // dimensionless. Without more information, alpha=0 is often assumed.
 
     // Adjust Ball-Berry parameters in response to water stress
-    double const b0_adj = StomWS * b0 + Gs_min * (1.0 - StomWS);
+    // Cap StomWS for the intercept to avoid negative values
+    double const StomWS_for_b0 = (StomWS > 1.0) ? 1.0 : StomWS;
```
leavesandlemmas: These lines suggest that all stomata water stress modules should clamp to the interval `[0,1]`. But that's a minor suggestion.

leavesandlemmas (approved):
You have my go ahead to merge, once the tests are added. I think an easy-to-fit function is a good idea, but it might be too difficult to split this module into two. One that computes soil water content from soil water potential (I believe this is also called the pressure head in hydrology, but don't cite me on that), and a second one that computes the stomata water stress parameter from soil water potential.


Personally I think we should just use other stomata models, and maybe let the ball-berry model not account for soil water stress at all, since it isn't part of the original model at all.

leighmatth:
> You have my go ahead to merge, once the tests are added. I think an easy-to-fit function is a good idea, but it might be too difficult to split this module into two. One that computes soil water content from soil water potential (I believe this is also called the pressure head in hydrology, but don't cite me on that), and a second one that computes the stomata water stress parameter from soil water potential.
> 
> Personally I think we should just use other stomata models, and maybe let the ball-berry model not account for soil water stress at all, since it isn't part of the original model at all.

Someone in my group is working on adding water potential (soil and organ level) to biocro with the idea of then relating the gs to water potential.

yufenghe22:
> It would be straightforward to create a module for the moisture retention curves, and the curves are easy enough to parameterize.

There is some empirical estimation of soil water potential in [existing BioCro module ](https://github.com/biocro/biocro/blob/ca094136b3581b9005220dfd01d59fa9c8a5fcf7/src/module_library/bucket_soil_drainage.h#L100) that is not being used.

yufenghe22:
@justinmcgrath Updates:
Fitting $b0$ and $b1$ with the default linear StomataWS function did not generate a good fitting. I think the default linear StomataWS creates too much variation due to its strong linear response to SWC. I also tried using the water potential on the x-axis, but I couldn't find a good fitting function between the optimized StomataWS and water potential. 
To address to confusion that may be raised by having $StomataWS > 1 $, I increased the default $b0$ and $b1$ values by 1.4 assuming these values represent the true water-stress-free maximums. The value 1.4 comes from the previous fitting of the StomataWS (See figure results above).
Now, all optimized StomataWS values are below 1 (see figures below). I also use REW as the independent variable instead of SWC. The performance of the predicted gs are slightly better, but similar to the previous results. Without implementing a more mechanistic water stress model with water potentials, I think this is a good fix for now. If this looks okay, I will fix the test and formatting.
image

eloch216:
I haven't taken a careful look at this yet. But I do have two immediate thoughts:

1. Because some of the module parameters were determined by a fitting procedure, it would be good to have a fitting script available in `inst/extdata` so the fit can be redone when necessary. As a related point, it may also be helpful to treat these parameters as module inputs rather than being hardcoded.

2. The main motivation for this new water stress function is to ensure the model reproduces the measured `gsw` values reported in Gray et al. (2016) when using the new soil water modules. But we haven't finished checking those modules, and there may still be mistakes in their code. If there are any changes to the behavior of these modules, there will be changes to the predicted soil water content and hence to `StomataWS`. So we don't truly know if the new water stress module solves the issue, or if the issue will even exist once the new soil water modules have been checked.

yufenghe22:
> I haven't taken a careful look at this yet. But I do have two immediate thoughts:
> 
> 1. Because some of the module parameters were determined by a fitting procedure, it would be good to have a fitting script available in `inst/extdata` so the fit can be redone when necessary. As a related point, it may also be helpful to treat these parameters as module inputs rather than being hardcoded.
> 2. The main motivation for this new water stress function is to ensure the model reproduces the measured `gsw` values reported in Gray et al. (2016) when using the new soil water modules. But we haven't finished checking those modules, and there may still be mistakes in their code. If there are any changes to the behavior of these modules, there will be changes to the predicted soil water content and hence to `StomataWS`. So we don't truly know if the new water stress module solves the issue, or if the issue will even exist once the new soil water modules have been checked.

- I can add the data and scripts to the `inst/extdata`.

- Not sure about the 2nd point is related to this PR. I think the soil water module is mostly checked, particularly corrected for the hardcoded numbers that were for daily time step. If there will be big changes in model behavior, we can repeat this process to update the fitted coefficients. Same with partitioning coefficients.

yufenghe22:
> If there are any changes to the behavior of these modules, there will be changes to the predicted soil water content and hence to `StomataWS`

The current soil water module does a pretty good job capturing the soil water content at different depths. I doubt that there would be drastic changes on the predictions after going through the those modules and fixing potential errors.

For those days when the gs data were collected, here is a comparison of the averaged swc based on rooting depths:
image

Here is a comparison for different years.

image

justinmcgrath:
> Fitting b0 and b1 with the default linear StomataWS function did not generate a good fitting. I think the default linear StomataWS creates too much variation due to its strong linear response to SWC. I also tried using the water potential on the x-axis, but I couldn't find a good fitting function between the optimized StomataWS and water potential.
> To address to confusion that may be raised by having StomataWS > 1 , I increased the default b0 and b1 values by 1.4 assuming these values represent the true water-stress-free maximums. The value 1.4 comes from the previous fitting of the StomataWS (See figure results above).

Scaling these by 1.4 is essentially what I was getting at. I think the underlying issue is that we have treated the Ball-Berry values from Andrew's group as representing optimal water status conditions, that is, when StomataWS = 1, but they really represent conditions at StomataWS = 1/1.4.

yufenghe22:
Working on adding calibration data and scripts to the `inst/extdata`.

leavesandlemmas (approved):

yufenghe22:
@eloch216 Added data and scripts for repeating the calibration processes. Please take a look when you get a chance.

data/soybean2.R
yufenghe22: Now this new fitted equation also uses the relative SWC, so I think `v2` might not be a bad idea.

eloch216:
I tried running the script and got this error message from the optimizer:

```
N = 17, M = 5 machine precision = 2.22045e-16
At X0, 0 variables are exactly at the bounds
At iterate     0  f=      0.31551  |proj g|=      0.14351
At iterate    10  f =    0.0053623  |proj g|=       0.14362
At iterate    20  f =   1.7229e-05  |proj g|=     0.0021612
Bad direction in the line search;
   refresh the lbfgs memory and restart the iteration.
Line search cannot locate an adequate point after 20 function
and gradient evaluations
final  value 0.000017 
stopped after 21 iterations
```

Have you seen this before? Is it because the values of `StomataWS_gradient` and `StomataWS_intercept` were already optimized?

eloch216:
I have a few minor suggestions for the script. Is it okay for me to commit them directly? Otherwise I could make a new PR.

Update: I will make a PR, because my suggestions have gotten more complicated 😂

yufenghe22:
> Have you seen this before? Is it because the values of `StomataWS_gradient` and `StomataWS_intercept` were already optimized

yes, i saw this as well, even when starting from a non-opt initials. But it can find a solution even with this warning

eloch216:
I am not sure if it's okay to store the data from Gray et al. (2016) as an RDS file. The documentation for `saveRDS` includes this:

> **Warning**
> 
> Files produced by `saveRDS` (or serialize to a file connection) are not suitable as an interchange format between machines, for example to download from a website. The files produced by `save` have a header identifying the file type and so are better protected against erroneous use.

yufenghe22:
> I am not sure if it's okay to store the data from Gray et al. (2016) as an RDS file. The documentation for `saveRDS` includes this:
> 
> > **Warning**
> > Files produced by `saveRDS` (or serialize to a file connection) are not suitable as an interchange format between machines, for example to download from a website. The files produced by `save` have a header identifying the file type and so are better protected against erroneous use.

I've been using RDS between Linux and Mac without any issues, including multi-GB files. That said, I agree it’s not ideal for general distribution due to portability and usability concerns. I’ve switched to `CSV` to ensure compatibility for a wider range of users.

yufenghe22:
- Merged the lastest changes from the soil water branch.
- Re-calibrated the `StomataWS_gradient` and  `StomataWS_intercept`.
- The biomass results are quite different now. The partitioning coefficients will need to be updated. I will open another PR once I complete that after merging this.

See updated results below: 
image

image

PR #285, opened 2026-04-03
Added some documentation to the multidimensional roots library

leavesandlemmas:
I spent an hour adding some documentation to the multidimensional root solving library code with a little from AI.  I also made a few minor changes. 

1. Root-solving methods should make all class methods that implement the method private, but make `zero_finding_method` a friend. 
2. Introduced a `Status` enum to replace the `is_valid` bool. I think the intended usage is clearer now, if someone wanted to introduce a new root-solving method. 

These changes could probably be done to the one-dimensional root solving library, but that's not a priority right now.

### DIFF
--- a/src/math/roots/multidim/zeros.h
+++ b/src/math/roots/multidim/zeros.h
@@ -8,30 +8,97 @@
 #include "../../linalg/base.h"
 namespace root_multidim
 {
-// termination states
+/**
+ * @brief Termination status codes for zero-finding solvers.
+ *
+ * Returned inside result_t to indicate why iteration stopped.
+ * Successful termination is indicated by `residual_zero` or `delta_x_zero`;
+ * all other values indicate failure or a limit was reached.
+ */
 enum class Flag {
-    residual_zero,
-    delta_x_zero,
-    zero_is_nonfinite,
-    max_iterations,
-    function_is_nonfinite,
-    singular_matrix
+    residual_zero,          // f(x) == 0
+    delta_x_zero,           // current_x == last_x
+    zero_is_nonfinite,      // x = NaN, Inf, -Inf,
+    max_iterations,         //
+    function_is_nonfinite,  // f(x) = NaN, Inf, -Inf,
+    singular_matrix         // Esimtate
 };
 
+/**
+ * @brief Holds the outcome of a zero-finding solve.
+ *
+ * Aggregates the final estimate, residual, iteration count, and the
+ * reason iteration stopped.
+ *
+ * @tparam Dim Dimension of the system (number of equations = unknowns).
+ */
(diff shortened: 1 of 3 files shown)


leavesandlemmas:
@pxm726 feel free to comment on the PR, if you have questions. The code is mostly the same on `main` and `develop` but if you have questions / comments  I can add stuff to the documentation in the code, directly. 

For Ed and Yufeng, I am not in a hurry to merge this and its mostly comestic anyways, so low priority for you all. Just FYI.

yufenghe22:
Is this supposed to be merged to the `main` or `develop`?

eloch216:
> Is this supposed to be merged to the `main` or `develop`?

The target should be `develop`. Scott, can you change it? That will also make it easier to see the new additions.

leavesandlemmas:
@eloch216  @yufenghe22 Fixed. 

Apparently it is possible to change the default target branch of a new PR, but I do not have access to that setting for BioCro. Do either of you?

eloch216:
> Apparently it is possible to change the default target branch of a new PR, but I do not have access to that setting for BioCro. Do either of you?

That's a good idea. I can change it. With our "new" git workflow, it's much less common to need `main` as the target, so this makes to change.

Update: I think this is not possible. The internet tells me: "To set the default target (base) branch for pull requests on GitHub, you must change the repository's default branch. GitHub does not currently support setting a different default base branch for PRs while keeping a separate branch as the primary repository view." We need to keep `main` as the "default" branch since that's where visitors should be directed. So we will just have to remember to change the target branch for feature branch PRs.

src/math/roots/multidim/zeros.h
```diff
@@ -8,30 +8,97 @@
...
+ * @brief CRTP base class for iterative zero-finding methods.
+ *
+ * Provides the outer solve loop, convergence tolerance helpers, and
+ * result packaging. Concrete methods (e.g. `broyden<Dim>`) derive from
+ * this class via the Curiously Recurring Template Pattern and must
+ * implement three member functions:
+ *
+ * | Function | Signature | Purpose |
+ * |---|---|---|
+ * | `initialize` | `bool initialize(F&& fun, Args&&... args)` | Set up state from the initial guess; return `false` to abort immediately. |
```
eloch216: I know this table will be formatted nicely when the doxygen documentation is built. But it's usually nice to manually include extra blank spaces as necessary so the `|` characters line up with each other. This makes the table easier to read in the original file.

eloch216:
I am always a fan of more documentation so I think this is great. I just have one very small comment about a table (above).

yufenghe22 (approved):
Looks good to me.

eloch216 (approved):

PR #287, opened 2026-04-15
Change soil types

yufenghe22:
I realize there are a few small changes I made on my local repo that I'd like to merge.
- Use type 6 (clay_loam) and 7 (silty_clay_loam) to represent Champaign's soil types to better match well-known local soil characteristics.
- Reset all parameters values to match the Table 9.1 in Campbell and Norman Introduction to Environmental Biophysics. As much as I believe these values have large uncertainties, it's better to match a recognized text to avoid unnecessary confusion.
- Lowered the saturation value for clay_loam to keep the alignment of the high flow with observed data. 

This change does not change much on the overall agreement with observed SWC (results omitted). The difference between the old and new (this) version on biomasses and SWC are shown below. They are overall small. Also, we will need to re-optimize the partitioning coefficients once the new soil water and StomataWS modules are merged.

Code snippet for plotting is show below:
```
result_old$source <- "old"
result_new$source <- "new"
combined_df <- rbind(result_new, result_old)
# This puts all soil_water_content columns into a single 'variable' column
plot_data <- combined_df %>%
  pivot_longer(
    cols = starts_with("soil_water_content"),
    names_to = "sensor",
    values_to = "swc_value"
  )
ggplot(plot_data, aes(x = fractional_doy, y = swc_value, color = source)) +
  geom_line(size = 0.8) +
  facet_wrap(~sensor, scales = "free_y", ncol = 2) +
  ylim(c(0,0.5))+
  labs(
    x = "Time",
    y = "Volumetric Water Content",
    color = "Dataset"
  ) +
  theme_minimal()
xyplot(Leaf + Stem + Root + Shell + Grain  ~ fractional_doy, 
       group = source, plot_data , grid=TRUE, type ='l' , auto.key = T)
```

image
image


eloch216:
> As much as I believe these values have large uncertainties, it's better to match a recognized text to avoid unnecessary confusion.

I agree with this. It's best to have a reference to cite, even if it is uncertain.

> Use type 6 (clay_loam) and 7 (silty_clay_loam) to represent Champaign's soil types to better match well-known local soil characteristics.

Is there a reference to cite for a description of Champaign's soil? Did the info come from the USGS Soil Survey (https://websoilsurvey.nrcs.usda.gov/app/)? If so, it would be helpful to include this in `man/soybean2.Rd`.

> Lowered the saturation value for clay_loam to keep the alignment of the high flow with observed data.

Can you double check that the comment above the definition of `soil_parameters` in `soil_type_selector.h` is up-to-date?

yufenghe22:
> Is there a reference to cite for a description of Champaign's soil? Did the info come from the USGS Soil Survey (https://websoilsurvey.nrcs.usda.gov/app/)? If so, it would be helpful to include this in `man/soybean2.Rd`.
> 
> > Lowered the saturation value for clay_loam to keep the alignment of the high flow with observed data.
> 
> Can you double check that the comment above the definition of `soil_parameters` in `soil_type_selector.h` is up-to-date?

Yes. That's the one Justin posted. Added some info to  `man/soybean2.Rd` and updated the comment in `soil_type_selector.h`.

eloch216 (approved):

PR #288, opened 2026-04-20
Units added to soil_water_downflow.h

dilkaran:
Units to soil_water_downflow.h are added based on dssat-csm-os/Soil/SoilWater/INFIL.for and dssat-csm-os/Soil/SoilWater/WATBAL.for.

### DIFF
--- a/src/module_library/soil_water_downflow.h
+++ b/src/module_library/soil_water_downflow.h
@@ -11,7 +11,14 @@ namespace standardBML
  * @class soil_water_downflow
  *
  * @brief Calculates water flow through the soil.
- *
+ * * Based on DSSAT source files:
+ *   - Soil/SoilWater/WATBAL.for, subroutine WATBAL
+ *     computation of infiltrated_water (WINF) and potential_infiltration (PINF)
+ *     from available_water (WATAVL) and surface_runoff (RUNOFF)
+ *   - Soil/SoilWater/INFIL.for, subroutine INFIL
+ *     layer-by-layer infiltration and drainage
+ * Note: DSSAT rates are on a daily timestep (mm/d, cm/d);
+ *       BioCro rates are on a hourly timestep (mm/hr, cm/hr).
  */
 class soil_water_downflow : public direct_module
 {
@@ -95,7 +102,7 @@ class soil_water_downflow : public direct_module
 
    private:
     // References to input quantities
-    double const& available_water;
+    double const& available_water;  
     double const& surface_runoff;
     double const& swcon;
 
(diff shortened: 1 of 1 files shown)


src/module_library/soil_water_downflow.h
```diff
@@ -168,69 +175,70 @@ class soil_water_downflow : public direct_module
...
-        "soil_saturated_conductivity_6",
-        "soil_saturation_capacity_6",
-        "soil_field_capacity_6"  //
+        "available_water",  // mm / hr. Water available for infiltration or runoff (rainfall plus irrigation) 
+        "surface_runoff",   // mm / hr. 
+        "swcon",            // 1 / hr. Soil water conductivity constant; whole profile drainage rate coefficient
+
+        "soil_depth_1",  // cm. Thickness of soil layer 1
+        "soil_water_content_1",  // cm^3 / cm^3 
+        "soil_saturated_conductivity_1",  // kg s m^-3. TODO: confirm units. data/soil_parameters.R has units of J * s / m^3 for soil saturated conductivity.
```
yufenghe22: That file is incorrect in the unit. In the `src/module_library/soil_type_selector.h`, i've fixed that. It should be `kg s m-3`

yufenghe22 (approved):

eloch216:
To better understand the meaning and units of `potential_infiltration`, I took a look at the `infil` function. In that function, it is compared against `hold`, which I believe is the "holding capacity" of the soil, expressed as a water content with units of `cm`. From this, I believe that `potential_infiltration` should also have units of `cm`.

It looks like this is another situation where there is an implicit multiplication by `timestep`. In other words, the `potential_infiltration` variable represents the amount of water that would be added to the soil if infiltration proceeds at a rate of `infiltrated_water` for one time step.

To address this, I would recommend adding a timestep definition (`double constexpr timestep = 1.0; // hr`) and calculating `potential_infiltration` as `infiltrated_water * timestep * mm_to_cm`. Then it will have the proper units of `cm`.

See the "soil evaporation" branch (https://github.com/biocro/biocro/pull/292) for an example of doing this.

dilkaran:
> To better understand the meaning and units of `potential_infiltration`, I took a look at the `infil` function. In that function, it is compared against `hold`, which I believe is the "holding capacity" of the soil, expressed as a water content with units of `cm`. From this, I believe that `potential_infiltration` should also have units of `cm`.
> 
> It looks like this is another situation where there is an implicit multiplication by `timestep`. In other words, the `potential_infiltration` variable represents the amount of water that would be added to the soil if infiltration proceeds at a rate of `infiltrated_water` for one time step.

So, if we follow this logic, then `drain` should be mm instead of mm/hr because it is:
`double drain = potential_infiltration * 10.0;`
In DSSAT,  `DRAIN = PINF * 10.0`, thus the units should be mm, but in the description:
 `DRAIN      Drainage rate from soil profile (mm/d)`

If we are passing `potential_infiltration` as `infiltrated_water * timestep * mm_to_cm`. Then units for drain should be just mm, right?

src/module_library/soil_water_downflow.h
dilkaran: I will remove the TODO

eloch216:
> > To better understand the meaning and units of `potential_infiltration`, I took a look at the `infil` function. In that function, it is compared against `hold`, which I believe is the "holding capacity" of the soil, expressed as a water content with units of `cm`. From this, I believe that `potential_infiltration` should also have units of `cm`.
> > It looks like this is another situation where there is an implicit multiplication by `timestep`. In other words, the `potential_infiltration` variable represents the amount of water that would be added to the soil if infiltration proceeds at a rate of `infiltrated_water` for one time step.
> 
> So, if we follow this logic, then `drain` should be mm instead of mm/hr because it is: `double drain = potential_infiltration * 10.0;` In DSSAT, `DRAIN = PINF * 10.0`, thus the units should be mm, but in the description: `DRAIN Drainage rate from soil profile (mm/d)`
> 
> If we are passing `potential_infiltration` as `infiltrated_water * timestep * mm_to_cm`. Then units for drain should be just mm, right?

Yes, I think `drain` would have units of `mm`. However, it looks like `drain` is not used for anything -- it's not the input to any other module. And it doesn't seem very useful -- anyone who wanted this value could easily calculate it from `infiltrated_water` after a simulation was run. We could probably just remove it as an output from this module.

eloch216:
> > > To better understand the meaning and units of `potential_infiltration`, I took a look at the `infil` function. In that function, it is compared against `hold`, which I believe is the "holding capacity" of the soil, expressed as a water content with units of `cm`. From this, I believe that `potential_infiltration` should also have units of `cm`.
> > > It looks like this is another situation where there is an implicit multiplication by `timestep`. In other words, the `potential_infiltration` variable represents the amount of water that would be added to the soil if infiltration proceeds at a rate of `infiltrated_water` for one time step.
> > 
> > 
> > So, if we follow this logic, then `drain` should be mm instead of mm/hr because it is: `double drain = potential_infiltration * 10.0;` In DSSAT, `DRAIN = PINF * 10.0`, thus the units should be mm, but in the description: `DRAIN Drainage rate from soil profile (mm/d)`
> > If we are passing `potential_infiltration` as `infiltrated_water * timestep * mm_to_cm`. Then units for drain should be just mm, right?
> 
> Yes, I think `drain` would have units of `mm`. However, it looks like `drain` is not used for anything -- it's not the input to any other module. And it doesn't seem very useful -- anyone who wanted this value could easily calculate it from `infiltrated_water` after a simulation was run. We could probably just remove it as an output from this module.

Update: I have looked at `infil` a bit more and it seems that `potential_infiltration` can be modified by that function, so `drain` is not the same as `infiltrated_water`. So we should probably keep it as a module output.

yufenghe22:
> To better understand the meaning and units of `potential_infiltration`, I took a look at the `infil` function. In that function, it is compared against `hold`, which I believe is the "holding capacity" of the soil, expressed as a water content with units of `cm`.

I think `potential_infiltration` should be in unit of `mm/d` or `mm/hr`. When comparing with `hold` (in unit of `cm`), it assumed a fixed time step of 1 day, so what the if statement should become `potential_infiltration * timestep > hold`. 

This is because  `potential_infiltration` comes from `infiltrated_water` and `double infiltrated_water = available_water - surface_runoff;`. Since `available_water` is basically the `precipitation` input, it must be in the rate unit. Physically `potential_infiltration` means the amount of water that is able to go into the soil in this time step, therefore a rate quantity.

There's other unit inconsistency when adding these quantity like this equation in `soil_water_flow_functions.cpp`: `drn[l] = potential_infiltration - hold + drainage_rate;  // cm/hr`. It's clear `hold` is in `cm` and `drainage_rate` is in `cm/d` (I just follow DSSAT dimension here). If `potential_infiltration` is in `cm/d`, we have a dimension mismatch here. `drn[l]` is clearly a rate. What the mass balance should look like is: 

```
total_mass_drained = potential_infiltration *timestep - hold + drainage_rate*timestep
drn[l] = total_mass_drained/timestep
```

eloch216:
> > To better understand the meaning and units of `potential_infiltration`, I took a look at the `infil` function. In that function, it is compared against `hold`, which I believe is the "holding capacity" of the soil, expressed as a water content with units of `cm`.
> 
> I think `potential_infiltration` should be in unit of `mm/d` or `mm/hr`. When comparing with `hold` (in unit of `cm`), it assumed a fixed time step of 1 day, so what the if statement should become `potential_infiltration * timestep > hold`.

I have started going through the units in `soil_water_flow_functions.h` since it is the next file assigned to me. I began with `infil` since that is the first function in the file. After reading it carefully and consulting an original reference (Ritchie 1998 -- available at https://doi.org/10.1007/978-94-017-3624-4_3), I think I have come to the opposite conclusion to you.

From what I can tell, `infil` is designed to determine the total amount of downward water flow out of each soil layer. Part of this is "drainage," which is flow that would occur even without any infiltration. The other component is infiltration. As infiltrated water moves down through the soil profile, it is absorbed by each layer (up to its saturated water content), and then the remainder is passed on to the layer below. This involves comparing the total infiltration that could potentially occur over the next time step (in `cm`) against the holding capacity of the layer (in `cm`). Because of this, it seems more natural to me to consider `potential_infiltration` to be an amount of water (in `cm`) rather than a rate (in `cm / hr`).

Ultimately, the inconsistency issues with units in the DSSAT code can be resolved in multiple ways, so there isn't really a "correct" way to do it.

@dilkaran I would recommend just leaving `potential_infiltration` as it currently is. Later, I will make a PR about `soil_water_flow_functions`, and at that point we can address any remaining discrepancies.

eloch216:
A little more explanation for my choice:

> This is because `potential_infiltration` comes from `infiltrated_water` and `double infiltrated_water = available_water - surface_runoff;`. Since `available_water` is basically the `precipitation` input, it must be in the rate unit. Physically `potential_infiltration` means the amount of water that is able to go into the soil in this time step, therefore a rate quantity.

Here I completely agree that `infiltrated_water` must be a rate.

> There's other unit inconsistency when adding these quantity like this equation in `soil_water_flow_functions.cpp`: `drn[l] = potential_infiltration - hold + drainage_rate; // cm/hr`. It's clear `hold` is in `cm` and `drainage_rate` is in `cm/d` (I just follow DSSAT dimension here). If `potential_infiltration` is in `cm/d`, we have a dimension mismatch here. `drn[l]` is clearly a rate. What the mass balance should look like is:
> 
> ```
> total_mass_drained = potential_infiltration *timestep - hold + drainage_rate*timestep
> drn[l] = total_mass_drained/timestep
> ``

I also agree about this inconsistency. I guess I just chose to resolve it in a different way. If `potential_infiltration` is in `cm`, then we can have the following mass balance:

```
total_mass_drained = potential_infiltration_cm - hold + drainage_rate*timestep
drn[l] = total_mass_drained/timestep
```

where I have used `potential_infiltration_cm` to distinguish it from your version. This is equivalent to

```
drn[l] = (potential_infiltration_cm - hold) / timestep + drainage_rate
```

In your version, this would be

```
drn[l] = potential_infiltration - hold / timestep + drainage_rate
```

To me, `(potential_infiltration_cm - hold) / timestep` seems to have an intuitive meaning since I can recognize `potential_infiltration_cm - hold` as the excess water above the holding capacity. But the interpretation of `hold / timestep` is not as clear.

yufenghe22:
> Ultimately, the inconsistency issues with units in the DSSAT code can be resolved in multiple ways, so there isn't really a "correct" way to do it.

I guess this stems from the fact that `infiltration` is a loose word. It can either refer to `infiltrated mass` or `infiltration rate`. It's like we say `distance` vs `change in distance`, but we specially refer the latter as `velocity` to avoid ambiguity.

https://soilquality.nres.illinois.edu/infiltration/

dilkaran:
> Ultimately, the inconsistency issues with units in the DSSAT code can be resolved in multiple ways, so there isn't really a "correct" way to do it.
> 
> @dilkaran I would recommend just leaving `potential_infiltration` as it currently is. Later, I will make a PR about `soil_water_flow_functions`, and at that point, we can address any remaining discrepancies.

Okay. I am going to leave `potential_infiltration` as `cm` and `drain` as `mm` and merge the pull request.

PR #289, opened 2026-04-20
Updates to water stress func

eloch216:
This PR makes a few changes to the new water stress stuff:
1. I made some small changes to the optimization script and its associated documentation. Now it prints out the new values for the gradient and intercept. I also added `dev.new()` before the plotting commands, and wrapped the call to `summary` in a `print` command. When I ran this script using `source`, the second plot overwrote the first, and the fit summary was not printed. Some of this behavior depends on the particular operating system and environment (RGui vs RStudio), but adding `dev.new()` and explicit `print` statements make sure it always works properly.
2. Although it was not written this way, the old `stomata_water_stress_linear` sets `StomataWS = REW`. Thus, it is actually a special case of the new module. So I consolidated the two modules, and now there is just one `stomata_water_stress_linear` module.


yufenghe22:
> 2\. Although it was not written this way, the old `stomata_water_stress_linear` sets `StomataWS = REW`. Thus, it is actually a special case of the new module.

This is a nice unification. Thanks for the changes!

yufenghe22 (approved):
Looking good.

PR #290, opened 2026-04-20
Units added to soil_water_tiledrain.h

dilkaran:
Units added to `soil_water_tiledrain.h` referencing `dssat-csm-os/Soil/SoilWater/TILEDRAIN.for`. Units for parameters related to water content in soil (e.g. water content and field capacity etc) were adopted from `soil_water_downflow.h`.

Unit I assigned to `tile_conductivity` is cm / hr. However, in `TILEDRAIN.for` submodule of DSSAT, two definitions of a parameter `KTILE` (counterpart of `tile_conductivity`) exists.
```
KTILE - Conductivity of tile layer, mm/day
KTILE -  Conductivity of tile drain, cm/day. 
```

In code KTILE has two formulations
```
!      KTILE = SWCN(TDLNO) * HEAD / (SFDRN * 100.)
      KTILE = ETDR * HEAD
```
First equation likely corresponds to first definition as it is property of soil layer and soil conductivity. 

As you can see first equation/code is commented out and BioCro also adopted the latter equation as
`tile_drain_conductivity (tile_conductivity) = tile_drainage_rate * head;`

Thus, I adopted units based on the second definition of `KTILE` in the DSSAT. 

Following is added as TODO comment in my edits as well. However, I wanted to bring it to your attention here as well.
While reviewing `soil_water_flow_functions.cpp` to understand `cumulative_tile_flow` and `total_tile_flow` I found `total_tile_flow` is likely being reset at every time step within function `tile_flow`. In DSSAT, `total_tile_flow` (named `TDFC`) is initialized just once per simulation. Thus, in DSSAT its units are 'cm' but as per my understanding its current formulation in BioCro is effectively, `total_tile_flow` = `cumulative_tile_flow` (tile flow per time step), thus for now the units of `total_tile_flow` are written as cm/hr. Please review tile_flow function and let me know if I am interpreting the syntax incorrectly.


justinmcgrath:
Thanks Dilkaran. One thing to change is not including [water] and [soil] in the units. SI convention is only units are listed. The name quantity name should make it clear what is being referred to, and in the cases here that's mostly clear. To be explicit, the description after the units could say "volumetric water content" etc.

I haven't looked at the code itself, so I don't really know, but why aren't the delta quantities per time?

yufenghe22 (approved):
Looks good to me. Has this been clang formatted?

yufenghe22:
> I haven't looked at the code itself, so I don't really know, but why aren't the delta quantities per time?

It is per time. But in terms of its unit, it should be `cm3/cm3`. The mass balance of water (not rate) is something like this:
$$SW_{t} = SW_{t-1} + \Delta T$$, where $$\Delta T$$ just represents the change of mass per time and it does not need to know what the time step is. The actual rate is baked into other variables like `tile_drainage_rate`

yufenghe22:
> However, I wanted to bring it to your attention here as well.
> While reviewing `soil_water_flow_functions.cpp` to understand `cumulative_tile_flow` and `total_tile_flow` I found `total_tile_flow` is likely being reset at every time step within function `tile_flow`.

Yes. The current `total_tile_flow` is not correctly accumulated because it is initialized to 0 at the beginning of `soil_water_flow_functions.cpp`. `total_tile_flow` is just sum of tile flow from the beginning of model runs. Looks like `total_tile_flow` is a differential module variable. Do we have to create a differential module for it? To avoid that, since it is a diagnostic variable and can be calculated in post-process, we could just remove it

dilkaran:
> Thanks Dilkaran. One thing to change is not including [water] and [soil] in the units. SI convention is only units are listed. The name quantity name should make it clear what is being referred to, and in the cases here that's mostly clear. 

Thanks Justin. I changed the format of units.

dilkaran:
> Looks good to me. Has this been clang formatted?

I didn't apply the clang formatting as Ed said he can do it at once for all files.

dilkaran:
> Yes. The current `total_tile_flow` is not correctly accumulated because it is initialized to 0 at the beginning of `soil_water_flow_functions.cpp`. `total_tile_flow` is just sum of tile flow from the beginning of model runs. Looks like `total_tile_flow` is a differential module variable. Do we have to create a differential module for it? To avoid that, since it is a diagnostic variable and can be calculated in post-process, we could just remove it

Good to know. So for now the units for `total_tile_flow` are correct. We can leave the TODO as the comment to note in this version `total_tile_flow` is not a counter for seasonal flow.

src/module_library/soil_water_tiledrain.h
```diff
@@ -161,62 +162,68 @@ class soil_water_tiledrain : public direct_module
...
 
 string_vector soil_water_tiledrain::get_outputs()
 {
     return {
-        "td_layer_num",          // tile drain layer number
-        "cumulative_tile_flow",  // cm
-        "total_tile_flow",
+        "td_layer_num",          // Dimensionless. Soil layer number containing the tile drain
+        "cumulative_tile_flow",  // cm / hr.  per-timestep tile drain flow; corresponds to DSSAT's TDFD var
+        "total_tile_flow",     // cm / hr. //TODO - I think it currently equals cumulative_tile_flow of
```
eloch216: The unit comments that are provided in the `get_inputs` and `get_outputs` method for each quantity should not be multi-line.

The reason for this is that these comments are used to generate the content of this unit website: https://biocro.org/BioCro-documentation/quantity_docs/quantities.html

In short, there is a script that looks for a comment next to each quantity, and assumes that it contains the units. I don't think it can handle multi-line comments.

Typically these unit comments should be limited to just the units. Some additional info can be included if it is very short. But a long discussion like this should be put somewhere else, either as a comment in the `do_operation` method or in the main doxygen comment that describes the module class.

eloch216:
> > However, I wanted to bring it to your attention here as well.
> > While reviewing `soil_water_flow_functions.cpp` to understand `cumulative_tile_flow` and `total_tile_flow` I found `total_tile_flow` is likely being reset at every time step within function `tile_flow`.
> 
> Yes. The current `total_tile_flow` is not correctly accumulated because it is initialized to 0 at the beginning of `soil_water_flow_functions.cpp`. `total_tile_flow` is just sum of tile flow from the beginning of model runs. Looks like `total_tile_flow` is a differential module variable. Do we have to create a differential module for it? To avoid that, since it is a diagnostic variable and can be calculated in post-process, we could just remove it

Based on this, I would suggest removing `total_tile_flow`. It appears to be numerically identical to `cumulative_tile_flow` so it doesn't really add much.

This also seems like an odd use of the word "cumulative." It might be better to call this the `tile_flow_rate`, since it has rate units and is relevant only for a single time step. Would this make sense to you?

src/module_library/soil_water_tiledrain.h
eloch216: That being said, I think `total_tile_flow` should probably be removed (see my other comment below). So the message here about multi-line comments is more of an "FYI."

dilkaran: `total_tile_flow` has been removed. Yufeng will remove it from `soil_water_flow_functions.cpp`.

dilkaran:
> This also seems like an odd use of the word "cumulative." It might be better to call this the `tile_flow_rate`, since it has rate units and is relevant only for a single time step. Would this make sense to you?

I think `tile_flow_rate` describes the variable well. The authors of DSSAT probably used cumulative because it sums up the contributions of multiple layers in the tile drainage rate. 

@yufenghe22, what do you think about changing the name of the variable?

yufenghe22:
> > This also seems like an odd use of the word "cumulative." It might be better to call this the `tile_flow_rate`, since it has rate units and is relevant only for a single time step. Would this make sense to you?
> 
> I think `tile_flow_rate` describes the variable well. The authors of DSSAT probably used cumulative because it sums up the contributions of multiple layers in the tile drainage rate.
> 
> @yufenghe22, what do you think about changing the name of the variable?

I agree. It's just the `tile flow`. Originally I thought it's accumulated by layers, but it's not.

eloch216:
@dilkaran It looks like there is a test failure because `total_tile_flow` was removed from the `soybean2` model outputs. Of course, this was done intentionally, so it is not a problem. To fix this, it will be necessary to update the stored `soybean2` regression test data. There are instructions for doing this in `tests/testthat/test.CropModels.R`.

Changing the `cumulative_tile_flow` output to `tile_flow_rate` will also cause a similar test failure. I would recommend changing the quantity name, and then updating the stored data so the tests all pass. This way, you will only need to update the stored data once.

dilkaran:
> @dilkaran It looks like there is a test failure because `total_tile_flow` was removed from the `soybean2` model outputs. Of course, this was done intentionally, so it is not a problem. To fix this, it will be necessary to update the stored `soybean2` regression test data. There are instructions for doing this in `tests/testthat/test.CropModels.R`.
> 
> Changing the `cumulative_tile_flow` output to `tile_flow_rate` will also cause a similar test failure. I would recommend changing the quantity name, and then updating the stored data so the tests all pass. This way, you will only need to update the stored data once.

Do we need to edit the `tileDrain_str` in `soil_water_flow_functions.h` simultaneously, or it won't be an issue, and these changes can be made separately?

eloch216:
> > @dilkaran It looks like there is a test failure because `total_tile_flow` was removed from the `soybean2` model outputs. Of course, this was done intentionally, so it is not a problem. To fix this, it will be necessary to update the stored `soybean2` regression test data. There are instructions for doing this in `tests/testthat/test.CropModels.R`.
> > Changing the `cumulative_tile_flow` output to `tile_flow_rate` will also cause a similar test failure. I would recommend changing the quantity name, and then updating the stored data so the tests all pass. This way, you will only need to update the stored data once.
> 
> Do we need to edit the `tileDrain_str` in `soil_water_flow_functions.h` simultaneously, or it won't be an issue, and these changes can be made separately?

I am working on `soil_water_flow_functions.h` and can remove it as part of my PR. It's easier if we don't have overlap between PRs -- then we can avoid having merge conflicts.

eloch216:
@dilkaran This PR was not ready to merge because the checks were failing. There was a compiler error. Now the error has been merged into the `add-new-soil-water-modules`, and BioCro cannot be compiled on that branch.

The error occurred because the `cumulative_tile_flow` element of `tileDrain_str` was not renamed to `tile_flow_rate`. So the line `tileDrain.tile_flow_rate = 0.0;` does not work. This is easy to fix. I will make a new PR for it.

The lesson here is that you should always wait for the checks to complete, even if you think they will pass. We have all made mistakes like this, so it's not a big deal, but please learn from it!

yufenghe22:
> This PR was not ready to merge because the checks were failing.

My bad. I advised to merge because I thought the failing was caused by the test file `soybean2_simulation.csv`. I thought the test files would need to be updated anyway on the `add-new-soil-water-modules `after individual PR is merged. Didn't realize the variables names are incompatible.

eloch216:
> > This PR was not ready to merge because the checks were failing.
> 
> My bad. I advised to merge because I thought the failing was caused by the test file `soybean2_simulation.csv`. I thought the test files would need to be updated anyway on the `add-new-soil-water-modules `after individual PR is merged. Didn't realize the variables names are incompatible.

I think it's better to always keep the checks passing. If we wait for several PRs to be merged and then fix the tests, it would be tough to diagnose the source of a mistake if we find one.

PR #291, opened 2026-04-23
Units added to soil_water_upflow.h

dilkaran:
Units added to soil_water_upflow.h. Also corrected @class tag.

### DIFF
--- a/src/module_library/soil_water_upflow.h
+++ b/src/module_library/soil_water_upflow.h
@@ -8,9 +8,12 @@
 namespace standardBML
 {
 /**
- * @class soil_water_downflow
+ * @class soil_water_upflow
  *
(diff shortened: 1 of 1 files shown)


yufenghe22 (approved):

eloch216:
I think this is almost ready to go. A few requests:

1. There are some places where old code was commented out. These should be deleted.

```
    // if flood <= 0??? // Current depth of flooding (mm)
    // if (surface_runoff <= 0) {
```

and

```
    // } else {
    //     for (int l = 0; l < nlayers; l++){
    //           upFlow.sw_delta_U[l] = 0.0;
    //       }
    // }
```

2. It looks like the `surface_runoff` input is not used anywhere by this module, so it should be removed.

Issue #286, opened 2026-04-13
Documenting units in new soil water modules


eloch216:
The new soil water modules (PR https://github.com/biocro/biocro/pull/241) are almost ready to go. The main barrier to merging that PR is that units are not documented in all the new code. This issue, which is a follow-up to a meeting held earlier today, is to organize our efforts to add the units as fast as possible.

In a separate comment below, I will assign several files to each person. I have tried to assign the "trickier" files to Yufeng and myself, but it's not always possible to tell which ones will be easiest. Before getting to the assignments, I have also provided some instructions for what to do with each file.

I can only assign this to one person, so I will just tag the other people involved: @yufenghe22 @leavesandlemmas @justinmcgrath @leighmatth (I will update once Dilkaran has been added as a collaborator)

Thank you all in advance for your help!

**General Procedure**

To stay organized, let's follow this procedure:
1. Pick one of the files from your list of assignments. Make a new branch based on the `add-new-soil-water-modules` branch, and give it a name that reflects the file you are working on. For example, if you are working on `soil_water_uptake.h`, the new branch would be called `soil-water-uptake-units` or something like that.
2. Add units (more instructions are provided below)
3. Make a PR to merge your new branch back into `add-new-soil-water-modules`.

In other words, as soon as you finish adding units to one file, make a PR for it so it can be discussed and merged.

**Where to Add Units**

For a module class, units must be added in several places:
- The `get_inputs` method
- The `get_outputs` method
- The `do_operation` method

For a free function, units must also be added in several places:
- The input argument list in the function implementation
- (If a function has a declaration and implementation in separate files, it is nice, but not required, to also include the units in the declaration)
- The body of the function

Units should be added as comments following each line of code (or each calculation, if one command spans multiple lines), formatted like `// units`. For example: `// micromol / m^2 / s` or `// dimensionless`. A more thorough description of unit formatting can be found here: https://biocro.org/BioCro-documentation/latest/bookdown/contributing-to-biocro.html#document-units-in-the-code

Adding all these units is very tedious, but historically it has been invaluable for identifying errors and ensuring the equations are easy to follow.

[...]

PR #292, opened 2026-04-23
Soil evaporation units

eloch216:
This PR is for adding units to `soil_evaporation2.h` and `soil_evaporation_functions.h` (https://github.com/biocro/biocro/issues/286).

Here I have reorganized the code and added more documentation in addition to documenting the units. I know my instructions said to not do this. However, the code in these files was extremely confusing, and it was difficult to document the units without getting a better understanding of what the code was supposed to do. So I figured I might as well write down what I learned.

Here are a few reasons for why the units in this code was difficult to understand:

1. The DSSAT code included lines that improperly mixed units together. For example, line 118 in `SOILEV.for` reads `SUMES2 = SUMES2 + ES - WINF`. According to the documentation, `SUMES2` and `WINF` have units of `mm`, while `ES` has units of `mm / day`. So, it's not possible to add these variables together as written.

2. The original publication describing the model (Ritchie 1972) also includes equations that improperly mix units together. For example, the left-hand-side of Equation (8) has units of `mm / day`, while the right-hand-side has units of `mm`. (Side note: this is one of the most important equations in the paper. It's wild to me that its units don't make sense.)

3. The DSSAT code and the original publication indicate conflicting units for some quantities. For example, by comparing the DSSAT code to the flow chart in the paper, it is clear that `WINF` in the code is intended to be the same as $P$ in the paper. However, DSSAT gives units of `mm` for `WINF` while the paper gives units of `mm / day` for $P$.

Most of these issues are related to a failure to distinguish between rates of evaporation (with units of `mm / day` in DSSAT and the paper) and the total amount of evaporation that occurs over a day (with units of `mm`). Thus, they can typically be fixed by explicitly including the time step used. For example, if the DSSAT code above becomes `SUMES2 = SUMES2 + ES * timestep - WINF` where `timestep` equals 1 day, there is no issue because `ES * timestep` has the same units as the other variables.

I did my best to sort out these issues and create a consistent set of units within this module. There are still some parts of the code that I don't really understand. I have left notes about them as comments in the code itself.

While working on this, I noticed that many of the inputs were not actually used for anything, so I got rid of them.

I also decided to change the name of the module from `soil_evaporation2` to `soil_evaporation_ritchie`. I think the `ritchie` suffix better indicates the model being used here, and is consistent with other BioCro modules such as `solar_position_michalsky`.


src/module_library/soil_evaporation_ritchie.h
```diff
@@ -0,0 +1,658 @@
...
+    using std::min;
+
+    // Define conversion constants to avoid magic numbers
+    double constexpr cm_to_mm = 10.0;       // mm / cm
+    double constexpr hours_per_day = 24.0;  // hr / day
+
+    // Hard-coded model parameter values
+    double constexpr canopyHeight = 1.0;  // m
+    double constexpr SWEF_depth = 30;     // cm
+    double constexpr timestep = 1;        // hr
```
yufenghe22, line 362: I think it's nice to add an explicit `timestep` variable here. For non-euler solver, what is the default unit of its $\Delta t$? If its by default `second`, then I wonder if this module would be compatible with non-euler solver by changing the `hours_per_day` to `second_per_day`. In this case, we use `second` constantly across BioCro. But the issue is that we'd need to know the $\Delta t$ from a non-euler solver to feed into the function `ritchie_s1_to_s2`.

src/module_library/soil_evaporation_ritchie.h
```diff
@@ -0,0 +1,658 @@
...
+ *  we have also looked to DSSAT, and our code is largely based on its SOILEV
+ *  subroutine.
+ *
+ *  The DSSAT module cites Ritchie (1972) as its main reference, but there are
+ *  a few differences between the DSSAT code and the equations presented in
+ *  Ritchie (1972). For example, Ritchie (1972) represents the "rainfall or
+ *  irrigation rate" using the symbol `P`, with units of `mm / day`. Comparing
+ *  the equations in Ritchie (1972) against those in SOILEV, it is clear that
+ *  the variable `WINF`, defined as the "potential precipitation for
+ *  infiltration," is intended to be identical to `P`. Yet, the units of `WINF`
```
yufenghe22, line 151: `WINF` should have the unit of `mm/day` or `mm/timestep`. I think the unit note in `SOILEV.for` is incorrect. In [another module](https://github.com/DSSAT/dssat-csm-os/blob/26a327d5cdefb4b732eeefe430d42ec38187b004/SPAM/SPAM.for#L673), it's noted as `mm/day`.

justinmcgrath:
Do these changes alter the results?

I like the new name much more. Name changes are easier to follow when done at the end though, so that it's easier to see the differences.

yufenghe22:
> Do these changes alter the results?

probably not. the added `timestep` is 1 as a multiplier.

src/module_library/soil_evaporation_ritchie.h
```diff
@@ -0,0 +1,658 @@
...
+        "windspeed",                         // m / s
+        "windspeed_height"                   // m
+    };
+}
+
+string_vector soil_evaporation_ritchie::get_outputs()
+{
+    return {
+        "days_stage2",            // day
+        "soil_evaporation_rate",  // mm / hr
```
yufenghe22: Just want to point out that the unit of this evaporation used to be `Mg/ha/hr`.

eloch216:
> Do these changes alter the results?

Nope. All the changes are purely to make the units consistent and make the code easier to follow.

I am sure I didn't change any of the module's behavior because the `soybean2` regression tests all passed without needing to alter the stored simulation results.

src/module_library/soil_evaporation_ritchie.h
eloch216: When? These units have not changed since I copied these modules from the `soil-tipping-bucket` repository: https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/55d110945c41c0f33f0df0aadb7519f5b6014234/src/module_library/soil_evaporation.h#L202

A conversion factor would be needed to convert from `mm / hr` to `Mg / ha / hr`. I think it works out to `1 mm = 10 Mg / ha`.

One hectare is `1e10 mm^2`, so `1 mm` of water over `1 ha` has a volume of `1e10 mm^3 = 1e7 cm^3`. The density of water is approximately `1 g / cm^3`. So this would be a mass of `1e7 g = 10 Mg`. Hence `1 mm` of water is equivalent to `10 Mg / ha`.

Do we need to do this?

src/module_library/soil_evaporation_ritchie.h
eloch216: I indicated `mm / hr` here for this module. It's annoying that DSSAT gives different units in different places.

Before these modules are merged into `develop`, I am going to double check that our units are consistent. This can be done with the `create_quantity_table.R` script in the `script` directory of this repository.

src/module_library/soil_evaporation_ritchie.h
yufenghe22: Sorry. I meant in the old `soil_evaporation.h` module, the unit of the output is `Mg/ha/hr`. This new module's output is in `mm/hr`. Just to make sure when soil evaporation is used in other modules, the units are used correctly, for example, as an input in [here](https://github.com/biocro/biocro/pull/293/changes#diff-f729ea259453fa17ec45ee50827eddf07bc319d556023c61b39482beaa5be84aR164).

yufenghe22: it's probably safer to convert it to `Mg/ha`? I see the soil evaporation is used in several old soil water modules: `bucket_soil_drainage.h`, `cumulative_water_dynamics.h`, `multi_layer_soil_profile.h`, `one_layer_soil_profile.h`, where I assume they all expect `Mg/ha`.

eloch216: Yes, it's absolutely essential that every module using a particular quantity expects it to have the same units. I didn't realize we now had different units floating around for `soil_evaporation_rate`. We will need to double check all quantities for this type of potential issue.

src/module_library/soil_evaporation_ritchie.h
eloch216: > For non-euler solver, what is the default unit of its Δt?

All ODE solvers used in BioCro are unit-agnostic. The only requirement is that the `time` column of the drivers has the same unit as the time basis for derivatives calculated by differential modules. With this convention, one time unit passes whenever the value of `time` increases by 1, but it doesn't matter what this unit actually is.

In our standard module library, our modules all use "per hour" for derivatives, and hence `time` must be in hours.

src/module_library/soil_evaporation_ritchie.h
yufenghe22: OK. I'll modify the other module assuming the `soil_evaporation_rate ` is in unit of `Mg/ha/hr`. And yes, I think you should multiply the output by `10` for the conversion.

src/module_library/soil_evaporation_ritchie.h
eloch216: > I wonder if this module would be compatible with non-euler solver

> we'd need to know the Δt from a non-euler solver to feed into the function `ritchie_s1_to_s2`.

The incompatibility between this module and non-Euler solvers is more severe than simply not knowing the value of $\Delta t$.

In many places, this model assumes that $E(t_{i+1}) = E(t_i) + \Delta E$ for $\Delta E = \frac{dE}{dt}|_{t=t_i} \cdot \Delta t$, where:
- $E$ is the total water evaporated from the soil surface
- $\frac{dE}{dt}|_{t=t_i}$ is the instantaneous soil evaporation rate at time $t_i$
- $t_i$ is the current time
- $t_{i+1}$ is the next time
- $\Delta t$ is the time step
 
This is only true for an Euler solver. Even for a second-order Runge-Kutta solver (the next simplest solver after Euler),  $E(t_{i+1}) \neq E(t_i) + \frac{dE}{dt}|_{t=t_i} \cdot \Delta t$ in general. So, for any other solver, the assumptions of this model will not be met.

For adaptive solvers, things are even more complicated because there is not necessarily a well-defined time step.

yufenghe22: I don't know much about how boost solvers work, but looks like there's this [controlled stepper for the boost solver ](https://www.boost.org/doc/libs/latest/libs/numeric/odeint/doc/html/boost_numeric_odeint/concepts/controlled_stepper.html), and it has this `dt` quantity. If we can track the `dt` for any solvers, feeding that `dt` into the `ritchie_s1_to_s2` should make this module adaptable to non-euler solvers.

yufenghe22: okay . just saw your last reply. I'll think it more.

eloch216: > If we can track the `dt` for any solvers, feeding that `dt` into the `ritchie_s1_to_s2` should make this module adaptable to non-euler solvers.

We will not be doing that because it would fundamentally mix together the model's equations and the ODE method used to solve them.

The problem here is that this model is was designed in a way that mixes its equations with the ODE method used to solve them. BioCro is not really intended to represent such models.

Plus, as mentioned above, knowing the time step would not address the more fundamental issue that $E(t_{i+1}) \neq E(t_i) + \frac{dE}{dt}|_{t=t_i} \cdot \Delta t$ when using a non-Euler solver.

src/module_library/soil_evaporation_ritchie.h
eloch216: I'll change it. I agree it makes more sense to use `Mg / ha / hr`.

eloch216:
I changed the `soil_evaporation_rate` units to `Mg / ha / hr`. Of course, this changed the module behavior, and the `soybean2` regression test needed to be updated.

The changes in the `units_multi_layer_soil_profile` branch should "counteract" these changes in the sense that the soil water content values should revert to what they were before.

Maybe it would make sense to merge that branch into this one so we can confirm?

yufenghe22:
> I changed the `soil_evaporation_rate` units to `Mg / ha / hr`. Of course, this changed the module behavior, and the `soybean2` regression test needed to be updated.

Why would it alter results? I treated its unit as `mm/hr` before and made conversion as needed. Now we just use a different unit.

eloch216:
> > I changed the `soil_evaporation_rate` units to `Mg / ha / hr`. Of course, this changed the module behavior, and the `soybean2` regression test needed to be updated.
> 
> Why would it alter results? I treated its unit as `mm/hr` before and made conversion as needed. Now we just use a different unit.

Is that conversion done in the `multi_layer_soil_profile` module? It would need to be removed now. I didn't change that module since I didn't want to create a merge conflict with your branch for that module.

I can try a different approach.

yufenghe22:
> Is that conversion done in the `multi_layer_soil_profile` module? It would need to be removed now. I didn't change that module since I didn't want to create a merge conflict with your branch for that module.

Yes. There's been a conversion in the `multi_layer_soil_profile` module. I will change that to match the unit. You can make changes here as needed.

> Maybe it would make sense to merge that branch into this one so we can confirm?

Would you like to the merge that to here? Looks like I cannot just change the target branch, so I guess I will start a new PR

eloch216:
> > Is that conversion done in the `multi_layer_soil_profile` module? It would need to be removed now. I didn't change that module since I didn't want to create a merge conflict with your branch for that module.
> 
> Yes. There's been a conversion in the `multi_layer_soil_profile` module. I will change that to match the unit. You can make changes here as needed.
> 
> > Maybe it would make sense to merge that branch into this one so we can confirm?
> 
> Would you like to the merge that to here? Looks like I cannot just change the target branch, so I guess I will start a new PR

Yeah, I think it might be simplest to merge that to this branch. That's weird that the target can't be changed. But making a new PR is easy.

yufenghe22:
> Yeah, I think it might be simplest to merge that to this branch. That's weird that the target can't be changed. But making a new PR is easy.

Could it be permission related? It should have an `edit` button near the branch names, but I did not see it. Anyway, I've opened a new PR and implemented a conversion based on the soil evaporation's unit as `Mg/ha/hr`.

eloch216:
I confirmed that the only change to `tests/testdata/soybean2_simulation.csv` is that `soil_evaporation_rate` is higher by a factor of 10, due to the change in units from `mm / hr` to `Mg / ha / hr`. So I think this PR can be merged.

PR #294, opened 2026-04-24
add Units for multi layer soil profile

yufenghe22:
This module is homemade, not based on any DSSAT modules. I've noted this in the module description.
I've applied clang format to the file, and it changed the array's alignment for some reason, probably due to the unit comment I added.


src/module_library/multi_layer_soil_profile.h
```diff
@@ -159,16 +161,14 @@ class multi_layer_soil_profile : public differential_module
...
-        "deltaT_1",
-        "uptake_layer_1",
+        "soil_depth_1",          // cm
+        "soil_water_content_1",  // cm^3/cm^3
+        "deltaS_1",              // cm^3/cm^3/hr
+        "deltaU_1",              // cm^3/cm^3/hr
+        "deltaT_1",              // cm^3/cm^3/hr
+        "uptake_layer_1",        //  Mg/ha/hr
 
         "soil_depth_2",
```
eloch216: Units need to be specified for every item in `get_inputs` and `get_outputs`, even if it is repetitive. This is because the unit comments are used to build the unit database available at https://biocro.org/BioCro-documentation/quantity_docs/quantities.html.

src/module_library/multi_layer_soil_profile.h
```diff
@@ -221,68 +221,67 @@ string_vector multi_layer_soil_profile::get_outputs()
...
+    constexpr double cm_to_mm = 10.0;
+    constexpr double MG_HA_to_mm = 0.1;  // Mg/ha of water = 0.1 mm
     // Calculate total change in soil water content
-    double delta_soil_water_content[nlayers];
+    double delta_soil_water_content[nlayers];  // cm^3/cm^3/hr
     for (int l = 0; l < nlayers; l++) {
         // adding uptake because value is negative
-        delta_soil_water_content[l] = swdeltS[l] + swdeltU[l] + swdeltT[l] + (uptake[l] / (100 * soil_depth[l]));
+        delta_soil_water_content[l] = swdeltS[l] + swdeltU[l] + swdeltT[l] + (uptake[l] * MG_HA_to_mm / (cm_to_mm * soil_depth[l]));
         if (soil_water_content[l] + delta_soil_water_content[l] < 0) {
```
eloch216, line 276: If `soil_water_content` has units `cm^3 / cm^3` and `delta_soil_water_content` has units `cm^3 / cm^3 / hr`, these quantities cannot be compared. There is probably an implicit multiplication by the `timestep` here.

src/module_library/multi_layer_soil_profile.h
```diff
@@ -221,68 +221,67 @@ string_vector multi_layer_soil_profile::get_outputs()
...
+
+    constexpr double cm_to_mm = 10.0;
+    constexpr double MG_HA_to_mm = 0.1;  // Mg/ha of water = 0.1 mm
     // Calculate total change in soil water content
-    double delta_soil_water_content[nlayers];
+    double delta_soil_water_content[nlayers];  // cm^3/cm^3/hr
     for (int l = 0; l < nlayers; l++) {
         // adding uptake because value is negative
-        delta_soil_water_content[l] = swdeltS[l] + swdeltU[l] + swdeltT[l] + (uptake[l] / (100 * soil_depth[l]));
+        delta_soil_water_content[l] = swdeltS[l] + swdeltU[l] + swdeltT[l] + (uptake[l] * MG_HA_to_mm / (cm_to_mm * soil_depth[l]));
```
eloch216: Units need to be specified here

src/module_library/multi_layer_soil_profile.h
```diff
@@ -221,68 +221,67 @@ string_vector multi_layer_soil_profile::get_outputs()
...
 
-    // Mg/ha/hr
-    // double soil_evap = soil_evaporation_rate/ 10.0; // Mg/ha/hr to mm/hr
-    // remove soil evaporation from first layer, but don't let water content go
-    // negative. This is a crude, and hopefully temporary, fix. -mlm
-    // soil_evaporation_rate from the new soil evapo function is in mm/hr? -YH
-    delta_soil_water_content[0] = delta_soil_water_content[0] - soil_evaporation_rate / (10.0 * soil_depth[0]);
+    // convert evaporation to water cotent: (E(Mg/ha/hr)*MG_HA_to_mm)/depth(mm)
+    const double evaporated_water_cotent = (soil_evaporation_rate * MG_HA_to_mm) / (soil_depth[0] * cm_to_mm);  // cm^3 / cm^3 / hr
+    delta_soil_water_content[0] = delta_soil_water_content[0] - evaporated_water_cotent;
```
eloch216, line 283: Units need to be specified here

eloch216:
I think I am still unclear on the units of `deltaS`, `deltaU`, etc.

In the tile drain module (and some other modules), they are given as `cm^3 / cm^3` (https://github.com/biocro/biocro/pull/290), but here they are given as `cm^3 / cm^3 / hr`. Which are they?

From this comment (https://github.com/biocro/biocro/pull/290#issuecomment-4289097916), it sounds like they really are rates of change of volumetric soil water content with units of `cm^3 / cm^3 / hr = hr^(-1)`. But in many places in the code, they are used as if they were changes of volumetric soil water content with units of `cm^3 / cm^3 = dimensionless`. In these cases there is an implicit multiplication by a time step of 1 hour.

yufenghe22:
> From this comment ([#290 (comment)](https://github.com/biocro/biocro/pull/290#issuecomment-4289097916)), it sounds like they really are rates of change of volumetric soil water content with units of `cm^3 / cm^3 / hr = hr^(-1)`.

No. I think these delta quantities are incremental change in state with unit `cm^3 / cm^3`. And I noted their units incorrectly here (I should remove the `/hr`). The time step part should be baked into how these deltas are calculated. For example, `deltaS` is an output of `soil_water_downflow.h`, which comes from the function `infil` of `soil_water_flow_functions.cpp`. In there, for example, there's this mass equation: `swtemp[l] = swtemp[l] + (potential_infiltration - drn[l]) / soil_depth[l];`. `drn` is a rate and it should be multiplied by `timestep` of `1`. Because we changed the rates to `per hour`, this implicit multiplication of `timestep` is often omitted, but I think it's better to add it explicitly like what we did with the `soil evaporation` module.

src/module_library/multi_layer_soil_profile.h
yufenghe22: done

src/module_library/multi_layer_soil_profile.h
yufenghe22: yes. for those rates (uptake and soil_evaporation_rate), I multiply a `timestep` of 1 to them.

src/module_library/multi_layer_soil_profile.h
yufenghe22: done

eloch216 (approved):
Thanks!

PR #296, opened 2026-04-29
add units to dynamic_rooting

yufenghe22:
Added units to the soil_water_dynamic_rooting.h module.


yufenghe22:
@justinmcgrath I'm also testing to use the old `rsdf*Root` to estimate rooting depth. Using `rsdf of 0.44` and it looks like the root can quickly have access top 5 layers (75 cm in depth) by the middle of vegetative stage. Of course, the actual water uptake is weighted by [an exponential decay function](https://github.com/biocro/biocro/blob/a8bb454d07d771d7821444bf19fbf18b6cf9e5e0/src/module_library/soil_water_uptake.h#L237) that weights more on top layers for water uptake.
image
image

justinmcgrath:
I don't remember exactly why this came up - something about the new model basing the depth solely on time and not root mass?

yufenghe22:
> I don't remember exactly why this came up - something about the new model basing the depth solely on time and not root mass?

Yes. At first, we use a fixed number for `max_rooting_layer`, for example, 4. But I found that this does not give a good result in early season's soil water content. 
However, I think maybe I did not have a proper exponential decay function at that time. Overall, this time-dependent of `max_rooting_layer` is too crude. I'm thinking to go back to the constant one. There's not much difference in soil water content between this "dynamic" one and a constant. See comparison below:
image
image

dilkaran (approved):
Unit edits look correct to me.

eloch216:
> @justinmcgrath I'm also testing to use the old `rsdf*Root` to estimate rooting depth. Using `rsdf of 0.44` and it looks like the root can quickly have access top 5 layers (75 cm in depth) by the middle of vegetative stage.

> > I don't remember exactly why this came up - something about the new model basing the depth solely on time and not root mass?
> 
> Yes. At first, we use a fixed number for `max_rooting_layer`, for example, 4. But I found that this does not give a good result in early season's soil water content. However, I think maybe I did not have a proper exponential decay function at that time. Overall, this time-dependent of `max_rooting_layer` is too crude. I'm thinking to go back to the constant one. There's not much difference in soil water content between this "dynamic" one and a constant.

When using the `rsdf * Root` method, does the predicted soil water content agree with the measured water content early in the season?

Even if there isn't much difference in the results for these particular simulations, I think it would still be preferable to base the rooting depth on the root mass. Conceptually this makes more sense than using a constant value or basing it only on the development index. E.g., using a constant value of 4 would mean that the roots can access water up to 50 cm deep as soon as the seeds germinate, which is not possible for a real plant.

yufenghe22:
> When using the `rsdf * Root` method, does the predicted soil water content agree with the measured water content early in the season?

This would introduce much more (unrealistic?) variation in early season as root is only able to access the surface layer soil water. Surface soil water can fluctuate a lot as it's exposed for evaporation, leading to a 0 StomataWS (complete dry surface soil). This causes much lower root biomasses due to the impacts during the emergence period.

In my segment function, I started with layer 3 instead of 1 to smooth things out. Maybe we can assume the root can at least have access to the top 2 layers instead of purely based on the `rsdf * Root`?

In these tests, `dynamic` means using the `rsdf*Root` method and `constant` means using a constant rooting depth of 5.
image
image

eloch216:
> Maybe we can assume the root can at least have access to the top 2 layers instead of purely based on the `rsdf * Root`?

If this works well, it would be my preference. The depth could be calculated as `std::min(min_root_depth, rsdf * Root)`, where `min_root_depth` is an input parameter. Then `min_root_depth` could be chosen to ensure `max_rooting_layer` is always 2 or above. Having it as an input parameter would make it easier to adjust in the future.

yufenghe22:
Using a dynamic method with `std::max(min_root_depth, rsdf * Root)` or `dynamic-Biomass` in the plots generates similar result as the the case with constant `max_rooting_layer` of 5. `dynamic-DVI` differs more from the two. I think the `dynamic-Biomass` looks good to use. Just pushed the changes.
image
image

Using a `min_root_depht` of 2, here is how the `max_rooting_layer` changes over time:
image

yufenghe22:
looks like the testing is failing because my base is not up-to-date with `[add-new-soil-water-modules]`. Do you want me to merge the base branch first? Because we work on units on individual branch and may not update base branch in time, which would cause such issue

> The `soybean2` simulation result is missing required columns from the saved result: cumulative_tile_flow, potential_infiltration, total_tile_flow.

eloch216:
> looks like the testing is failing because my base is not up-to-date with `[add-new-soil-water-modules]`. Do you want me to merge the base branch first? Because we work on units on individual branch and may not update base branch in time, which would cause such issue
> 
> > The `soybean2` simulation result is missing required columns from the saved result: cumulative_tile_flow, potential_infiltration, total_tile_flow.

Yes, please update this branch from `add-new-soil-water-modules` and then update tests as necessary so they pass. Here it is expected that a change in the test data will be required, because some of the module behavior has changed.

eloch216:
It's confusing to me that `min_root_depth` is a layer number, not a depth. I think this should be "rephrased" so `min_root_depth` is a depth in `cm`.

In that case, line 107 would become

```c++
double rootDepth = std::min(min_root_depth, rsdf * Root * m_to_cm); // cm
```

Then there is no need for line 124:

```c++
layers_count = std::max(layers_count,static_cast<int>(min_root_depth));
```

eloch216:
It looks like `soil_depth_1` and `soil_depth_2` are 5 and 10 cm, respectively. So if `min_root_depth` is greater than 5 cm and less than 15 cm, I believe the minimum rooting layer would be layer 2.

yufenghe22:
Looks like there's a documentation fail during the GitHub tests. Can we ignore it?

```
Run tar -czf biocro-docs-from-$VERSION.tgz -C $BIOCRO_DOCUMENTATION_ROOT/$VERSION .
tar: biocro_documentation_root/296_merge: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
Error: Process completed with exit code 2.
```

eloch216:
> Looks like there's a documentation fail during the GitHub tests. Can we ignore it?
> 
> ```
> Run tar -czf biocro-docs-from-$VERSION.tgz -C $BIOCRO_DOCUMENTATION_ROOT/$VERSION .
> tar: biocro_documentation_root/296_merge: Cannot open: No such file or directory
> tar: Error is not recoverable: exiting now
> Error: Process completed with exit code 2.
> ```

Yeah, there seems to be an issue with the `tinytex` LaTeX distribution that's preventing the vignettes from being built. For the moment, we can ignore documentation failures. This should resolve on its own in a few days, I believe.

yufenghe22:
> > Looks like there's a documentation fail during the GitHub tests. Can we ignore it?
> > ```
> > Run tar -czf biocro-docs-from-$VERSION.tgz -C $BIOCRO_DOCUMENTATION_ROOT/$VERSION .
> > tar: biocro_documentation_root/296_merge: Cannot open: No such file or directory
> > tar: Error is not recoverable: exiting now
> > Error: Process completed with exit code 2.
> > ```
> 
> Yeah, there seems to be an issue with the `tinytex` LaTeX distribution that's preventing the vignettes from being built. For the moment, we can ignore documentation failures. This should resolve on its own in a few days, I believe.

Is this good to go? @eloch216

eloch216 (approved):

PR #297, opened 2026-04-29
add units to soil_water_uptake.h

yufenghe22:
This module had a PR before, so the changes are minor here.

### DIFF
--- a/src/module_library/soil_water_uptake.h
+++ b/src/module_library/soil_water_uptake.h
@@ -7,6 +7,15 @@
 
 namespace standardBML
 {
+/**
+ *  @class soil_water_uptake
+ *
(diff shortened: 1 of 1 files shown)


dilkaran:
I think it would be better to write cm^3 instead of cm3. My understanding is that each unit has to have one space character between them. So Mg/ha/hr would be Mg / ha / hr.

eloch216:
Besides Dilkaran's comment above, my only request is that the comment line saying `// uptake in layer = ET*(depth of layer/total depth of root zone)` should either be removed or updated, because it is not correct anymore.

yufenghe22:
Okay. Both problems have been addressed

dilkaran (approved):

eloch216 (approved):

PR #298, opened 2026-04-29
updated units; removed unused code

leavesandlemmas:
My assigned contribution for Issue #286


src/module_library/soil_surface_runoff.h
```diff
@@ -106,52 +106,53 @@ class soil_surface_runoff : public direct_module
...
 }
 
 void soil_surface_runoff::do_operation() const
 {
     int nlayers = 6;
-    double available_water = precip + irrigation;                // mm
-    double soil_storage = 254.0 * (100.0 / curve_number - 1.0);  // in to mm
+    double available_water = precip + irrigation;                // mm / hr
+    double constexpr mm_per_inch = 254;
+    double soil_storage = mm_per_inch * (100.0 / curve_number - 1.0);  // mm / hr
```
yufenghe22: I think the storage should be in `mm`. See definition [here](https://www.hec.usace.army.mil/confluence/rasdocs/ras1dtechref/6.4/overview-of-optional-capabilities/modeling-precipitation-and-infiltration/curve-number#:~:text=%E2%88%9210-,where,is%20in%20inches.,-The%20curve%20number). This means the other rates like `precip` should be converted to `mass` first by multiplying `timestep`. The final `surface_runoff` needs to be calculated by dividing mass by `timestep`

src/module_library/soil_type_selector.h
```diff
@@ -257,82 +257,65 @@ class soil_type_selector : public direct_module
...
+        "soil_saturated_conductivity_1", // kg s / m^3
+        "soil_saturation_capacity_1",    // m^3 / m^3
+        "soil_field_capacity_1",  // m^3 / m^3
+        "soil_wilting_point_1", // m^3 / m^3
+        "soil_bulk_density_1", // Mg / m^3
+
+        "soil_silt_content_2", // unitless
+        "soil_clay_content_2", // unitless
+        "soil_saturated_conductivity_2", // kg s / m^3
+        "soil_saturation_capacity_2",  //
```
yufenghe22: missing one here

src/module_library/surface_excess_water.h
yufenghe22, line 1: this module can be deleted. It was for test and is not used anywhere.

src/module_library/soil_surface_runoff.h
eloch216: I looked into the `soil_surface_runoff` module a bit more and learned a few things.

1. It is based off the `RNOFF` subroutine from DSSAT (https://github.com/DSSAT/dssat-csm-os/blob/develop/Soil/SoilWater/RNOFF.for)
2. The DSSAT code implements the "Williams-SCS curve number technique." I found a nice reference for this: Williams et al. (2011) (https://doi.org/10.1061/(ASCE)HE.1943-5584.0000529)

Equation 2 in Williams et al. (2011) and the DSSAT code documentation both indicate that our `soil_storage` variable should be in units of `mm`. So I agree with Yufeng's comment.

Scott, can you change the unit as suggested by Yufeng, and also include the two references above in the documentation for this module?

src/module_library/soil_type_selector.h
```diff
@@ -257,82 +257,65 @@ class soil_type_selector : public direct_module
...
+        "soil_type_indicator_5", // unitless int
+        "soil_type_indicator_6"}; // unitless int
 }
 
 string_vector soil_type_selector::get_outputs()
 {
     return {
-        "soil_silt_content_1",
-        "soil_clay_content_1",
-        // "soil_sand_content_1",
```
eloch216, line 273: Thanks for removing the commented-out elements from `get_outputs()`. It looks like there are still corresponding commented-out entries in the module class initializer list (e.g. `// soil_sand_content_1_op(get_op(output_quantities, "soil_sand_content_1")),`) and the module class member declarations (e.g. `// double* soil_sand_content_1_op;`). Can you also remove these?

src/module_library/soil_surface_runoff.h
eloch216: Just noticed that Yufeng shared a reference too. That one could also be added to the module documentation.

src/module_library/soil_surface_runoff.h
```diff
@@ -50,6 +56,8 @@ class soil_surface_runoff : public direct_module
           soil_wilting_point_6{get_input(input_quantities, "soil_wilting_point_6")},
           soil_saturation_capacity_6{get_input(input_quantities, "soil_saturation_capacity_6")},
 
+          timestep{get_input(input_quantities, "timestep")},
```
eloch216: This `timestep` is the time spacing of the `drivers`, not the time step being used by the ODE solver, so it is not appropriate to use it here. See this comment: https://github.com/biocro/biocro/pull/300#discussion_r3174305421

Instead, the module should include a hard-coded timestep parameter. We have done this for a few other DSSAT modules. For example: https://github.com/biocro/biocro/blob/a24001040a767882b2e1f0c34b6141dfe69e2f5f/src/module_library/soil_water_downflow.h#L246

src/module_library/soil_surface_runoff.h
eloch216: Units need to be documented here

eloch216:
The units here are still not quite right. The `soil_storage` variable should be in `mm`, not `mm / hr`. It has been changed in `get_outputs`, but not in the code inside `do_operation`. 

Because of this change in units, a few other later calculations also need to change.

1. If `pb` is the "threshold amount of rainfall that can occur before runoff starts," it should be in `mm` as well. So it should probably be calculated as `double pb = available_water * timestep - soil_initial_abstraction * soil_storage;` because `available_water` is a rate in `mm / hr`.

2. The calculation of `surface_runoff` uses Equation 1 from Williams et al. (2012). It gives a total amount of runoff in `mm`. This needs to be documented in the code.

3. The denominator of this equation also needs to become `available_water * timestep + (1.0 - soil_initial_abstraction) * soil_storage` for the units to work out.

Please double check that all units are documented and consistent with each other.

src/module_library/soil_surface_runoff.h
leavesandlemmas: Is it not m^3 / m^3?

eloch216: It is. But that needs to be documented. Every line where something is calculated should indicate the units, even if they seem obvious to you.

src/module_library/soil_surface_runoff.h
leavesandlemmas: Make up yall's mind. I originally had a hardcoded conversion .....  Is it 1 hr  or 24 hr per day?

eloch216: > Make up yall's mind. I originally had a hardcoded conversion ..... Is it 1 hr or 24 hr per day?

Those are just two different types of conversion.

Sometimes there are constants in an equation whose values were originally given in units involving days. These need to be converted to hourly values using `hours_per_day`.

In other places, there are comparisons or additions that improperly combine rates and amounts. For example, a variable in `cm` being added to one in `cm / hr`. In these cases, we need to multiply or divide by the timestep as necessary to get the units to agree.

In this module, I don't believe there are any conversions from a day basis to an hour basis, but there are places where the time step is needed. And as described above, its value (1 hour) must be hardcoded.

eloch216:
Why was `tests/testthat.R` modified?

justinmcgrath:
This is not related to this issue, but I forgot to mention this at the meeting today. There's a severe vulnerability in Linux. If you use Linux, check whether your distribution has fixed it and update ASAP. The initial handling was bad, so distributions are handling this at different speeds.

eloch216:
This looks good, except the changes to `tests/testthat.R` which seem unrelated to the subject of this PR.

yufenghe22 (approved):

eloch216:
In the interest of time, I'll just go ahead and merge this now

PR #300, opened 2026-05-01
Infil and satflo units

eloch216:
There are four functions in `soil_water_flow`. Rather than waiting until I am done with all of them, I thought it might make sense to make a PR for just the first two while I work on the next ones.

Besides documenting units, I also renamed a few things. For example, I found it confusing to have variables named `drn`, `drain`, and `drainage_rate` within a single function -- it was hard to remember which one is which.

I also moved the calculation of `potential_infiltration` inside the `infil` function instead of calculating it in the `soil_water_downflow` module. I think it makes more sense there.

Finally, I tried to clean up the code a bit. There were many weird things, such as reassigning one variable to another name for no reason (`double currentswcon = swcon;`). Or initializing a variable to 0 and then resetting its value to something else right away. All of this made it more difficult to follow.

There are changes to the stored test data, but they are very minor. They are due to (1) removing `potential_infiltration` as a module output and (2) converting `excess` from `cm / hr` to `mm / hr` (to make it more consistent with other rates such as precipitation rate, total drainage rate, infiltration rate, etc).


eloch216:
Overall, I would say I am a bit puzzled about the design of the `infil` and `satflo` functions. Offhand, I would have expected that the `infil` function could still be used when `infiltration_rate` is zero. But the `satflo` function is definitely not just a special case of `infil`. It seems to follow a different type of reasoning, and I have not been able to find a publication that describes it. (Although I have not looked very hard.)

leavesandlemmas:
I tried to follow the logic from the code when working on #245  and I gave up. It was going to take too much time.

src/module_library/soil_water_downflow.h
```diff
@@ -220,29 +218,30 @@ string_vector soil_water_downflow::get_inputs()
...
+        "drn_4",  // cm / hr. Drainage rate through soil layer 4
+        "drn_5",  // cm / hr. Drainage rate through soil layer 5
+        "drn_6"   // cm / hr. Drainage rate through soil layer 6
     };
 }
 
 void soil_water_downflow::do_operation() const
 {
+    // Define hard-coded parameter values
+    double constexpr timestep = 1.0;  // hr
```
yufenghe22: There is a `timestep` input, should we use that?

src/module_library/soil_water_flow_functions.h
```diff
@@ -7,11 +7,10 @@ static int const max_soil_layers = 10;
...
 struct infilWater_str {
-    double drain;                        // Drainage rate from soil profile (mm/hr)
-    double drn[max_soil_layers];         // Drainage rate through soil layer l (cm/hr)
-    double excess_water;                 // Excess water to be added to runoff (cm/hr)
-    double sw_delta_S[max_soil_layers];  // Change in soil water content due to drainage in layer l
-                                         // (cm3 [water] / cm3 [soil])
+    double overall_drainage_rate;           // mm / hr       - Drainage rate from the lowest soil layer
+    double downward_flux[max_soil_layers];  // cm / hr       - Rate of downward water flow out of this layer
+    double excess_water_rate;               // mm / hr       - Rate of excess water appearing at soil surface
+    double sw_delta_S[max_soil_layers];     // dimensionless - Change in soil water content due to downflow
```
yufenghe22: should be `cm^3/cm^3`

src/module_library/soil_water_flow_functions.cpp
```diff
@@ -3,208 +3,410 @@
...
-        // double hold = (soil_saturation_capacity[l] - swtemp[l]) * soil_depth[l];
-        if (swtemp[l] >= soil_field_capacity[l] + 0.003) {
-            // Rprintf("swtemp[%i] > soil_field_capacity_%i. \n", l,l);
-            drmx[l] = (swtemp[l] - soil_field_capacity[l]) * currentswcon * soil_depth[l];
-            drmx[l] = std::max(0.0, drmx[l]);
-        }
+        // Drainage will occur if the soil water content in this layer is above
+        // the field capacity
+        double const drainage_rate =
+            swtemp[l] >= soil_field_capacity[l] + 0.003
```
yufenghe22: 0.003 is like an epsilon, very small number?

src/module_library/soil_water_flow_functions.cpp
```diff
@@ -3,208 +3,410 @@
...
+            // its saturation capacity, and water in excess of the holding
+            // capacity is passed directly to the layer below.
+
+            // Calculate the drainage rate that would occur when the soil layer
+            // is fully saturated, in the absence of any infiltration
+            drainage_rate = swcon * (soil_saturation_capacity[l] - soil_field_capacity[l]) * soil_depth[l];  // cm / hr
+
+            // Reduce the soil water conductivity constant in the top layer to
+            // allow increased evaporation from wet soil surfaces
             if (l == 0) {
```
yufenghe22, line 126: when `l==0`, what does it do? Looks like the code is deleted.

yufenghe22:
>It seems to follow a different type of reasoning, and I have not been able to find a publication that describes it. (Although I have not looked very hard.)

The `INFIL` part can be found on page 45 in [here](https://link-springer-com.proxy2.library.illinois.edu/chapter/10.1007/978-94-017-3624-4_3). I also have not found reference to the `SATFLO`.

src/module_library/soil_water_downflow.h
eloch216: Unfortunately there is still some potentially unclear terminology in the time-related parts of BioCro. The input parameter `timestep` is the time step (in hours) used in the `drivers` input. It is not necessarily the time step being used by the ODE solver. There is no way for a module to know the time step being used by the ODE solver.

In principle, the `timestep` parameter requirement could be removed because it's possible to determine its value directly from the drivers themselves. Eventually it will be removed, but it's not a priority right now.

src/module_library/soil_water_flow_functions.h
eloch216: Okay, I can change these to be more specific about the origin of the dimensionless value. But, since it is a ratio, there is no need to include any SI prefixes. They can all just be `m^3 / m^3`.

src/module_library/soil_water_flow_functions.cpp
eloch216: Good point, I could define this as an `epsilon` instead of leaving it as a magic number.

src/module_library/soil_water_flow_functions.cpp
eloch216: The original code was written as

```c++
double drainage_rate;

// Some intervening code

if (l == 0) {
    // reduce SWCON in top layer to allow for increased evaporation for wet soils
    drainage_rate = 0.9 * currentswcon * (soil_saturation_capacity[l] - soil_field_capacity[l]) * soil_depth[l];  // cm/hr
} else {
    drainage_rate = currentswcon * (soil_saturation_capacity[l] - soil_field_capacity[l]) * soil_depth[l];
}
```

I wanted to reduce repitition here and clarify that the difference between the top layer and the other layers is that the drainage rate from the top layer is reduced. So I rewrote it as

```c++
double constexpr swconrf = 0.9;   // dimensionless - swcon reduction factor

// Some intervening code

double drainage_rate = 0.0;  // cm / hr

// Some intervening code

drainage_rate = swcon * (soil_saturation_capacity[l] - soil_field_capacity[l]) * soil_depth[l];  // cm / hr

if (l == 0) {
    drainage_rate = swconrf * drainage_rate;  // cm / hr
}
```

This is equivalent but has less repetition.

eloch216:
> > It seems to follow a different type of reasoning, and I have not been able to find a publication that describes it. (Although I have not looked very hard.)
> 
> The `INFIL` part can be found on page 45 in [here](https://link-springer-com.proxy2.library.illinois.edu/chapter/10.1007/978-94-017-3624-4_3). I also have not found reference to the `SATFLO`.

I tried looking around a little bit more. The best I found is this paper: https://doi.org/10.1097/01.ss.0000148740.90616.fd. I don't think it describes the model directly, but it contains several references, and one of them might. I think I will add this to the `satflo` documentation as a possible reference.

BTW, the source in your link is referenced in the documentation for `infil`.

src/module_library/soil_water_flow_functions.cpp
yufenghe22: Thanks. This makes sense

yufenghe22 (approved):

eloch216:
I guess the weird documentation error has resolved itself!

PR #301, opened 2026-05-01
Fix bug in CanAC 

leavesandlemmas:
While refactoring code I found a bug in the updates from #274. In `CanAC`, the lambda function `func_direct` passes `direct_gsw_estimate` to `leaf_energy_balance` rather than the updated `gs` value. This PR fixes that bug.

I would not have noticed it (and clearly none of us did) but I was running the code through an AI for a different purpose, and it flagged this bug.


leavesandlemmas:
failed check is unrelated to tests or code changes. something to do with latex installation

yufenghe22 (approved):
AI seems good at understanding program's logic and identifying bugs.

eloch216 (approved):
Good catch! Nice to see AI doing something useful. I think you can ignore the documentation check failure and merge this whenever you'd like.

PR #303, opened 2026-05-04
Upflow tileflow units

eloch216:
I think this is the last unit-related PR to be made for the new soil water modules. It's for the `up_flow` and `tile_flow` functions.


src/module_library/soil_water_flow_functions.cpp
```diff
@@ -426,227 +426,293 @@ infilWater_str satflo(
...
+
+        double const thet2 =
+            std::max(0.0, std::min(swtemp[m] - soil_wilting_point[m], esw[m]));  // m^3 / m^3
+
+        double const normalized_wc =
+            0.5 * (thet1 * soil_depth[l] + thet2 * soil_depth[m]) /
+            (soil_depth[l] + soil_depth[m]);  // m^3 / m^3
+
+        double const dbar =
+            std::min(max_dbar_hr, soil_diffusivity_hr * exp(35.4 * normalized_wc));  // cm / hr
```
yufenghe22: The slope `35.4` is a fitted parameter based on Rose (1968). At the exponential, the term $\exp(35.4 \times \theta)$ grows very quickly. 
```
Rose D A (1968) Water movement in porous materials III. Evaporation of water from soil. British
Journal Applied Physics Series 2, Volume 1:1779-1791.
```
Image

yufenghe22 (approved):

src/module_library/soil_water_flow_functions.cpp
yufenghe22: The above screenshot also shows where the `0.5` comes from. Source: Pg 49 in here (https://doi.org/10.1007/978-94-017-3624-4_3). This is already in the `source` reference.

yufenghe22: Otherwise, this looks good to me.

src/module_library/soil_water_flow_functions.cpp
```diff
@@ -426,227 +426,293 @@ infilWater_str satflo(
...
 
-        // Redistribute water from upper layers
-        // Assume that water is limited by user-specified tile drainage rate
-        // rather than by each layer's Ksat.
-        double drn_total = 0.0;  // cm/hr
-        double swdeltT_total = 0.0;
+        // Redistribute water from upper layers. Assume that water is limited by
+        // user-specified tile drainage rate rather than by each layer's Ksat.
+        double drn_total = 0.0;  // cm
         double excess = 0.0;
```
dilkaran: Does this line (`excess`) need a units comment?

eloch216: Whoops, yes it does!

src/module_library/soil_water_flow_functions.cpp
eloch216: I will add this as a reference.

As an aside, I am not sure this equation is implemented properly in the DSSAT code (and hence our code). In the Ritchie equation, it looks like `DBAR = 0.88 * exp(35.4 * theta_avg)`, where `theta_avg = 0.5 * (theta_1 + theta_2)`. This would be an appropriate average water content if the two soil layers have the same thickness. If the layers have different thickness, a thickness-weighted version could be used: `theta_avg = (theta_1 * thickness_1 + theta_2 * thickness_2) / (thickness_1 + thickness_2)`. The DSSAT code looks like it intended to use this, but left the factor 0.5 in place. So we have `theta_avg = 0.5 * (theta_1 * thickness_1 + theta_2 * thickness_2) / (thickness_1 + thickness_2)`.

yufenghe22: yes. I agree. 0.5 is a special case when the two layers are with the same thickness. In that case, we should probably drop 0.5.

eloch216: I'll try doing this and see how much it changes the results. Hopefully not too much!

eloch216:
Removing the factor of `0.5` from the calculation of `theta_avg` has a fairly large effect on the soil water content and biomass. See images below, where "stored" is the result with the factor of 0.5 and "new" is the result without it. Nevertheless, I think it makes more sense to omit the 0.5.

Picture1

justinmcgrath:
If the idea is to replicate DSSAT, I'd do the fix in a different module, and also let the people at DSSAT know about this.

eloch216:
> If the idea is to replicate DSSAT, I'd do the fix in a different module

I don't think our goal is to replicate DSSAT.

We have already diverged from it in a major way by using hourly time steps instead of daily time steps. When the original source for a DSSAT module can be found, as is the case here, I think our goal is to implement the original model as best as we can. DSSAT is a helpful resource since it often contains additional code that is needed to address numerical errors that can occur when using the Euler method. But I think our ultimate goal is to create a set of modules whose logic can be explained and defended as much as possible, rather than a copy of DSSAT.

Another point is that we won't be constantly combing DSSAT for changes to their code. If they alter something in the next release of DSSAT, we probably won't know about it. I think it's better to think of DSSAT as a source of inspiration, rather than something we are trying to exactly replicate.

> and also let the people at DSSAT know about this.

That's probably a good idea. I will try to remember if we have noticed any other issues with the DSSAT code. (Besides the major issue of inconsistent units.)

yufenghe22:
> Removing the factor of `0.5` from the calculation of `theta_avg` has a fairly large effect on the soil water content and biomass.

I think this is a good thing. It means this is something worthwhile/sensitive to fix. Also the curve looks nicer at the surface for `soil_water_content_1` without those flat parts. I think we should use the corrected version. There will be a couple of tedious steps if we ignore it now and decide to fix it later, like all the parameterization things.

eloch216:
> > Removing the factor of `0.5` from the calculation of `theta_avg` has a fairly large effect on the soil water content and biomass.
> 
> I think this is a good thing. It means this is something worthwhile/sensitive to fix. Also the curve looks nicer at the surface for `soil_water_content_1` without those flat parts. I think we should use the corrected version. There will be a couple of tedious steps if we ignore it now and decide to fix it later, like all the parameterization things.

Great! I'll update from the target branch to address the merge conflict, and then merge this.

PR #304, opened 2026-05-06
Add c3 gm

eloch216:
This PR includes mesophyll conductance in our C3 photosynthesis calculations. Adding it to `c3photoC` was pretty straightforward -- mesophyll conductance is just another conductance in series with the boundary layer conductance and stomatal conductance. There is also an established temperature response for it, published by Carl in 2002. Most of the changes here are simply a consequence of needing new input parameters (`gm_at_25`, `gm_Ha`, `gm_Hd`, and `gm_S`). I have set `gm_at_25` to infinity, which produces exactly the same results as before.

A note about units: `gm` is usually reported in `mol / m^2 / s / bar`, unlike stomatal and boundary layer conductance, which are usually reported in `mol / m^2 / s`. It is also reported as a conductivity to CO2 diffusion, unlike stomatal and boundary layer conductance, which are usually reported as conductivities to H2O diffusion. This explains the small differences in the way these conductivities are treated in the code.


eloch216:
I also confirmed that the temperature response looks correct by recreating the trace in Figure 1 of Carl's paper (https://doi.org/10.1104/pp.008250) in R, and overlaying the calculated one on the original figure.

Picture2

The code for calculating the response:

```r
library(BioCro)
library(lattice)

rc <- module_response_curve(
  'BioCro:c3_parameters',
  soybean$parameters,
  data.frame(
    Tleaf = seq(10, 40, by = 0.1)
  )
)

xyplot(
  gm_norm ~ Tleaf,
  data = rc,
  type = 'l',
  xlab = 'Leaf temperature (degrees C)',
  ylab = 'Mesophyll conductance normalized to unity at 25 degrees C',
  ylim = c(0, 2.5)
)
```

yufenghe22:
When we use a non-infinite `gm`, we'd also need to modify the `Vcmax25` and `Jmax25` values right? Without going back to A-Ci fitting, would this approximation work? I mean, for the base BioCro, we probably do not have access to Carl's A-Ci datasets to repeat the curve fitting.

eloch216:
> When we use a non-infinite `gm`, we'd also need to modify the `Vcmax25` and `Jmax25` values right? Without going back to A-Ci fitting, would this approximation work? I mean, for the base BioCro, we probably do not have access to Carl's A-Ci datasets to repeat the curve fitting.

I think it's essentially impossible to reliably determine chloroplastic values of $V_{cmax}$ and $J_{max}$ from apparent values.

When using values of $V_{cmax}$ and $J_{max}$ from the literature, the most important consideration is to use the same value of $g_m$ that was assumed when the curves were fit. Our current soybean values come from https://doi.org/10.1104/pp.008250. As far as I can tell, these are "apparent" values estimated by assuming $C_c = C_i$. (The paper does report values of $g_m$, but I don't think these were used when estimating $V_{cmax}$ and $J_{max}$.) So the most appropriate thing is to use $g_m = \infty$.

When actual A-Ci curves are available, they can be fit assuming any value of $g_m$, as you pointed out. But this is rare in the literature, especially in older papers. Many papers don't even show an A-Ci curve, and only report the estimated FvCB parameter values.

yufenghe22:
> I think it's essentially impossible to reliably determine chloroplastic values of

There's some empirical relationship used in regional models. See [Eq 4 in this paper](https://onlinelibrary.wiley.com/doi/full/10.1111/pce.12213). But I agree with you, looking at Fig 3 & 4, the correlations are not great.

yufenghe22 (approved):
If this is not documented, we could note that the base BioCro's photosynthesis parameters (Vcmax25, Jmax25, TPU25, and Rd25, and ?) are apparent parameters that are based on Ci assuming infinite gm. If someone uses a non-infinite gm, all these parameters need to be changed accordingly.

src/module_library/conductance_helpers.h
justinmcgrath: I have seen "amount basis" and more commonly "molar basis". I have never seen "molecular basis", and it sounds odd to me, since molecular doesn't describe the units.

justinmcgrath: It's a fraction of the "amount of molecules" using SI terminology, and the "number of molecules" using more common terminology. But molecules are not a dimension, and this wording sounds strange to me - akin to "plants grown in elevated CO2 were larger".

src/module_library/c3photo.cpp
```diff
@@ -39,7 +40,7 @@ photosynthesis_outputs c3photoC(
     double const b1,                           // dimensionless
     double const Gs_min,                       // mol / m^2 / s
     double const Ca,                           // micromol / mol
-    double const AP,                           // Pa (TEMPORARILY UNUSED)
+    double const AP,                           // Pa
```
leavesandlemmas: Can we rename AP to atmosphere_pressure or air_pressure?

src/module_library/FvCB_assim.cpp
```diff
@@ -136,20 +136,20 @@ FvCB_outputs FvCB_assim(
     FvCB_outputs result;
 
     // Calculate rates
-    if (Ci < -eps_zero) {
-        throw std::range_error("Thrown in FvCB_assim: Ci is negative.");
-    } else if (Ci <= eps_zero) {
-        // RuBP-saturated net assimilation rate when Ci is 0
+    if (Cc < -eps_zero) {
```
justinmcgrath, line 139: Do you need to check both Ci and Cc? Maybe Ci can't be negative if Cc is not negative?

src/module_library/conductance_helpers.h
```diff
@@ -116,4 +116,18 @@ inline double sequential_conductance(
     return 1.0 / (1.0 / conductance_1 + 1.0 / conductance_2);  // same units as conductance_1
 }
 
+/**
+ *  @brief Calculates the total conductance across three sequential gas paths.
+ *
+ *  This is simply an extension of the two-path version.
+ */
+inline double sequential_conductance(
```
leavesandlemmas: Consider replacing this overload with a fold so any number of paths can be used. Or using a variadic template.

### Fold version

```cpp
#include <initiailizer_list>
#include <numeric>
double series_conductance(std::initializer_list<double> ilist)
{
    auto recip = [](double k) { return 1.0 / k; };
    double resistance = std::transform_reduce(
        ilist.begin(), ilist.end(),
        0.0,
        std::plus<>{},
        recip);
    return 1.0 / resistance;
}
```

then at the call site, you just need braces.
```cpp
double kt = series_conductance({k1, k2, k3});
```

### Variadic template

```cpp
template <typename... Args>
double series_conductance(Args... conductances)
{
    return 1.0 / ((1.0 / conductances) + ...);
}
```

This version can replace both the two and three conductances in series.

src/module_library/FvCB_assim.cpp
justinmcgrath: I suppose that in this function it doesn't matter what Ci is, so Ci would be checked somewhere else if needed.

justinmcgrath (approved):

justinmcgrath:
> When we use a non-infinite `gm`, we'd also need to modify the `Vcmax25` and `Jmax25` values right? Without going back to A-Ci fitting, would this approximation work? I mean, for the base BioCro, we probably do not have access to Carl's A-Ci datasets to repeat the curve fitting.

>> I think it's essentially impossible to reliably determine chloroplastic values of V c m a x and J m a x from apparent values.

I looked into this a bit and came to the worrying conclusion that the expected value of the apparent Vcmax you get from an ACi curve depends on the values of Ci you used, which is not at all what you want from parameter estimation. To get to "true" vcmax, I'm pretty sure you at least need to know the Ci values that were used, and at that point you're basically into refitting the curves anyway.

The upshot for BioCro is that people will need to treat all of these parameters as a set.

In terms of what we do when fitting A-Ci curves, I've always been skeptical of trying to estimate gm from them as well, but that may be less bad than an estimate the depends on the Ci values.

leavesandlemmas (commented):
Overall, my only thought is that it won't be easy to switch between this version and the old version. Or do you recover the old version by setting `gm_at_25 = INF ` ?   I don't think there is any way to do that without refactoring much of the FvCB code and its client code, but I think it might be useful to consider what changes might be necessary.

Otherwise, my main suggestion is to use a variadic template to automatically generate overloads for `sequential_conductance` (and maybe rename to `series_conductance`) or use a fold (`std::accumulate` or `std::transform_reduce`).  That way, we reduce the duplicate implementations and get a four-path or five-path sequential conductance for free.

justinmcgrath:
I've seen at least one person use "apparent Vcmax" to mean the Vcmax based on the activation state, so kc,cat * [E_activated] rather than kc,cat * [E_total]. So even when using Cc, there would be an apparent Vcmax. The argument was that possibly not all molecules are activated when taking ACi curves. Terminology is annoying, and I don't know what to do with this information. We could accept that there's a conflict or say Ci-based Vcmax vs Cc-based Vcmax, or maybe there's something less cumbersome.

djaiswal:
[Grassi and Magnani (2005)](https://onlinelibrary.wiley.com/doi/full/10.1111/j.1365-3040.2005.01333.x) report both Vcmax(Ci) and Vcmax(Cc) separately in Table 5. Also, if we introduce gm into the framework, we may additionally need to account for the [temperature response of gm](https://doi.org/10.1104/pp.008250) in future (which is good, but also complicated). We encountered similar challenges while fitting cassava A–Ci curves, where incorporating gm introduced additional complexity in parameter estimation and temperature sensitivity has not been easy.
Edit:- I just saw that @eloch216  already added temperature response !

src/module_library/FvCB_assim.cpp
eloch216: $C_i$ cannot be negative if $C_c$ is not negative, provided we are at steady-state and all conductances are positive (which we always assume).

At steady state, the flux of CO2 across each barrier (boundary layer, stomata, mesophyll) is the same, equal to $A_n$. Then the CO2 concentrations in the atmosphere, at the leaf surface, in the intercellular spaces, and in the chloroplast are related to each other as

$C_s = C_a - A_n / g_{bc}$

$C_i = C_s - A_n / g_{sc}$

$C_c = C_i - A_n / g_{mc}$

So when $A_n < 0$, we have $C_a <= C_s <= C_i <= C_c$, where equality can occur if a conductance is infinitely large. In this case there is no worry about $C_i$ or $C_c$ being negative, provided $C_a$ is not negative.

When $A_n >= 0$, we have $C_a >= C_s >= C_i >= C_c$. In other words, all the CO2 concentrations lie on the interval $[C_c, C_a]$. So if $C_c$ is not negative, then $C_i$ is guaranteed to also not be negative.

Besides this, the FvCB function only represents the response of the C3 cycle to CO2 in the chloroplast. So it does not need to know about $C_i$, or any other CO2 concentration besides $C_c$.

src/module_library/c3photo.cpp
eloch216: Sure. It is called `atmospheric_pressure` when it is a module input or output, so we can use that here.

src/module_library/conductance_helpers.h
eloch216: Scott R was always opposed to using templates when they are not absolutely necessary, due to the confusing error messages they can sometimes generate. I have definitely internalized that attitude.

I think it's doubtful we will ever need more than three conductances, at least in this context of gas flow to/through the leaf. But, since you already wrote the "fold version," I can implement it.

leavesandlemmas: There are ways to improve the error messages, especially using concepts, type_traits, and `static_assert`. But compiler / standard library support for this is better in latter versions of C++ https://en.cppreference.com/cpp/language/constraints

eloch216:
> If this is not documented, we could note that the base BioCro's photosynthesis parameters (Vcmax25, Jmax25, TPU25, and Rd25, and ?) are apparent parameters that are based on Ci assuming infinite gm. If someone uses a non-infinite gm, all these parameters need to be changed accordingly.

I will add this to `man/soybean.Rd`

> Overall, my only thought is that it won't be easy to switch between this version and the old version. Or do you recover the old version by setting `gm_at_25 = INF`?

Yes, the old behavior is now a special case where `gm_at_25` is `Inf`.

In that case, `1 / Gt = 1 / gbc + 1 / gsc + 1 / gm` reduces to `1 / Gt = 1 / gbc / 1 / gsc` (as it was before `gm` was added). And, `Ci = Cc + An / gm` reduces to `Ci = Cc`, so there is no difference between the two (and the values of `Cc` passed to the FvCB model are identical to the values of `Ci` that were passed before).

I forgot to mention this in my original PR message, but I have confirmed that no outputs change when using `gm_at_25 = Inf`. The only changes to the stored `soybean` and `soybean2` simulation results are the addition of new `Cc` columns. Likewise for the model test cases. The only exception is the "automatically generated" test cases, because these have `gm_at_25 = 1`.

src/module_library/conductance_helpers.h
eloch216: I thought we had discussed this before and decided on "molecular" precisely because it does not refer to a particular choice of unit. But I can change it to "molar."

eloch216: I can change it to "number of molecules."

I am not a fan of "amount" because it sounds so generic to me. A dozen eggs, a liter of eggs, and a kg of eggs are all amounts of eggs to me. But one of these specifications is a number of eggs, one is a volume of eggs, and one is a mass of eggs.

leavesandlemmas: mol / mol  is also called mole fraction (when its a fraction of a unity). I would be tempted to call it molar concentration but that unfortunately can also mean mol / L

src/module_library/conductance_helpers.h
justinmcgrath: I don't think it's worth implementing things that don't have a clear purpose. The is presently no need for a variable length version and also no apparent future need for one.

justinmcgrath (approved):

justinmcgrath:
GitHub is being weird and I can't make this comment in line. The units use bar, but since there the overall goal of moving to coherent units, maybe it makes sense to use Pa instead.

eloch216:
I just pushed two new commits. In the first one, I tried to implement the changes that were suggested in the comments above. In the second one, I added some more test cases that use a finite `gm`. (Previously, I had only set `gm_at_25 = Inf` in the test cases so I could confirm that `Ci` had not changed.)

src/module_library/conductance_helpers.h
leavesandlemmas: Why have this function at all then?  Just inline expression manually.

justinmcgrath (approved):

src/module_library/conductance_helpers.h
eloch216: Okay, I can add the "mole fraction" terminology to the explanation.

eloch216:
> GitHub is being weird and I can't make this comment in line. The units use bar, but since there the overall goal of moving to coherent units, maybe it makes sense to use Pa instead.

I guess I have become internally conflicted about this goal. Coherent units would be nice from a coding point of view, but only if they are used EVERYWHERE in the code. If they are only used sometimes, there are still many unit conversions that need to take place. And there is a benefit to using the units that are most often presented in the literature.

That being said, I have seen a few papers reporting gm in units of `mol / m^2 / s / Pa`. It always throws me off. I think they actually use `micromol / m^2 / s / Pa`, where the conversion is `1 micromol / m^2 / s / Pa` = `0.1 mol / m^2 / s / bar`. So I think the values are high until I notice the units are different.

Anyway, I can change it.

eloch216:
> I've seen at least one person use "apparent Vcmax" to mean the Vcmax based on the activation state, so kc,cat * [E_activated] rather than kc,cat * [E_total]. So even when using Cc, there would be an apparent Vcmax. The argument was that possibly not all molecules are activated when taking ACi curves. Terminology is annoying, and I don't know what to do with this information. We could accept that there's a conflict or say Ci-based Vcmax vs Cc-based Vcmax, or maybe there's something less cumbersome.

I agree that terminology is annoying. I think the clearest approach is to always explain what's meant by "apparent," even though it is a fairly standard term. E.g. "values of apparent Vcmax, estimated by assuming Cc = Ci" or something like that.

> We encountered similar challenges while fitting cassava A–Ci curves, where incorporating gm introduced additional complexity in parameter estimation and temperature sensitivity has not been easy.

> In terms of what we do when fitting A-Ci curves, I've always been skeptical of trying to estimate gm from them as well, but that may be less bad than an estimate the depends on the Ci values.

> If someone uses a non-infinite gm, all these parameters need to be changed accordingly.

To be clear, the goal of this PR is simply to give people the option to use a finite value of `gm`. It doesn't require anyone to do that, and using an infinite value is probably the best choice in many cases. And I don't think we need to provide any advice about how to estimate `gm`.

src/module_library/conductance_helpers.h
eloch216: I think the point is that overloading a simple function to provide support for either two or three conductances can accomplish the same goal as the "fold version" or the "variadic template" (because we only need two or three conductances) but with less effort on our part. I think the code was also easier to read in the overloaded version, because now I need to know or infer what `std::transform_reduce` does. In general, we need to use our time wisely, and sometimes creating a fully generalized version of a simple function is not really worthwhile. I would have just kept the overloaded version if you hadn't already provided code as part of your suggestion.

eloch216: Looks like there is compiler issue for Windows with R version 4.1:

```c++

module_library/conductance_helpers.cpp: In function 'double sequential_conductance(std::initializer_list<double>)':
module_library/conductance_helpers.cpp:51:36: error: 'transform_reduce' is not a member of 'std'
     double const resistance = std::transform_reduce(
                                    ^~~~~~~~~~~~~~~~
module_library/conductance_helpers.cpp:51:36: note: suggested alternative: 'transform'
     double const resistance = std::transform_reduce(
                                    ^~~~~~~~~~~~~~~~
                                    transform
make: *** [Makevars:57: module_library/conductance_helpers.o] Error 1
```

eloch216: Looks like `rtools40` (which is used for R versions 4.0.0 to 4.1.3) uses the `gcc-8-3.0` toolchain. According to the internet, most C++17 library features are supported in this version, though "some specialized features were still being refined."

So it seems that `std::transform_reduce` might be one of those features.

Issues like this are another reason to prefer simpler solutions that are more likely to work across all platforms / R versions / C++ versions.

@leavesandlemmas Can you find a solution to this? Otherwise I will just revert back to the overloaded version.

justinmcgrath: I would have written the expressions inline, but I wasn't going to ask for a change when what was there worked.

justinmcgrath: accumulate probably does the same sort of thing if that's available.

justinmcgrath: Or std::reduce.

eloch216: I am doing my best to avoid learning what `std::transform_reduce` does, so I asked Gemini to convert it for me. I will see if it did a good job.

leavesandlemmas: Standby, I can test it on my machine.

eloch216: Gemini created another lambda that combines the "get reciprocal" and "add stuff together" steps, and then uses `std::accumulate` to get the sum of all the reciprocals.

leavesandlemmas: I was going to suggest 

```cpp
#include <initiailizer_list>
#include <numeric>
double series_conductance(std::initializer_list<double> ilist)
{
    auto harmonic_sum  = [](double acc, double k) { return  acc + 1.0 / k; };
    double resistance = std::accumulate(
        ilist.begin(), ilist.end(),
        0.0,
        harmonic_sum);
    return 1.0 / resistance;
}
```

leavesandlemmas (approved):

Issue #174, opened 2025-04-04
Using an integral to compute the multi-layer canopy ?


leavesandlemmas:
Making this pull request to document a thought I have had for a while.

As I understand it, BioCro uses a multilayer canopy model to compute the canopy photosynthesis rate (or whole-plant photosynthesis rate). The idea is that the upper canopy shades the lower canopy so that while you can compute the leaf's photosynthesis rate from its local environment, the local environment differs throughout the height of the canopy. I haven't been able to find a paper cited in the documentation of BioCro that describes the logic of the multilayer model, but in the derivations I have seen, the canopy photosynthesis rate equals an integral over the cumulative leaf area.

I think it might be worth expressing the multilayer modules in BioCro as an integral explicitly, because we can then use other numerical quadrature / integration methods to compute the integral. The potential advantage is that the canopy photosynthesis rate could be accurately computed with fewer evaluations of the leaf photosynthesis rate. Evaluating the leaf photosynthesis rate ultimately requires a root-solving algorithm for leaf temperature and Ci (and maybe H2O or other variables) so it could be quite expensive.  And if ePhotosynthesis is used for the leaf biochemistry, then  evaluating leaf photosynthesis is even more expensive, even with potential improvements to ePhotosynthesis.

### Multi-layer models compute an integral. 

The idea in a nut shell is that a beam of light with intensity $Q$ diminishes proportional to the amount of leaf area per infinitesimally thin layer of the canopy that the beam passes through (Beers-Lambert Law):

$$
dQ =  - k Q dL
$$

$dL$ is the leaf area of an infinitesimally thin layer. Since this is per land area, $dL$ has the same units as leaf area index. The leaf area index is the sum of all layers, or integral $L = \int dL$.  Let $z$ denote position along the vertical axis from the ground, thus  $dL = \lambda_z  dz$ where $\lambda_z$ is the leaf area per unit volume, which may vary through the canopy.  Let $z=0$ be the ground, and $H$ indicate the canopy height.     Thus, the light intensity $Q_z$ at height $z$ is 

$$
Q_z = Q_H \exp \left( - k \int_H^z \lambda_z dz \right)
$$ 

This simplifies if we define $L_z$ to be the leaf area index of the canopy above the height $z$ so that

$$
L_z = \int_H^z \lambda_z dz \qquad L = \int_H^0 \lambda_z dz
$$ 

[...]

PR #305, opened 2026-05-06
Make the canopy photosynthesis integral explicit, compute using numerical quadrature 

leavesandlemmas:
This PR addresses a long standing goal of making the canopy integral explicit #174.  Canopy photosynthesis is computed by integrating the leaf photosynthetic rate over all heights of the canopy to adjust for the varying light absorption (or rather we integrate over leaf area density).  Standard biocro uses ten different depths (layers) to evaluate this integral. 

This PR has two main changes:

1. The canopy integrand  in `c3CanAC` and `CanAC` is explicit. A `CanopyIntegrand` data type manages the interface between the leaf photosynthesis and the summations in   `c3CanAC` and `CanAC`  
2. I wrote a simple library from numerical quadrature. `quad.h`  The library has three basic types of quadrature formulas: Open and Closed Newton-Cotes, and Gauss-Legendre quadrature. 

Gauss-Legendre is more efficient than the mid-point rule that BioCro currently uses. With these changes, the average time to run the c3 canopy photosynthesis module for an entire season drops from 240 ms to 100 ms on my computer (the number of canopy integrand evaluations drops from 10 to 4 per dynamical system evaluation) but the average photosynthetic rate changes by less than 0.1%.    

All other changes basically support these two changes. I'll give a more detailed summary below. This PR does not change the fundamental behavior of `multilayer_canopy_properties`; it only updates the code to use the refactor versions of sunML and lightModel. 

Unfortunately, the bulk of this code originated around the time the soil modules were added and prior to the recent updates, so there's some "updates to the soil modules". They should not be different than the final version of #241  but double check me on that. I may have screwed up the merge commit. 

### Quadrature `quad.h` 

The numerical quadrature library adds Open and Closed Newton-Cotes, and Gauss-Legendre quadrature.  All formulas have been implemented as composite rules. Given an integral 

$$
I = \int_a^b f(x)dx
$$

the integral is split into $n$ subintegrals:

$$
I = I_0 + I_1 + \cdots = \sum_{i =1}^{n} \int_{x_{i-1}}^{x_{i}} f(x) dx
$$

Then each integral is approximaed using the desired formula. While there are formulas for computing the weights and abscissa for degree formula, there can be expensive to evaluate and there's little benefit to higher-order rules over using a composite rule. So using precomputed values is ok for our purposes.

This PR uses 2nd order gauss-legendre quadrature, while the previous ten-layer version effectively uses a composite midpoint rule.  2nd order Gauss-Legrend quadrature is exact for polynomials up to degree 3 with only two evaluations of the integrand. Thus it is much more efficient than the midpoint rule so the approximation converges with fewer evaluations.  

### Changes to `c3_canopy` and `c4_canopy`

Basically these functions are the same, except the `nlayers` argument now corresponds to the number of composite integrals in the quadrature. So for 2nd gauss-legendre quadrature,  `nlayers = 5`  yields 10 total evaluations of the canopy integrand. The equation solving inside the leaf photosynthesis functions is expensive so minimizing the number of evaluations without losing accuracy will speed up simulations.

I used the following rscript to benchmark 
```r
# benchmark_c3_canopy.r

library(BioCro)
library(microbenchmark)

params <- within(soybean$parameters, {
    lai       <- 4       
    nlayers   <- 2   # increase for higher accuracy 
    lnb0      <- 1
    lnb1      <- 1
    StomataWS <- 1
})

modules <- list(
    "BioCro:c3_canopy",
    "BioCro:solar_position_michalsky",
    "BioCro:stefan_boltzmann_longwave",
    "BioCro:canopy_gbw_thornley",
    "BioCro:height_from_lai"
)

bm <- microbenchmark(
    run_biocro(
        list(),
        params,
        soybean_weather[['2006']],
        modules,
        list()
    ),
    times = 10
)

print(bm)

sim <- run_biocro(
        list(),
        params,
        soybean_weather[['2006']],
        modules,
        list()
)

out <- c("canopy_assimilation_molar_flux", "canopy_gross_assimilation_molar_flux")
summary(sim[out])
```
Using `nlayers = 5` gives much better accuracy than the previous version for the about same computational cost, while using `nalyers = 2` gets the same average photosynthetic rate up 0.1% but for less than half of the computational cost. On my computer the average time to run the canopy photosynthesis module for an entire season drops from 240 ms to 100 ms. 

### `CanopyIntegrand` 

The canopy integrand computes the sunlit and shaded photosynthetic rates. It is essentially just extracting duplicated code block in the for-loop. Annoying  C4photo uses incident rather than absorbed ppfd so there's an extra template parameter to select between the two. Ideally, that could be removed once C4photo and C3photo are refactored a bit. I added the `canopy_light_distribution` module so anyone can examine the canopy light by layer, quite easily.  

### `module/core`

I added a new folder for c++ files that aren't modules but are used inside modules.  It was quite confusing for me at first to figure out where the photosynthesis calculations were done exactly, and some file folder organization would be helpful. It also signals to people that this code is not a module.  

### `CanopyLight` and `AtmosphereLightScattering` 

Both of these objects are essentially renaming and minor refactors of the `sunML`  and `Light_model` structs, respectively. THe names are terribly unhelpful to new developers. I used to open `Light_model` thinking thats where the leaf absorptance / reflectance was computed, for instance.  In C++ (and rust and python), it is quite common to use Camel case for type names and snake case for variable names. That way the variable can have the same name if desired without identifier clash. 

The only real change is that `CanopyLight` no longer builds and stores an array of light layers. Instead, it computes the light layer for a cumlative LAI directly.

### `LeafAssim`

Effectively this is a renaming of `canopy_photosynthesis_outputs`.  The quadrature  code can automatically integrate all of the outputs if `canopy_photosynthesis_outputs` is given vector space operations (vector addition and scalar multiplication). But then the name of `outputs` seemed non ideal.  

I also renamed `GrossAssim` to `carboxylation` since that's what it is biologically. But I think we should still write `leaf_respiration` rather than non photorespiratory co2 release rate.  Mostly because its shorter and its what most people call that parameter anyways.  If respiration should only refer to mitochrondrial CO2 efflux, then photorespiration should not be called respiration either, right? 

### Leaf nitrogen and wind speed profiles

I rewrote the leaf nitrogen and wind speed profiles to simply compute those values for a given cumulative lai. This basically eliminates the need for AuxBio.h entirely from the Canopy photosynthesis code. 

```cpp
inline double leaf_nitrogen_profile(double cumulative_lai, double LeafN, double kpLN)
{
    return LeafN * std::exp(-kpLN * cumulative_lai);
}

inline double wind_speed_profile(double cumulative_lai, double wind_speed)
{
    constexpr double k = 0.7;
    return wind_speed * std::exp(-k * cumulative_lai);
}
```


leavesandlemmas:
This PR builds and passes all tests. RCMD check found no errors, only compiler warnings which seem to be boost-related so not related to this PR.

leavesandlemmas:
@eloch216  When you have time, would you look at the error logs and confirm that the warnings are Boost related, and that is the only reason the build-check is failing?

eloch216:
Just as a heads up, I won't be able to look at this until next week.

I do agree that the check failures are all coming from boost-induced compiler warnings. We will need to deal with that eventually. I guess this PR started using a part of boost that we had never included before?

I do see changes to soil water modules when I look at the `diff`. So I think you will need to update this from `develop`. There are also some new merge conflicts due to the mesophyll conductance PR.

src/module_library/soil_type_selector.h
yufenghe22, line 1: Did you run clang-format to this file? I think that messes up with the alignment of the table. We manually aligned them. You may revert this file.

eloch216: The version on `develop` has a properly formatted table. I think this should be taken care of when this branch is updated from `develop`. The changes here have nothing to do with the soil modules.

leavesandlemmas: I manually copied the file. A merge commit did not reformat the table.

leavesandlemmas:
> to agree that the check failures are all coming from boost-induced compiler warnings. We will need to deal with that eventually. I guess this PR started using a part of boost that we had never included before?

unlikely, as it doesnt use boost for anything.

eloch216:
Okay, I am confused.

After updating from `develop`, there are still differences in three soil-related files. I guess this is because you manually edited them?

Checking the `diff` more carefully, there are changes to the `inc` and `framework` submodules. This may be the root cause of the `boost` related errors, especially the changes to `inc`.

image

I think these may have happened during this commit: https://github.com/biocro/biocro/commit/bfed240fc03f8bd63f7e57482ecd216c8ca2245b

Please make sure the changes in this branch are limited to just using quadrature. It seems there are unrelated changes in here that are causing check failures and confusion.

leavesandlemmas:
I figured it out. It's using the old version of boost for some reason. 


I dont see any changes related to the framework code, so I dont know how to get rid of the framework changes... I dont really understand how the submodules work. I have a fork of the framework repo on a different machine. But I dont understand why merging from develop doesn't update the associated submodule commits.

eloch216:
Glad the submodule issue has been addressed. Can you also make `soil_type_selector.h`, `soil_water_tiledrain.h`, and `soil_water_upflow.h` agree with the versions on `develop` so they do not appear in the `diff` here? This PR should not be making any changes to those files.

eloch216:
I'm starting to look at this PR. So far I have read through the information in the PR but haven't looked at the code yet. I already have a few comments/questions.

One is a general question -- would it be possible to use an adaptive quadrature method, as is done by R's `integrate` function?

> While there are formulas for computing the weights and abscissa for degree formula, there can be expensive to evaluate and there's little benefit to higher-order rules over using a composite rule. So using precomputed values is ok for our purposes.

I don't understand what this means.

> Basically these functions are the same, except the `nlayers` argument now corresponds to the number of composite integrals in the quadrature.

Is it safe for me to equate the "number of layers" with the "number of composite integrals in the quadrature?" It seems like it is, but I want to make sure.

In my understanding, our previous approach could be thought of as a composite integral where a very simple rule is used to estimate each subintegral, requiring one function evaluation per subinterval. Now we are using a more complicated rule that requires two function evaluations per subintegral, but is nevertheless much more efficient. Is that correct?

> I added a new folder for c++ files that aren't modules but are used inside modules.

That's a good idea.

> In C++ (and rust and python), it is quite common to use Camel case for type names and snake case for variable names.

Our BioCro style guidelines indicate that we should prefer snake case: https://biocro.org/BioCro-documentation/latest/bookdown/contributing-to-biocro.html#formatting-code Honestly, I probably wouldn't have noticed or cared about this, except that you called attention to it.

All style conventions are arbitrary, and there are always differences in personal opinion. The point of a style convention is to make sure the code is all formatted in a similar way, which makes it easier to read and follow. A style convention will never make every individual coder happy. I wish you had simply followed the guidelines, instead of opening yet another pointless discussion about them.

The work is already done, so it's fine to leave it as-is. But, in the future, please keep in mind that it's easier to just follow an existing style convention instead of trying to alter it to fit your preferences. Will you go through all the BioCro code so it follows your new convention? If not, we will just have two different styles floating around. To be clear, I'm not suggesting to actually update older code with a new style, since that would be a waste of time. The point is to highlight why there's little benefit to changing a style preference.

I understand that some of our code does already use CamelCase. These are generally historical artifacts from before we added this guideline. We try to make new code consistent with the guidelines.

> `LeafAssim`: Effectively this is a renaming of `canopy_photosynthesis_outputs`.

If it is just a renaming, shouldn't it be `CanopyAssim`? Or does it now refer to leaf-level assimilation instead of canopy-level assmiliation?

> I also renamed `GrossAssim` to `carboxylation` since that's what it is biologically. But I think we should still write `leaf_respiration` rather than non photorespiratory co2 release rate. Mostly because its shorter and its what most people call that parameter anyways. If respiration should only refer to mitochrondrial CO2 efflux, then photorespiration should not be called respiration either, right?

Terminology is another area where it's not possible to make everyone happy.

I am ambivalent about `GrossAssim` vs `carboxylation`, since they are synonyms in the case of leaf gas exchange. (Some people use `GrossAssim = An + RL`, which has never made sense to me, since `An + RL` still includes losses from photorespiration and can be negative at very low CO2 concentrations.)

For the other term, it may be true that many people still call it "respiration." But on the other hand, many of the most influential FvCB-related papers in recent years have stopped calling it that, in favor of "non-photorespiratory CO2 release." Personally, I don't find a shorter name to really justify a change in terminology -- we are no longer living in the 1950's where there were hard limits on the length of variable names. I would rather use "non-photorespiratory CO2 release" because I think the field is moving towards that terminology.

If I had my druthers, I would have suggested to not make these name changes here, since they are unrelated to the main purpose of the PR. But I am also losing steam regarding terminology. As long as it's clear what each term represents, the name is fine.

> I rewrote the leaf nitrogen and wind speed profiles to simply compute those values for a given cumulative lai.

Nice. I've been wanting to do this for some time. It seemed unnecessarily complicated to have them compute a whole vector of values.

A long time ago, these were also not consistent about whether they used the LAI at the start, midpoint, or end of a layer. Just another reason to avoid having them calculate the whole profile. It's easier to keep them consistent if some other function determines the appropriate LAI and sends it to each function.

leavesandlemmas:
> A long time ago, these were also not consistent about whether they used the LAI at the start, midpoint, or end of a layer. Just another reason to avoid having them calculate the whole profile. It's easier to keep them consistent if some other function determines the appropriate LAI and sends it to each function.

They still used the top of the layer, while the layers were evaluated at their midpoint when I started this branch. 

> Is it safe for me to equate the "number of layers" with the "number of composite integrals in the quadrature?" It seems like it is, but I want to make sure.

Precisely. The `nlayers` sets the number of subintervals.  Although upon reading this, I realize that it would be better to have the number of layers correspond to the number of integrand evaluations, that way ten layers would correspond to ten evaluations. Basically the abscissa  = layers.  It would have to be rounded to the nearest integer of subintervals.

> In my understanding, our previous approach could be thought of as a composite integral where a very simple rule is used to estimate each subintegral, requiring one function evaluation per subinterval. Now we are using a more complicated rule that requires two function evaluations per subintegral, but is nevertheless much more efficient. Is that correct?

Correct. Basically, gauss-legendre quadrature works by fitting a polynomial to the function, then integrating that polynomial exactly. But the evaluation points (abscissa) are not evenly spaced, so they can be chosen to minimize the approximation error. Specifically, they are the roots Legendre polynomials. $n$ evaluations integrates a polynomial of degree $2n − 1$ exactly (so two points is cubic, three points a quartic, etc.).  

For comparison , the midpoint rule evaluates at one point but it is only exact for a linear polynomial. The error is proportional to the first the second derivative of the function over the interval.  So the composite rule requires the many more layers / subintervals to converge than a gasus-legendre rule with the same number of evaluations.

> One is a general question -- would it be possible to use an adaptive quadrature method, as is done by R's integrate function?

Yes, I will look into it.  R uses some version of Quadpack, which apparently uses a 21-point gauss-kronod rule as its starting point to decide if more subdivisions are necessary. That's probably overkill for this particular problem.  Basically all adaptive rules look at how much the integral changes as you increase the number of subintegrals to decide if they have shrunk the error below tolerance. There's some sophisticated computational tricks to minimize the number of evaluations while doing that, but it is the main idea.

leavesandlemmas:
> > While there are formulas for computing the weights and abscissa for degree formula, there can be expensive to evaluate and there's little benefit to higher-order rules over using a composite rule. So using precomputed values is ok for our purposes.
> 
> I don't understand what this means.

Would you elaborate on what part you don't understand ?

eloch216:
> > > While there are formulas for computing the weights and abscissa for degree formula, there can be expensive to evaluate and there's little benefit to higher-order rules over using a composite rule. So using precomputed values is ok for our purposes.
> > 
> > 
> > I don't understand what this means.
> 
> Would you elaborate on what part you don't understand ?

I would say I don't understand most of it. But maybe it will make more sense once I have looked at the code. I am not very familiar with Gauss quadrature.

- What does it mean to compute the abscissa?
- What are the weights applied to?
- What is a "degree formula?"
- Which values are precomputed?

leavesandlemmas:
Ah ok.  First, I miss typed "degree formula". I meant there is a formula to compute the quadrature rule for any order. I should have said order instead of degree, but the rule's order does correspond to the degree of the polynomial. 

The idea of gauss-quadrature is to compute the integral

$$
I = \int_{-1}^1 f(x) dx
$$

by approximating it with a weighted sum

$$
I \approx \sum_i  w_i f(x_i )   
$$ 

The evaluation points of the integrand $x_i$ are called abscissa and the $w_i$ are called weights.  Suppose you could write $f$ as a linear combination of polynomials.

$$
f(x) = \sum_k \alpha_k  g_k( x) 
$$

then the integral could be computed exactly by integrating $g_k$. 

$$
I = \sum_k \alpha_k \int_{-1}^1 g_k(x) dx = \sum_k \alpha_k w_k
$$

The $\alpha_k$ could be computed using the projection formula from linear algebra. The integral 

$$
\langle f, g_k \rangle =  \int f(x) g_k (x) dx
$$

is an inner product for functions (like the dot product of vectors).  So 

$$
\alpha_k = \frac{\langle f, g_k \rangle}{\langle g_k, g_k \rangle } 
$$

Now, obviously that requires computing an integral so it doesn't immediately help, but the trick is to use a special set of polynomials which are orthogonal.  So that whenever $k \neq m$

$$
\langle g_m , g_k \rangle  =0 
$$ 

Then, for reasons I dont fully understand, you only need the roots of $g_n$ to determine the $\alpha_k$ for all lower order polynomials.  So the abscissa end up being the roots of a $n$-th order polynomial. Once the roots are known there are formulas for the weights. But to compute the roots directly requires using something like Newton's method to find the roots, or apparently using QR decomposition. So what I was saying was, I didn't bother trying to do that, I just looked up tables with the abscissa and weights for the quadrature rules up to order 4.   

Gauss-legendre works less well on integrands which are not exactly polynomials, but pretty much any smooth function looks like a polynomial if the integral is a small enough.

leavesandlemmas:
@eloch216  I am not sure adaptive quadrature is necessary.  Here's the results of running a benchmark test with C3 canopy.  I'll put the script at the end. This runs just the `c3_canopy` module with required associated modules over the course of a season. So no ODE integration, just computing the integral over various weather conditions.  The table shows the time average over that season.  The average is already the same to 3 digits. 

```
----- Benchmark Result -----

Unit: milliseconds
                              expr      min        lq     mean   median       uq      max neval
 run_canopy(lai = 5, nlayers = 1L)  55.1956  56.35545 106.9814  57.4043 169.1005 208.6640   100
 run_canopy(lai = 5, nlayers = 2L) 104.8913 106.42285 214.2069 108.6499 331.4110 374.7975   100
 run_canopy(lai = 5, nlayers = 3L) 154.8526 156.51105 269.6776 158.2833 487.1410 598.7600   100
 run_canopy(lai = 5, nlayers = 4L) 203.9018 207.18530 428.3095 212.3735 661.6168 751.0599   100
 run_canopy(lai = 5, nlayers = 5L) 254.0590 256.70280 479.9068 259.0661 824.1053 913.3605   100

----- Average Over Season -----

   func evals        assim  gross_assim transpiration
            2      8.86218      17.5477       1.47147
            4      9.00174      17.7216       1.48194
            6      9.01221      17.7347       1.48268
            8      9.01453      17.7375       1.48288
           10      9.01529      17.7385       1.48298
```

I will try testing shorter time periods later. Here's the script

```r

writeLines("Running Benchmark...")
library(BioCro)
library(microbenchmark)


# benchmark test
run_canopy <- function(lai, nlayers, doy_range=NULL, photo = "c3_canopy") {
    params <- within(soybean$parameters, {
        lai       <- lai
        nlayers   <- nlayers
        StomataWS <- 1
        # not sure these are correct, but they arent use with `lnfun = 0`
        lnb0      <-  LeafN_0 
        lnb1      <- LeafN 
        lnfun <- 0
    })

    modules <- list(
        paste0("BioCro:", photo),
        "BioCro:solar_position_michalsky",
        "BioCro:stefan_boltzmann_longwave",
        "BioCro:canopy_gbw_thornley",
        "BioCro:height_from_lai"
    )

    drivers <- soybean_weather[["2006"]]
    if (!is.null(doy_range)){
        doy_offset <- soybean_weather[["2006"]][1, "doy"]
        if (length(doy_range) == 1) {
            start <- 1 + 24 * (doy_range - doy_offset)
            end <- start + 24
            idx <- seq(start, end) 
            drivers <- drivers[idx,]
        } else {
            start <- 1 + 24 * (doy_range[1] - doy_offset)
            end <- 1 + 24 * (doy_range[2] - doy_offset)
            idx <- seq(start, end) 
            drivers <- drivers[idx,]
        }
    } 
    day <- soybean_weather[["2006"]]

    run_biocro(
        list(),
        params,
        drivers,
        modules,
        list()
    )
}

# run benchmark for increasing layers.
options(width = 200)
nlayers <- 1:5
calls <- lapply(nlayers, function(n) bquote(run_canopy(lai = 5, nlayers = .(n))))
bm <- do.call(microbenchmark, c(calls, list(times = 100, unit = "ms")))
writeLines("\n----- Benchmark Result -----\n")
print(bm)


out_vars <- c("canopy_assimilation_molar_flux", "canopy_gross_assimilation_molar_flux", "canopy_transpiration_rate")

avg <- sapply(nlayers, function(layers)colMeans(run_canopy(5, layers)[out_vars]))
avg_df <- t(avg) |> as.data.frame()
# second order rule evaluates the function twice per interval
func_evals <- 2 * nlayers 
avg_df <- cbind( func_evals , avg_df)

# renaming to fit in console 
colnames(avg_df) <- c("func evals", "assim", "gross_assim", "transpiration")
writeLines("\n----- Average Over Season -----\n")
print(format(avg_df, digits = 6,  width = 12), row.names = FALSE)

```

eloch216:
I still haven't looked over everything. But in general I think this is a nice improvement. My main concerns are making sure that we don't introduce any code duplication, making sure we don't lose anything important, and making sure the new code is documented. These are always tricky points that require some time to think about. I think I will have some comments about them later.

> Ah ok. First, I miss typed "degree formula". I meant there is a formula to compute the quadrature rule for any order. I should have said order instead of degree, but the rule's order does correspond to the degree of the polynomial.
> 
> The idea of gauss-quadrature is to compute the integral

Thanks, this was helpful. I think I understand it (enough) now.

> @eloch216 I am not sure adaptive quadrature is necessary. Here's the results of running a benchmark test with C3 canopy.

I'm not sure this test directly addresses what I had in mind. To be fair, I didn't really explain what I meant.

I was thinking that earlier in the season, when total canopy LAI is small, it may be possible to use fewer layers to save time. For example, say we know that 10 layers works well when the LAI is 6. So here each layer is 0.6 thick, in LAI units. From this perspective, we would really only need 1 layer when LAI is smaller than 0.6. An adaptive solver could determine how many layers are needed.

This can be seen in the testing code by switching to a small LAI, say 1 instead of 5. In that case, I get:

```r
----- Benchmark Result -----

Unit: milliseconds
                                expr      min        lq      mean   median       uq      max neval
 run_canopy(lai = 0.5, nlayers = 1L)  89.8486  90.71525  92.11012  91.1858  91.9064 110.8747   100
 run_canopy(lai = 0.5, nlayers = 2L) 171.4527 172.27520 173.95417 172.8118 173.9552 187.2147   100
 run_canopy(lai = 0.5, nlayers = 3L) 252.2440 253.69790 256.08079 254.3218 255.9927 288.5473   100
 run_canopy(lai = 0.5, nlayers = 4L) 333.6776 334.94160 336.73244 335.6098 336.9334 368.0041   100
 run_canopy(lai = 0.5, nlayers = 5L) 414.3618 416.36070 421.69696 417.2828 419.4721 478.3076   100

----- Average Over Season -----

   func evals        assim  gross_assim transpiration
            2      5.29638      7.89164      0.638736
            4      5.29686      7.89231      0.638770
            6      5.29691      7.89236      0.638776
            8      5.29691      7.89236      0.638777
           10      5.29692      7.89237      0.638776
```

So here we could be using `n = 1` for agreement at 3 digits when LAI is 1. Your results show that we would need `n = 3` to achieve the same degree of convergence later in the season when LAI is 5. At LAI = 1, this is a time difference of `92.11012 ms` for 1 layer vs `256.08079 ms` for 3 layers. So using fewer layers at small LAI could cut the time in half without changing the results. It's possible this could translate into significant savings during the course of a season, especially if tighter tolerances are used. (E.g. if we wanted agreement at 4 digits, 1 layer would still suffice at LAI = 1 but we would need at least 5 layers at LAI = 5, leading to larger relative time savings at low LAI.)

But, if it's a hassle to write an adaptive solver, it may not be necessary. I am just wondering about it. This could totally be a task for the future, and doesn't need to happen right away.

I also have a more general question about adaptive Gaussian quadrature. It seems that an algorithm could determine how many subintegrals to compute, assuming a fixed order of quadrature being used. Or, it could determine the order of quadrature to use, assuming a fixed number of subintegrals. I guess in our situation, where we are not able to arbitrarily use any order of quadrature because we are using precomputed weights/abscissa, it makes more sense to vary the number of subintegrals. But does the other approach also exist?

leavesandlemmas:
> I was thinking that earlier in the season, when total canopy LAI is small, it may be possible to use fewer layers to save time. For example, say we know that 10 layers works well when the LAI is 6. So here each layer is 0.6 thick, in LAI units. From this perspective, we would really only need 1 layer when LAI is smaller than 0.6. An adaptive solver could determine how many layers are needed.

It did occur to me that the number of layers does increase with LAI. The simplest "adaptive" quadrature would be to compute the number of subintervals from a desired minimum subinterval width, say `delta_lai = 2` so if `lai < 2`  then there's one interval, 2 intervals for lai 2-4 and so on.  I wasn't sure how well that'd work with the ODE integrators since it could create a jump in the computed value when the lai is near threshold. It means that the approximated integral is not a smooth function of time.

  > It seems that an algorithm could determine how many subintegrals to compute, assuming a fixed order of quadrature being used. Or, it could determine the order of quadrature to use, assuming a fixed number of subintegrals. ... But does the other approach also exist?

Sure, there is a gaussian quadrature rule up to any  $n$ for a basis set of  orthogonal polynomials (of which there are multiple possibilties).  You could crank up the order to test for convergence. 

> I guess in our situation, where we are not able to arbitrarily use any order of quadrature because we are using precomputed weights/abscissa, 

This isn't the only reason. The problem is that polynomials are bad global basis functions for anything other than polynomials. The approximation starts to stuff from the [Runge's phenomenon](https://en.wikipedia.org/wiki/Runge%27s_phenomenon) where the polynomials oscillate around the true function. So the error in gaussian quadrature might not decrease until you get to very high order. 

But because pretty much any function looks a low degree polynomial over a small enough interval, shrinking the interval tends to work better than increasing the order. But obviously there's a balancing act and what's best is problem-specific. 

There's a similar tradeoff with ODE integrators too.

leavesandlemmas:
Failure of R CMD CHECK is latex compilation realated:

# Copilot's Analysis  

The job failed due to a **LaTeX compilation error** while rendering the PDF vignettes. The specific issue is:

```
LaTeX failed to compile an_introduction_to_biocro.tex
```

Additionally, there's a secondary error about TeX Live being outdated:
```
✖ seems to be older than the local installation
✖ (revision 78988 of texlive-scripts);
✖ please use a different mirror and/or wait a day or two.
```

## Root Cause

The `pdf_vignette_index.Rmd` vignette is failing to render during the pkgdown documentation build because:

1. **TeX Live mirror issue**: The TeX Live installation is out of sync with the package manager mirror
2. **LaTeX compilation failure**: One or more of the generated `.tex` files (`an_introduction_to_biocro.tex`, `a_practical_guide_to_biocro.tex`, or `quantitative_model_comparison.tex`) cannot be compiled

## Recommendations

1. **Update TeX Live**: Add a step to update TeX Live before building documentation. In the `document.yml` workflow, add:
   ```yaml
   - name: Update TeX Live
     run: |
       tlmgr update --self
       tlmgr update --all
   ```

2. **Check for LaTeX errors**: Look at the `an_introduction_to_biocro.log` file (referenced in the error) to identify specific LaTeX compilation issues in your vignettes.

3. **Consider excluding problematic vignettes temporarily**: If the TeX Live issue persists, you can temporarily skip PDF vignette rendering by checking the `_pkgdown.yml` configuration to see if PDF vignettes can be excluded from the build until the environment is stable.

4. **Wait for TeX Live mirror sync**: The error message suggests waiting a day or two for the package manager mirror to sync with the latest TeX Live revision.

leavesandlemmas:
@eloch216  Btw, I didn't try to integrate (no pun intended) the quadrature stuff into the multilayer modules since I think it would require an extensive redesign of how they work. But I am not opposed to trying. Maybe there's a way to update them without too much trouble

eloch216:
I've now spent some time looking this over and have a few general thoughts to share.

**What I like the most**

- The code that calculates the integral is now easier to directly relate to our description of canopy photosynthesis.

  In our "Canopy Photosynthesis Models" vignette (https://biocro.org/BioCro-documentation/latest/pkgdown/articles/web_only/canopy_photosynthesis.html), we write $A_{n,canopy} = \int_{\ell = 0}^{L} A_{n,leaf}(\ell) d \ell$. Compare this integral to the code in `c3CanAC.cpp` that implements it:
  
  https://github.com/biocro/biocro/blob/014e4b86e09cf8ce9ce86ab963f628b587a66581/src/module_library/c3CanAC.cpp#L168-L170
  
  I think the resemblance could still be made more clear. But I think this is clearer than what was there previously.
   
- The code that defines the integrand is now easier to directly relate to our description of canopy photosynthesis.

  In the vignette, we describe the sunlit/shaded model as a refinement to the basic equation shown above, where the integrand is redefined as $A_{n,leaf}(\ell) = A_{n,leaf}^{sunlit}(\ell) \cdot f_{sunlit}(\ell) + A_{n,leaf}^{shaded}(\ell) \cdot f_{shaded}(\ell)$. Compare this integrand to the code in `core/photosynthesis.h` that implements it:
  
  https://github.com/biocro/biocro/blob/014e4b86e09cf8ce9ce86ab963f628b587a66581/src/module_library/core/photosynthesis.h#L148-L157
  
  Again, I think the resemblance could still be made more clear. For example, by separately storing the sunlit and shaded `leaf_assim` objects, the final line could be something like `return sunlit_assim * lp.sunlit.fraction + shaded_assim * lp.shaded.fraction;`. Then it would look identical to the equation in the vignette. I doubt this would affect the performance in a meaningful way.
   
- The integrals are now more efficient. For example, the `willow` model results with `nlayers = 3` (with the second-order Gauss quadrature, this means 6 photosynthesis evaluations for the sunlit leaves and 6 for shaded, for a total of 12 per time point) is identical to the old results that were calculated with `nlayers = 10` (with the midpoint method, this means 10 photosynthesis evaluations for the sunlit leaves and 10 for shaded, for a total of 20 per time point).

- I like the idea to put non-module files in a separate directory, although this is tangential to the purpose of implementing quadrature.

**The biggest issue to address**

Historically, the `c3_canopy` module has always been able to exactly reproduce the outputs from the `ten_layer_c3_canopy` module, provided `nlayers` was set to 10. Now, there is no value of `nlayers` where the `c3_canopy` module exactly agrees with the `ten_layer_c3_canopy` module. I am not completely sure how I think we should address this divergence.

There are a few reasons to have these two separate modules:

1. The `ten_layer_c3_canopy` module makes the layer-specific outputs (such as incident light, assimilation rate, temperature, etc) directly available as model outputs. This has been very useful for identifying potential issues in our photosynthesis calculations, and it can be used to easily visualize changes throughout the canopy.

2. The `c3_canopy` module internally calculates the intensities of direct and diffuse radiation. However, some weather data sources include their own measured values of direct and diffuse radiation, and it would be better to use these when they are available. The `ten_layer_c3_canopy` module allows users to directly use values from weather sets by simply omitting the associated `shortwave_atmospheric_scattering` and `incident_shortwave_from_ground_par` modules.

3. The `ten_layer_c3_canopy` module works by creating twenty copies of the `c3_leaf_photosynthesis` module (10 for sunlit and 10 for shaded leaves). This makes it very amenable to parallelization, since each of these instances can be run independently in parallel. This has been very important for running e-Photosynthesis-BioCro simulations. e-Photosynthesis is much more computationally intensive than our "normal" FvCB calculations, so it truly needs parallelization.

Of course, one option is to continue maintaining the two different C3 canopy modules, but no longer require that they produce the same results. This doesn't seem like a great option though.

Alternatively, I think it is possible that we could meet the needs described above with just a single type of C3 canopy module. In other words, we may be able to consolidate to just `c3_canopy`. This would necessitate a few other changes.

For point 1, we could provide another module that can be used along with the `module_response_curve` function to calculate sunlit assimilation, shaded leaf temperature, sunlit fraction, etc, at a sequence of cumulative LAI values. In other words, the layer-specific outputs would not be included in the default simulation output, but could be calculated afterwards.

For point 2, we could simply remove the internal calculations of direct and diffuse radiation, treating `par_incident_direct`, `par_incident_diffuse`, and the analogous `nir` quantities as inputs to `c3_canopy`.

Point 3 is the trickiest. In principle, it would also be possible to run each of the photosynthesis calculations in parallel when using Gaussian quadrature instead of the midpoint method. But, I am not sure if the code is written in a way that would allow this. I don't think I understand the new code quite enough to determine this on my own.

If the new quadrature code cannot accommodate parallelization, then I think we will need to continue providing two different C3 canopy models.

**What I don't like as much**

- Although the changes here make our photosynthesis calculations more efficient, they don't add any new features or abilities, as far as I can tell. In other words, they don't help us to achieve a new scientific goal, so they should not have been a high priority. A significant amount of time has already been spent on this, and more time will be needed to finalize it. That time could have been spent using BioCro for scientific purposes.

- Some parts of the code seem to be overly complicated, especially where free functions have been replaced by the constructors and methods of structs, as was done with the former free functions `lightME` and `sunML`. This new approach seems to be aimed at computer scientists, who may appreciate using more of C++'s features. However, I think the model equations have become obscured and are more difficult for a computational biologist to understand.

  For code that is intended to implement a mathematical model of some system relevant for biology, free functions are ideal. They have three great properties: (i) All the inputs are listed in one spot at the start, (ii) all the outputs are listed in one spot at the end (this is not guaranteed, but they can always be written this way), and (iii) they are relatively self-contained in the sense that all the calculations are done in the body of the function. (Of course, they may call other free functions. But those will also be relatively easy to follow.) Together, these properties make it easy to look at a free function and tell what it does.
  
  In contrast, these properties no longer hold for the `canopy_light` struct which has replaced `sunML`. First, some of its inputs are passed to its constructor. Then other inputs are passed to its `get_light_profile` method. I understand that this has been done to make the calculations more efficient, since, in principle, some of the calculations do not need to be repeated when the function is called repeatedly for different values of cumulative LAI. But, the same goal could be achieved with free functions, possibly along with the technique of partial application.
  
  Is there something I'm missing that necessitates a `struct` instead of a free function? I could be convinced, but I'm just not seeing it at the moment.

yufenghe22:
> * For example, the `willow` model results with `nlayers = 3` (with the second-order Gauss quadrature, this means 6 photosynthesis evaluations for the sunlit leaves and 6 for shaded, for a total of 12 per time point) is identical to the old results that were calculated with `nlayers = 10`

Just came across this comment. This is very useful. I have not read the whole conversation, but would it be usable for soybean for the `run_biocro`? I know this is for the canopy modules, what other tests or changes needed to determine the optimal `nlayers` for soybean and use it in the standard parameter file?

leavesandlemmas:
> > * For example, the `willow` model results with `nlayers = 3` (with the second-order Gauss quadrature, this means 6 photosynthesis evaluations for the sunlit leaves and 6 for shaded, for a total of 12 per time point) is identical to the old results that were calculated with `nlayers = 10`
> 
> Just came across this comment. This is very useful. I have not read the whole conversation, but would it be usable for soybean for the `run_biocro`? I know this is for the canopy modules, what other tests or changes needed to determine the optimal `nlayers` for soybean and use it in the standard parameter file?

I mostly tested the new modules with the soybean parameters. Unfortunately, the optimal number of layers depends on what we are optimizing for. Less computation ->  less layers. High accuracy -> more layers. But I think `nlayers =2 ` is probably a good default.

leavesandlemmas:
@eloch216  I don't know what is required for parallelization in practice. But in principle, it shouldnt be barrier. A fixed point quadrature rule is conceptually a `transform_reduce` and this link https://docs.nvidia.com/hpc-sdk/archive/22.3/compilers/c++-parallel-algorithms/index.html suggests that's good for parallelizing. 

But reducing the number of times ephotosynthesis needs to be evaluated to compute the canopy photosynthesis integral  would seem to alleviate the need for parallelization in the first place.

leavesandlemmas:
> For point 1, we could provide another module that can be used along with the module_response_curve function to calculate sunlit assimilation, shaded leaf temperature, sunlit fraction, etc, at a sequence of cumulative LAI values. In other words, the layer-specific outputs would not be included in the default simulation output, but could be calculated afterwards.

> For point 2, we could simply remove the internal calculations of direct and diffuse radiation, treating par_incident_direct, par_incident_diffuse, and the analogous nir quantities as inputs to c3_canopy.

It might be possible to address both issues using an "observer" pattern, which optionally echos or outputs the intermediate calculations per layer/abscissa  and the other. That way there wouldn't have to be copies of the same code with and without output of the intermediates. I figured out something like that for the phloem transport model, where I didnt need or want all intermediate calculations to be saved or read out all of the time, but it *is* useful for debugging.

yufenghe22:
> A fixed point quadrature rule is conceptually a `transform_reduce` and this link https://docs.nvidia.com/hpc-sdk/archive/22.3/compilers/c++-parallel-algorithms/index.html suggests that's good for parallelizing.

This is interesting. I just learn that C++ 17 can support parallelization. Why did you refer to NVIDIA's page or just a random search result? Looks like NVDIA even provides API for using their GPU for parallelization.

leavesandlemmas:
> > A fixed point quadrature rule is conceptually a `transform_reduce` and this link https://docs.nvidia.com/hpc-sdk/archive/22.3/compilers/c++-parallel-algorithms/index.html suggests that's good for parallelizing.
> 
> This is interesting. I just learn that C++ 17 can support parallelization. Why did you refer to NVIDIA's page or just a random search result? Looks like NVDIA even provides API for using their GPU for parallelization.

I just search the internet for "transform_reduce" and parallelization. There's a number of function templates in the standard lib of C++ 17  and later that are "functional programming" like.  Like `std::transform` or `std::accumulate`.   For instance, `std::transform` does the operation:

```cpp

for (size_t i = 0; i < x.len();++i) {
y[i] = op(x[i]);
}
```

where y and x are some sort of container (like `std::vector` or `std::array`) and `op` is any callable. But it's more general than the for-loop in two ways. The first is that the for-loop requires you to know how to traverse the container whereas `std::transform` uses the `iterator` concept from the standard library. So you don't have to write a different for-loop for different containers whose traversal/iteration is different.

The other aspect is that the for-loop is potentially stateful so the compiler can't assume that the calculation is safe to parallelize. But `std::transform` has an execution policy parameter which basically tells the compiler whether or not the order in which you evaluate the `x[i]` matters. At least, that's my understanding.  

I have an acquaintance that has done parallelization of scientific computing in C++, and he said it involves a lot of functional programming concepts. I am not an expert in parallelization however, and I have never really done it. 

If we were to support parallelization, I have an ODE integration algorithm in mind that would benefit from parallelization. As I have explained elsewhere, in standard biocro, the ODE has almost no feedback, and the feedback is on a time-scale of weeks, not hours. So all of the integration algorithms waste time when integrating the daily variation trying to account for feedback that isn't there.  But with parallelization, you could evaluate several days of canopy photosynthesis in parallel, then perform a mass integration step.

eloch216:
Just had a chat with Yufeng about parallelization. Apparently the photosynthesis calculations are *not* run in parallel when using ePhotosynthesis. I had misremembered something from years ago. So this is not something we need to worry about.

I think there are a few more changes that need to be made here. But I am thinking that this PR can be merged without too many more changes. Then a series of smaller PRs can be used to eventually consolidate down to just one C3 canopy module that uses quadrature.

src/module_library/core/canopy_light_distribution.cpp
```diff
@@ -421,14 +421,23 @@ LightProfile sunML(
...
+        leaf_reflectance_nir{leaf_reflectance_nir},
+        leaf_reflectance_par{leaf_reflectance_par},
+        leaf_transmittance_nir{leaf_transmittance_nir},
+        leaf_transmittance_par{leaf_transmittance_par},
+        par_energy_content{par_energy_content},
+        par_energy_fraction{par_energy_fraction},
+        absorptance_nir{1.0 - leaf_reflectance_nir - leaf_transmittance_nir},  // dimensionless
+        absorptance_par{1.0 - leaf_reflectance_par - leaf_transmittance_par}   // dimensionless
 {
-    if (nlayers < 1) {
```
eloch216, line 428: Why was this check removed? It's now possible to run the model with `nlayers = 0` or `nlayers = -1`. With `nlayers = 0`, the module outputs are all `NaN`. With `nlayers = -1`, they are all 0. Neither of these really makes any sense. I think it would be better to not allow these inputs in the first place.

Code I used to test this:

```r
info <- module_info('BioCro:c3_canopy', verbose = FALSE)

modin <- quantity_list_from_names(info$inputs)

modin <- within(modin, {
  leaf_reflectance_nir = 0.1
  leaf_reflectance_par = 0.1
  leaf_transmittance_nir = 0.1
  leaf_transmittance_par = 0.1
})

evaluate_module('BioCro:c3_canopy', within(modin, {nlayers = 0}))

evaluate_module('BioCro:c3_canopy', within(modin, {nlayers = -1}))
```

src/module_library/core/canopy_light_distribution.cpp
```diff
@@ -421,14 +421,23 @@ LightProfile sunML(
...
+        heightf{heightf},
+        k_diffuse{k_diffuse},
+        lai{lai},
+        leaf_reflectance_nir{leaf_reflectance_nir},
+        leaf_reflectance_par{leaf_reflectance_par},
+        leaf_transmittance_nir{leaf_transmittance_nir},
+        leaf_transmittance_par{leaf_transmittance_par},
+        par_energy_content{par_energy_content},
+        par_energy_fraction{par_energy_fraction},
+        absorptance_nir{1.0 - leaf_reflectance_nir - leaf_transmittance_nir},  // dimensionless
```
eloch216: This is something that seems like a needless complication. What is the benefit of performing some of the calculations inside the `struct` initializer list? These are much easier to follow if all the calculations are in one spot. Typically I ignore the initializer list, assuming it is "boilerplate" code that has to be there for C++ purposes but isn't implementing any part of the model.

eloch216:
This is a pretty significant change, so an entry should be made in `NEWS.md`, and the package version should be incremented to `3.3.1-3` from `3.3.1-2`.

src/module_library/core/canopy_light_distribution.cpp
yufenghe22: Because `absorptance_nir` is a `const double`, and has to be defined here.
To make it more readable, maybe add a memeber function like `calculate_absorptance`.

leavesandlemmas: Yes, I think that's why I resorted to arithmetic in the member list. The absorptance_nir can't be set once its computed.

leavesandlemmas: I think we could drop the `const` from the member fields and make `canopy_light` const once it has been constructed.  So you'd write 

```cpp
const canopy_light distr = canopy_light::from_solar( /* ...*/ ); 
```

The constructor could be made private to prevent unintentional modification of the member fields after construction.

src/module_library/core/canopy_light_distribution.cpp
leavesandlemmas: I think I removed it since `canopy_light` doesn't use that information anywhere. It's layer-agnostic now. But the check does need to go somewhere.

leavesandlemmas:
Okay, here's the changes I'll do by end of monday @eloch216 (making a list for reference) 

1. update version number.
2. add check for `nlayers > 1` 
3. Simplify the construction of `canopy_light`. 
4. Entry in `news.md`

leavesandlemmas:
Well I asked Claude code to catalog the changes needed to bring the multilayer canopy modules into line with numerical quadrature, and it was like "just make `ten_layer_c3_canopy` a wrapper around `c3CanAC` and remove the incompatible multilayer canopy modules" lol, not what I wanted.

eloch216 (approved):
I think this can be merged. There are some changes I would like to make, but they can be done in other PRs.

PR #307, opened 2026-05-12
Remove `lightME`

eloch216:
I'm going through the changes in PR #305 to make sure I understand them. I began with the new `core/atmosphere_light_scattering.h` file.

It seems that this is a replacement for `lightME.h`, and in that case, there's no need to keep the old version around. This PR fully removes `lightME.h` and updates the `shortwave_atmospheric_scattering` module to use `PhotoCore::AtmosphereLightScattering`.

As an unrelated change, I noticed that there were changes to `c3photoC.h` in PR #305, but it was just the addition of a blank line. I removed the blank line here so that file won't show up in the diff for the big PR.


eloch216:
Wrong target. I will close this and make a new PR.

PR #308, opened 2026-05-12
Remove light me

eloch216:
I'm going through the changes in PR #305 to make sure I understand them. I began with the new `core/atmosphere_light_scattering.h` file.

It seems that this is a replacement for `lightME.h`, and in that case, there's no need to keep the old version around. This PR fully removes `lightME.h` and updates the `shortwave_atmospheric_scattering` module to use `PhotoCore::AtmosphereLightScattering`.

As an unrelated change, I noticed that there were changes to `c3photoC.h` in PR #305, but it was just the addition of a blank line. I removed the blank line here so that file won't show up in the diff for the big PR.


eloch216:
Out of curiosity, what's the advantage of defining a constructor for the `PhotoCore::AtmosphereLightScattering` struct rather than making a free function that returns a `PhotoCore::AtmosphereLightScattering` struct?

leavesandlemmas (approved):

eloch216:
> Out of curiosity, what's the advantage of defining a constructor for the `PhotoCore::AtmosphereLightScattering` struct rather than making a free function that returns a `PhotoCore::AtmosphereLightScattering` struct?

After merging this PR, the diff for PR #305 shows `lightME.cpp` being moved to `core/atmosphere_light_scattering.cpp` so it's easy to see the differences between them.

It looks like one advantage is that it's not necessary to explicitly define a `struct` to return at the end of the function. So the code becomes shorter.

I guess this could be a possible disadvantage as well, because it would be possible to write a constructor that accidentally does not set one of the `struct` member values. It's harder to do that when it's necessary to build a return `struct` explicitly.

leavesandlemmas:
> It looks like one advantage is that it's not necessary to explicitly define a `struct` to return at the end of the function. So the code becomes shorter.

If the rules for aggregate initialization apply, you can just return an initilizer list e.g. `{a, b, c, ...}` .  

> Out of curiosity, what's the advantage of defining a constructor for the PhotoCore::AtmosphereLightScattering struct rather than making a free function that returns a PhotoCore::AtmosphereLightScattering struct?

I don't really remember why I wrote it that way.  In this case, there's  no particular benefit I suppose, other than one would have to have a different name for the free function (maybe by prepending `build` or `make` ?) , so the constructor doubles as the free function. 

I was probably influenced by the idea that the constructor should not constructor a value that isn't a valid one. So that there's a guarantee that the member fields are always valid. This is more an thing with SunML.

> I guess this could be a possible disadvantage as well, because it would be possible to write a constructor that accidentally does not set one of the struct member values. It's harder to do that when it's necessary to build a return struct explicitly.

The way i see it is that this is also possible when having a free function set the member fields of an object that has already be constructed with default initialization, and sadly the compiler does not enforce the rule that can't default initialize. Although they do sometimes warn about uninitialized member fields.

justinmcgrath:
A lot of thought effort was devoted to keeping the content of do_operation free of archaic parts of C++, so that it would simply be a list of equations and thus easy for people to understand. This makes it easier for new people to use the models without needing to learn esoteric aspects of C++. These changes load do_operation with what appear to mostly be unnecessary and often not useful C++ features. Where these are occasionally useful, there seems to be no reason for them to be used in do_operation. This is loading BioCro with aspects that are going to dissuade people from using it. I had also discussed prioritizing scientific tasks, restricting framework or logical changes to ones that facilitate the scientific goals. Overwhelmingly, I see changes on here that are not related to scientific goals directly or indirectly.

PR #311, opened 2026-05-13
More careful canopy assimilation tests

eloch216:
This PR improves the "C3 canopy module tests."

The purpose of these tests is to confirm that the `BioCro:c3_canopy` module produces the same results as the `BioCro:ten_layer_c3_canopy` module when `nlayers` is set to 10.

Here, I have tightened the tolerances. I have also added an additional test that checks whether the `BioCro:c3_canopy` module produces different results for fewer numbers of layers (9 and 1).

I think these more stringent tests will be helpful for PR #305

### DIFF
--- a/tests/testthat/test.canopy_modules.R
+++ b/tests/testthat/test.canopy_modules.R
@@ -1,7 +1,11 @@
+# Specify the tolerance to use and the default number of layers
+TOLERANCE      <- 1e-6
+DEFAULT_NLAYER <- 10
+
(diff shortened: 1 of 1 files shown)


eloch216:
It looks like the documentation workflow error has been successfully fixed. After this branch is merged into `develop`, I will update the canopy integral branch from `develop` so it can access the new test and be able to pass the GitHub checks.

leavesandlemmas (approved):

PR #315, opened 2026-05-13
Update actions to newer versions

eloch216:
I noticed a warning message on some of our GitHub actions:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4.1.1, actions/upload-artifact@v4.4.3. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

It turns out there are much newer versions of the `checkout` and `upload-artifact` actions, so here I am updating them. No idea if this is related to the documentation workflow errors.

### DIFF
--- a/.github/workflows/R-CMD-check-backend.yaml
+++ b/.github/workflows/R-CMD-check-backend.yaml
@@ -176,7 +176,7 @@ jobs:
 
       - name: Check out repository
         if: ${{ !inputs.dry-run }}
-        uses: actions/checkout@v4.1.1
+        uses: actions/checkout@v6.0.2
         with:
           submodules: true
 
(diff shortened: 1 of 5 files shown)


eloch216:
It's possible that these changes have fixed the documentation workflow. There were intermittent successes before, so it's hard to tell for sure, though.

eloch216:
I am just going to merge this in the hopes it will fix our other check failures.

PR #316, opened 2026-05-13
Update baseline partitions

yufenghe22:
Given recent changes in soil water modules, this PR updates the partitioning coefficients for the baseline Soybean-BioCro based on the 93B15 from years 2002 and 2005. 

- Below is a comparison of biomasses between the old (develop branch) and the new (this). We can see they differ a lot, thus we need to re-optimize the model.

image

- Below is a biomass comparison with observation after optimization. They agree pretty well. I'd say even better than before.

image


eloch216:
FYI - updating this branch from `develop` should now fix the issue with the failing documentation check.

eloch216:
Can you add a note to `man/soybean2.Rd` stating that the model was parameterized on 2026-05-13, as we typically do for the `soybean` model when it is updated?

> They agree pretty well. I'd say even better than before.

I agree, the fits looks great.

As Justin always says, it would be nice to quantify the agreement. Then we could see how the model quality changes with time -- hopefully it improves due to our updates to soil water and other parts of the model! 

Eventually we will need to consolidate the `soybean` and `soybean2` models. But that is a task for another PR.

yufenghe22:
> Can you add a note to `man/soybean2.Rd` stating that the model was parameterized on 2026-05-13, as we typically do for the `soybean` model when it is updated?

Added.

eloch216 (approved):

PR #317, opened 2026-05-13
Small changes

eloch216:
I noticed that the `leaf_assim` `struct` had a member called `whole_plant_growth_respiration` that was not used. Conceptually, this should not be in a leaf-level representation of assimilation, since it applies at the canopy (or whole-plant) level. So, I removed it.

I also reformatted some extremely long lines that had appeared due to `clang-format`.

While doing that, I noticed a function that was defined but never implemented anywhere (`AbiotEff`). This has nothing to do with the canopy integrals, but we might as well remove it here while we are altering `BioCro.h`.

### DIFF
--- a/data/willow.R
+++ b/data/willow.R
@@ -132,7 +132,7 @@ willow <- list(
         lnfun                        = 0,
         longitude                    = -88,
         min_gbw_canopy               = 0.005,
-        nlayers                      = 10,
+        nlayers                      = 2,
         O2                           = 210,
         par_energy_content           = 0.219,
         par_energy_fraction          = 0.5,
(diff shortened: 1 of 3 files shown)


eloch216:
I also determined that the willow regression tests still pass with `nlayers` reduced from 10 to 2. I believe that is a 40% reduction in photosynthesis evaluations compared to the original 10 layer midpoint rule approach, with no appreciable change in the model outputs. Nice!

leavesandlemmas:
> I also determined that the willow regression tests still pass with `nlayers` reduced from 10 to 2. I believe that is a 40% reduction in photosynthesis evaluations compared to the original 10 layer midpoint rule approach, with no appreciable change in the model outputs. Nice!

Wow, works even better than I thought.

PR #313, opened 2026-05-13
Update document.yml

eloch216:
There have been intermittent failures of the "build documentation" workflow for a while now (https://github.com/biocro/biocro/pull/296#issuecomment-4374036018):

image

Scott got some AI suggestions for fixing it: https://github.com/biocro/biocro/pull/305#issuecomment-4444240606

I am testing one of them here

PR #318, opened 2026-05-13
Try to fix documentation error

eloch216:
Another attempt at fixing the tex errors, following up on #313


eloch216:
This one seems like it may have worked

PR #321, opened 2026-05-16
fixes to #305

leavesandlemmas:
Requested changes to #305

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: BioCro
-Version: 3.3.1-2
+Version: 3.3.1-3
 Date: 2026-02-06
 Title: Modular Crop Growth Simulations
 Description: A cross-platform representation of models as sets of equations
--- a/NEWS.md
+++ b/NEWS.md
@@ -35,6 +35,16 @@ be directly added to this file to describe the related changes.
 
 ## Minor User-Facing Changes
 
+- Made canopy photosynthesis integral explicit for `c3_canopy` and `c4_canopy` modules: canopy photosynthesis is now computed using numerical quadrature, making the integration over canopy depth explicit. This change addresses a long-standing design goal and results in a significant performance imrpovement. Several other changes are related to this change.
+  - Added numerical quadrature library: A new C++ header file quad.h in src/math/quadrature/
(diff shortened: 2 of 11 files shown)


eloch216:
I made a few more changes here:
- Added new directories to doxygen files so they are included
- Removed duplicated `@brief` because they are strange in the final doxygen documentation
- Removed untrue "forward declarations" comment
- Fixed very long lines in `NEWS.md`

This PR addresses the issues I brought up in the other PR. I think the canopy light calculations are even more over-engineered than they were before. But at least there are no calculations in initializer lists, as far as I can tell.

One other small note: it seems that the `\n` in some of the error messages are not interpreted as line breaks when the errors are printed to the R terminal. Not sure why. It's not a big deal if it's difficult to fix, but maybe they should just be removed.

Example:

```r
info <- module_info('BioCro:c3_canopy', verbose = FALSE)

modin <- quantity_list_from_names(info$inputs)

modin <- within(modin, {
  leaf_reflectance_nir = 0.1
  leaf_reflectance_par = 0.1
  leaf_transmittance_nir = 0.1
  leaf_transmittance_par = 0.1
})

evaluate_module('BioCro:c3_canopy', within(modin, {nlayers = 0}))

evaluate_module('BioCro:c3_canopy', within(modin, {nlayers = -1}))
```

produces

```r
$error_msg
[1] "Caught exception in R_evaluate_module: \nthe number of quadrature subintervals must be a positive integer (n > 0) but n = 0"

$error_msg
[1] "Caught exception in R_evaluate_module: \nthe number of quadrature subintervals must be a positive integer (n > 0) but n = -1"
```

eloch216 (approved):

leavesandlemmas:
The line breaks render correctly when written to the console. E.g., when using `run_biocro` rather than when running `evaluate_module`. I don't think there's any way to fix that from C++; that's an R thing. If the error message is passed through `writeLines` then it displays correctly. 

I am torn about removing them. On one hand, I don't like it if they aren't displayed, but otherwise the error message  quickly becomes a wall text.

eloch216:
> The line breaks render correctly when written to the console. E.g., when using `run_biocro` rather than when running `evaluate_module`. 

Oh, okay. I think it's fine to leave them then.

Issue #320, opened 2026-05-15
Conservation of mass in new soil water modules


eloch216:
I wrote a script to test whether the mass of water is conserved across a season using the new water modules. From my results, it seems that water is conserved for some years, but not always, which is puzzling to me. Maybe I am missing a source or sink for water?

Here's what I did:

1. We can find the change in volumetric soil water content across the growing season for each soil layer, and multiply by depth to get the actual amount of water (in mm) gained or lost in each layer. These can be added up to find the total change in soil water across the entire profile.
2. Precipitation and irrigation are the main sources of water in the model. The rates of precipitation and irrigation can each be integrated over the season to find the total amount of water they add (in mm). These can be added up to find the total amount of water added to the system.
3. There are several ways water can be lost in the model, most significantly through transpiration and surface evaporation. The rates for each of these "loss" processes can each be integrated over the season to find the total amount of water they emit (in mm). These can be added up to find the total amount of water lost from the system.
4. If the mass of water is conserved, we should have `change_in_soil_water = gains - losses`.

I wrote a script to test this. It looks like water is conserved in 2005 but not in 2002. The discrepancy in 2002 is very small. Am I missing something? Detailed results, and the script I used, are below.

Results for year 2002:

```
total change in water content across all soil layers: -80.8772 
cumulative water gains from all potential sources: 378.2496 
cumulative water losses from all potential sources: -459.1262 
sum total of gains and losses: -80.87658 

                           type                 name value_in_mm
1  Change in soil water content   delta_soil_water_1   -7.002791
2  Change in soil water content   delta_soil_water_2  -13.315474
3  Change in soil water content   delta_soil_water_3  -16.155097
4  Change in soil water content   delta_soil_water_4  -14.473612
5  Change in soil water content   delta_soil_water_5  -13.579022
6  Change in soil water content   delta_soil_water_6  -16.351202
7        Cumulative water gains        precipitation  378.249600
8        Cumulative water gains           irrigation    0.000000
9        Cumulative water gains         unmet_demand    0.000000
10      Cumulative water losses canopy_transpiration -279.152925
11      Cumulative water losses drainage_from_bottom  -14.261392
12      Cumulative water losses    excess_at_surface  -20.805498
13      Cumulative water losses        tile_drainage   -8.699747
14      Cumulative water losses  surface_evaporation -134.332988
15      Cumulative water losses       surface_runoff   -1.873626
```

Results for year 2005:

```
total change in water content across all soil layers: -108.6523 
[...]

PR #322, opened 2026-05-18
Add a bilinear StomataWS module

eloch216:
This PR adds a new module called `BioCro:stomata_water_stress_bilinear`. It is designed to address an issue related to the `BioCro:stomata_water_stress_linear` module.

In short, using the default `soybean2` model parameters, the `BioCro:stomata_water_stress_linear` module returns a fairly high value of `StomataWS` (0.45) when the relative extractable water (REW) is zero. Because of this, transpiration can occur at a fairly high rate when the soil is at (or close to) the wilting point. In turn, this can lead to situations where the water lost to transpiration exceeds the extractable water in the soil. When this happens, transpiration is allowed to proceed, and the amount of excess water usage is recorded in the `unmet_demand` output quantity.

This issue is most likely to occur in years with low precipitation, such as 2017. The figure below shows an example comparing the cumulative "unmet demand" against the cumulative precipitation. By the end of the season, the unmet demand is a few percent of the total precipitation. Essentially, it represents a significant but un-physical source of water in the model.

image

The `BioCro:stomata_water_stress_linear` module parameters have been calibrated for the soybean model. But, as the figure below shows, the calibration only covers REW above 0.4.

image

So, I think it's possible that a different relationship holds at lower values of REW. I think a very simple approach is to use a bilinear relationship, where `StomataWS` falls to zero (or close to zero) as REW approaches zero. That's what the new module does. The figure below compares the two responses. The new module has two additional parameters. `REW_0` is the value of REW where the crossover occurs, and `StomataWS_min` is the value that occurs when REW is zero. (The photosynthesis model doesn't work well for `StomataWS = 0` so it's best to choose a small but nonzero minimum value.)

image

With the bilinear module, there is no unmet demand in 2017. Furthermore, there is a reduction in canopy transpiration as REW falls to zero, as shown in the figure below. The bilinear module also seems to prevent REW from dropping below 0. This is good, because `REW < 0` should be impossible.

image

There is a pretty significant reduction in the predicted yield for 2017 when using the bilinear version. This is because there is a drop in canopy assimilation late in the season, when nearly all carbon is being partitioned to the seed. I think this makes sense, and the previous yield value was probably an overprediction.

image

Our regression tests use weather data from 2002. There is more precipitation during that year, and switching to the bilinear StomataWS module has no impact on the simulation results. Thus, there was no need to update the stored test data.

I have tested this for a variety of years using the LD11 soybean model that Yufeng and I are developing elsewhere. Using `REW_0 = 0.1` and `StomataWS_min = 0.01` seems to prevent unmet demand across all the scenarios I have tested. But, `REW_0 = 0.05` does not. So I think 0.1 is the lowest value we can use.


eloch216:
The calibration plot shown in my comment above was generated by the `inst/extdata/optimize_StomataWS_linear.R` script. The other plots were generated by the following script. It's easy to alter it for testing other years or other parameter values.

```r
# Load libraries
library(BioCro)
library(PhotoGEA)
library(lattice)

# Clear workspace
rm(list = ls())

# Choose options
YEAR <- 2017

START_DOY <- 152
END_DOY   <- 278

REW_0         <- 0.10
STOMATAWS_MIN <- 0.01

# Make sure we are using the correct version of BioCro
bilinear_module <- 'BioCro:stomata_water_stress_bilinear'
linear_module   <- 'BioCro:stomata_water_stress_linear'

if (!bilinear_module %in% get_all_modules('BioCro')) {
  stop('Bilinear stomata water stress module was not found')
}

# Helping function for adding photoperiod length to a weather data set by
# running the oscillator clock model
add_photoperiod_length <- function(weather_data) {
  clock_output <- with(soybean_clock, {run_biocro(
    initial_values,
    parameters,
    weather_data,
    direct_modules,
    differential_modules,
    ode_solver
  )})
  weather_data[['day_length']] <- clock_output[['day_length']]
  return(weather_data)
}

# Helping function that runs the circadian clock model for the entire year and
# then truncates the data to the appropriate time range
process_weather <- function(weather_data, sow_doy, har_doy) {
  weather_data <- add_photoperiod_length(weather_data)
  
  to_keep <- weather_data[['doy']] >= sow_doy & weather_data[['doy']] <= har_doy
  
  weather_data[to_keep, ]
}

# Helping function that adds some new columns to a simulation result
extra_calculations <- function(res) {
  # Define some constants
  cm_to_mm        <- 10  # mm / cm
  Mg_per_Ha_to_mm <- 0.1 # Mg / ha / mm
  timestep        <- 1   # hr
  
  # Change the sign of unmet demand
  res$unmet_demand <- -res$unmet_demand
  
  # Calculate cumulative quantities
  res$cumulative_precip        <- cumsum(res$precip * timestep)                                      # mm
  res$cumulative_transpiration <- cumsum(res$canopy_transpiration_rate * timestep) * Mg_per_Ha_to_mm # mm
  res$cumulative_unmet_demand  <- cumsum(res$unmet_demand * timestep) * Mg_per_Ha_to_mm              # mm
  
  # Add relative extractable water
  res$REW <- (res$soil_water_content - res$soil_wilting_point) /
    (res$soil_field_capacity - res$soil_wilting_point)
  
  # Return
  res
}

# Get the processed weather data
WEATHER <- process_weather(weather[[as.character(YEAR)]], START_DOY, END_DOY)

# Run the model with the linear stomata water stress module
linear_soybean <-
  within(soybean2, {direct_modules$stomata_water_stress = linear_module})
  
linear_res <- with(linear_soybean, {run_biocro(
  initial_values,
  parameters,
  WEATHER,
  direct_modules,
  differential_modules,
  ode_solver
)})

linear_res <- extra_calculations(linear_res)

linear_res_total_precip       <- linear_res[nrow(linear_res), 'cumulative_precip']
linear_res_total_unmet_demand <- linear_res[nrow(linear_res), 'cumulative_unmet_demand']
linear_res_ratio              <- linear_res_total_unmet_demand / linear_res_total_precip

# Make a plot comparing unmet demand to precipitation
PhotoGEA::pdf_print(
  lattice::xyplot(
    cumulative_precip + cumulative_unmet_demand ~ fractional_doy,
    data = linear_res,
    type = 'l',
    grid = TRUE,
    ylab = 'Amount of water (mm)',
    xlab = paste0('Day of year (', YEAR, ')'),
    main = paste0(
      'Total precip: ', round(linear_res_total_precip),
      ' mm\n Total unmet demand: ',
      round(linear_res_total_unmet_demand),
      ' mm\nRatio: ', round(linear_res_ratio, digits = 3)
    )
  )
)

# Illustrate the difference between the linear and bilinear models
field_capacity <- soybean$parameters$soil_field_capacity 
wilting_point  <- soybean$parameters$soil_wilting_point

stomata_param <- list(
  StomataWS_gradient  = soybean2$parameters$StomataWS_gradient,
  StomataWS_intercept = soybean2$parameters$StomataWS_intercept,
  soil_field_capacity = field_capacity,
  soil_wilting_point  = wilting_point
)

stomata_vary <- data.frame(
  soil_water_content = seq(wilting_point, field_capacity, length.out = 101)
)

linear_response <- module_response_curve(
  linear_module,
  stomata_param,
  stomata_vary
)

bilinear_response <- module_response_curve(
  bilinear_module,
  within(stomata_param, {
    REW_0 = REW_0
    StomataWS_min = STOMATAWS_MIN
  }),
  stomata_vary
)

combined_response <- rbind(
  linear_response,
  bilinear_response[, colnames(linear_response)]
)

combined_response$REW <- with(combined_response, {
  (soil_water_content - soil_wilting_point) /
    (soil_field_capacity - soil_wilting_point)
})

PhotoGEA::pdf_print(
  lattice::xyplot(
    StomataWS ~ REW,
    group = module_name,
    data = combined_response,
    type = 'l',
    xlim = c(0, 1),
    ylim = c(0, 1),
    xlab = 'REW (dimensionless)',
    ylab = 'StomataWS (dimensionless)',
    main = paste0('Dashed gray lines show\nREW_0 = ', REW_0, '\nand\nStomataWS_min = ', STOMATAWS_MIN),
    par.settings = list(
      superpose.line = list(lty = c(1, 2))
    ),
    panel = function(...) {
      panel.lines(c(STOMATAWS_MIN, STOMATAWS_MIN) ~ c(0, 1), col = 'darkgrey', lty = 5)
      panel.lines(c(0, 1) ~ c(REW_0, REW_0), col = 'darkgrey', lty = 5)
      panel.xyplot(...)
    }
  )
)

# Run the model with the bilinear stomata water stress module
bilinear_soybean <- within(soybean2, {
  direct_modules$stomata_water_stress = bilinear_module
  parameters$REW_0 = REW_0
  parameters$StomataWS_min = STOMATAWS_MIN
})
  
bilinear_res <- with(bilinear_soybean, {run_biocro(
  initial_values,
  parameters,
  WEATHER,
  direct_modules,
  differential_modules,
  ode_solver
)})

bilinear_res <- extra_calculations(bilinear_res)

bilinear_res_total_precip       <- bilinear_res[nrow(bilinear_res), 'cumulative_precip']
bilinear_res_total_unmet_demand <- bilinear_res[nrow(bilinear_res), 'cumulative_unmet_demand']
bilinear_res_ratio              <- bilinear_res_total_unmet_demand / bilinear_res_total_precip

# Make a plot comparing unmet demand to precipitation for the bilinear version
PhotoGEA::pdf_print(
  lattice::xyplot(
    cumulative_precip + cumulative_unmet_demand ~ fractional_doy,
    data = bilinear_res,
    type = 'l',
    grid = TRUE,
    ylab = 'Amount of water (mm)',
    xlab = paste0('Day of year (', YEAR, ')'),
    main = paste0(
      'Total precip: ', round(bilinear_res_total_precip),
      ' mm\n Total unmet demand: ',
      round(bilinear_res_total_unmet_demand),
      ' mm\nRatio: ', round(bilinear_res_ratio, digits = 3)
    )
  )
)

# Combine the linear and bilinear simulation results
full_res <- rbind(
  within(linear_res,   {module_type = 'linear'}),
  within(bilinear_res, {module_type = 'bilinear'})
)

# Make a plot comparing biomass in the two different versions
PhotoGEA::pdf_print(
  lattice::xyplot(
    Leaf + Stem + Root + Grain ~ fractional_doy,
    group = module_type,
    data = full_res,
    type = 'l',
    xlab = paste0('Day of year (', YEAR, ')'),
    ylab = 'Biomass (Mg / ha)'
  )
)

# Make a plot comparing midday transpiration against soil water content in the
# two versions
PhotoGEA::pdf_print(
  lattice::xyplot(
    canopy_transpiration_rate ~ REW | module_type,
    group = REW < 0,
    data = full_res[full_res$hour == 12, ],
    auto.key = list(space = 'top', title = 'Is REW < 0?', cex.title = 1),
    grid = TRUE,
    type = 'p',
    par.settings = list(
      superpose.symbol = list(pch = 16)
    ),
    xlab = 'REW (dimensionless)',
    ylab = 'Mid-day danopy transpiration rate (Mg / ha / hr)'
  )
)
```

leavesandlemmas:
Does this fix the issue you discovered in #320 ?

leavesandlemmas:
> this is good, because `REW < 0` should be impossible.

Is that true? I don't see why it should be impossible? REW is normalizing the SWC between wilting point and field capacity right?  SWC could be below the wilting point

yufenghe22:
> In short, using the default `soybean2` model parameters, the `BioCro:stomata_water_stress_linear` module returns a fairly high value of `StomataWS` (0.45) when the relative extractable water (REW) is zero. Because of this, transpiration can occur at a fairly high rate when the soil is at (or close to) the wilting point.

Good observation. Instead of creating a feedback penalty on canopy ET which likely requires complex procedures, modifying the `StomataWS` under dry condition sounds like a good, easy fix. It's nice this can also make `unmet_demand` disappear.

yufenghe22:
> Does this fix the issue you discovered in #320 ?

No. #320 is looking into more detailed (small) differences in water balance. The `unmet_demand` is already included in the calculations in #320.

leavesandlemmas:
what  is unmet demand?

eloch216:
> > this is good, because `REW < 0` should be impossible.
> 
> Is that true? I don't see why it should be impossible? REW is normalizing the SWC between wilting point and field capacity right? SWC could be below the wilting point

Thanks for pointing this out. I think you are correct about this, but there is some nuance.

At or below the wilting point, transpiration cannot occur. So transpiration alone cannot reduce REW below zero. Other processes, such as surface evaporation driven by solar energy, can reduce REW below zero. I would guess that in locations and seasons where people grow crops, it would be quite rare for this to occur, though.

yufenghe22:
> what is unmet demand?

[Soil water uptake](https://github.com/biocro/biocro/blob/develop/src/module_library/soil_water_uptake.h) is calculated using canopy ET. When canopy ET is more than the total available soil water, this would create a water deficit, or `unmet_demand`. In principle, there should be a feedback from soil water availability to the ET, but this seems difficult to implement. One main difficulty is that, imo, canopy ET is integrated by leaf-level ET. A feedback to the canopy ET would require modification of `gs` at leaf-level. Unless, the soil water uptake is determined at the leaf-level from the beginning.

yufenghe22:
> Our regression tests use weather data from 2002. There is more precipitation during that year, and switching to the bilinear StomataWS module has no impact on the simulation results. Thus, there was no need to update the stored test data.

Could you also take a quick look at 2005? Because 2002 and 2005 are used to train the model. Just wonder if this would need a re-parameterization.

eloch216:
> > Our regression tests use weather data from 2002. There is more precipitation during that year, and switching to the bilinear StomataWS module has no impact on the simulation results. Thus, there was no need to update the stored test data.
> 
> Could you also take a quick look at 2005? Because 2002 and 2005 are used to train the model. Just wonder if this would need a re-parameterization.

Good point. The plots below show simulation results using `soybean_weather[['2002']]` and `soybean_weather[['2005']]`. It looks like 2005 does a change bit.

I can re-parameterize the model using the `inst/extdata/parameterize_soybean.R` script, unless you have been using a different script and this one is now out of date. It looks like the script does need to be updated to use `soybean2` instead of `soybean`.

image

image

yufenghe22:
> I can re-parameterize the model using the `inst/extdata/parameterize_soybean.R` script,

There are a couple of things that may be different when I run the baseline optimization.

1. Here are my to-be-optimized parameter list assuming zero respirations in grain in shell.
```
arg_names     <- c('alphaLeaf','alphaRoot','alphaStem','betaLeaf','betaRoot','betaStem',
                   'rateSeneLeaf','rateSeneStem','alphaSeneLeaf','betaSeneLeaf',
                   'alphaSeneStem','betaSeneStem','alphaShell','betaShell',
                   'grc_stem','grc_root',
                   'mrc_leaf','mrc_root','iSp')
```
2. Here are the weights:
`wts_on_errors <- data.frame("Stem" = 1, "Leaf" = 1, "Shell" = 0.5, "Seed" = 1,"TotalLitter" = 0.1,"Root" = 0.1,"LAI" = 1)`
3. Repeats optimization three times with different random seed numbers. I can send jobs to the cluster once you update the optimization script.
4. A constrain on the beginning of seed filling:
```
# Find the last DOY where the obs seed value is 0
# i.e., the DOY before seed is non-zero
    TrueValues <- ExpData[[i]]
    last_zero_doy <- max(TrueValues$DOY[TrueValues$Seed < 0.1])
    
    else if(doy_grain < last_zero_doy-14){ # doy_grain should not occur before seed is present
       penalty = 9999
```
5. include LAI in the error:
Download LAI observation data from [here](https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/main/run/optimization/Data/obs_lai_2002_2010.rds).
Read in LAI data and match the calibration years
```
obs_lai_years <- c(2002,2004:2010)
obs_lai <- readRDS('Data/obs_lai_2002_2010.rds')
#get the LAI for the calibration years
obs_lai_calib <- obs_lai[obs_lai_years %in% year]
```
In the objective function, add the error term:
```
#include errors on LAI
      true_lai  <- obs_lai[[i]]
      doy_inds  <- which(result$time %in% true_lai$Group.1)
      Pred_lai  <- result$lai[doy_inds]
      scale.lai <- max(true_lai$LAI_mean)
      err.lai   <- sum(((Pred_lai - true_lai$LAI_mean)/scale.lai)^2)/length(Pred_lai)
      cost      <- cost + wts2$LAI * err.lai
```
Here is a complete list of constrains I use, which is simpler than the one for LD11:
```
    if(is.na(doy_leaf)|is.na(doy_stem)|is.na(doy_shell)|is.na(doy_grain)){
       penalty = 9999
    }else if(abs(doy_leaf - doy_stem) > 5){
       penalty = 9999  #doy_leaf and doy_stem are separated by more than 5 days 
    }else if(doy_grain < last_zero_doy-14){ # doy_grain should not occur before seed is present
       penalty = 9999
    }else{
       penalty = 0
    }
```
6. This might be optional. I use Gray et al's precipitation data for the calibration (we already have the data file in the `inst/extdata/data/soyFACE_weather_data_2004thru2011.csv`):
 ```
 if(using_Gray_precip  & yr>=2004){ #Gray's does not have year 2002
    obs_weather_Gray_yeari = obs_weather_Gray[obs_weather_Gray$Year==yr,]
    obs_weather_Gray_yeari = obs_weather_Gray_yeari[obs_weather_Gray_yeari$DOY>=sow.date[i] &obs_weather_Gray_yeari$DOY<=harv.date[i],]
    obs_weather_Gray_yeari_hourly <- obs_weather_Gray_yeari %>%
      # Add an hourly sequence per day
      uncount(weights = 24, .id = "hour") %>%
      # Adjust the hour (0 to 23)
      mutate(hour = hour - 1,
             precip = precip.mm. / 24)

    # replace precip with Gray's
    # If Gray's data is shorter than the growing season we hard-coded, we only replace the shorter part
    weather_growing_season$precip[weather_growing_season$doy>=obs_weather_Gray_yeari$DOY[1] &weather_growing_season$doy<=tail(obs_weather_Gray_yeari$DOY,1)] = obs_weather_Gray_yeari_hourly$precip
  }
```

**Comment by eloch216** [comment]
> > I can re-parameterize the model using the `inst/extdata/parameterize_soybean.R` script,
> 
> There are a couple of things that may be different when I run the baseline optimization.

Thanks, I am working on adding these to the script. I have a few questions/comments.

>     1. Here are my to-be-optimized parameter list assuming zero respirations in grain in shell.

You are also setting `mrc_stem` equal to `mrc_leaf`, right?

>     4. A constrain on the beginning of seed filling:

This constraint is a bit awkward to implement in the `BioCroValidation` objective function because the `extra_penalty_function` is only applied to the result of a simulation. In other words, it does not have access to the observed values.

There are ways to make it work. But I think it would be simpler to just use `Rep` and `Seed` in the objective function, rather than `Shell` and `Seed`. Then there will automatically be a penalty if the reproductive mass is nonzero when it is observed to be zero. This is what we have been doing with the LD11 optimization. If it's okay with you, I would rather use this approach here, too.

UPDATE: I now see that both `Shell` and `Seed` are zero in the biomass data table at the start of the season. I guess this must not be sufficient to ensure the seed starts growing at the correct time. So I will use your constraint.

>     5. include LAI in the error:

I guess the LAI observations are from Grey et al. 2016? But they are not included in the supplemental information of that paper. Are they available online somewhere or did you just get them directly from Andrew? I also have the same question about the weather data.

**Comment by yufenghe22** [comment]
> You are also setting `mrc_stem` equal to `mrc_leaf`, right?

Yes.



> Then there will automatically be a penalty if the reproductive mass is nonzero when it is observed to be zero. This is what we have been doing with the LD11 optimization. If it's okay with you, I would rather use this approach here, too.
>UPDATE: I now see that both Shell and Seed are zero in the biomass data table at the start of the season. I guess this must not be sufficient to ensure the seed starts growing at the correct time. So I will use your constraint.

Yes. We can try either way. It might achieve similar result. Basically, I found that the onset DOY of seed development (defined as when `kGrain>0.01`) should occur no earlier than two weeks before the last DOY with zero seed observations. This worked out pretty well for me. Maybe we could also use the initial values of `shell` and `seed` from your LD11 measurements.


> I guess the LAI observations are from Grey et al. 2016? But they are not included in the supplemental information of that paper. Are they available online somewhere or did you just get them directly from Andrew? I also have the same question about the weather data.

Both LAI and weather data are included in their supplements. LAI (2004-2011) is included in a folder named `ExtendedDataFig2_Table_S2_LAI` and the weather is in a folder named `ExtendedDataFig1_Weather_Data_2004thru2011`. They are in a zip file of their supplement. LAI of 2002 is from Megan's Soybean paper. I just combined the LAI into a single file. 

I see the confusion. That data link is in a [weird place here](https://www.nature.com/articles/nplants2016132#Bib1:~:text=Plot%20mean%20data%20and%20statistical%20code%20for%20all%20analyses%20in%20the%20paper%20are%20deposited%20at%20Data%20Dryad%20(http%3A//dx.doi.org/10.5061/dryad.g0v62).).

**Comment by eloch216** [comment]
I have updated the scripts to accommodate the changes you described above. Some of the alterations need some extra explanation.

Since we are now using the LAI data and precipitation data from Grey et al. (2016), I included those as data sets in the `BioCroValidation` R package. The latest version can be installed using `remotes::install_github('biocro/BioCroValidation', ref = 'develop')`.

I also updated `BioCroValidation::objective_function` so the `extra_penalty_function` can have access to the observations that correspond to each simulation result. With this change, it is possible to add a penalty when grain growth starts too soon as compared to the observations.

A summary of the changes to `BioCroValidation` can be found here: https://github.com/biocro/BioCroValidation/pull/20

I also tried to make it easier to run the script with different numbers of processor cores and different randomization seed values. Now, these must be specified before calling the script. For example, `NCORES <- detectCores() - 1; SEED <- 1234; source('parameterize_soybean.R')` will use nearly all cores, and a seed value of 1234. The results will be saved in a directory called `outputs/soybean_optim_seed_1234`. Hopefully this is helpful. An error message will occur if these settings are not specified beforehand.

File: inst/extdata/parameterize_soybean.R [hunk 1]
```diff
@@ -255,19 +352,22 @@ bounds <- bounds_table(
...
     betaSeneLeaf  = c(bll,    0),
     betaSeneStem  = c(bll,    0),
     rateSeneLeaf  = c(0,      0.0125),
     rateSeneStem  = c(0,      0.005),
     mrc_leaf      = c(mll,    mul),
     mrc_root      = c(mll,    mul),
     grc_stem      = c(8e-4,   0.08),
-    grc_root      = c(0.0025, 0.075)
+    grc_root      = c(0.0025, 0.075),
+    iSp           = c(1,      5)
```
yufenghe22: I put much narrower bounds on iSp, [2.5,3.5] assuming the true value around 3 for 93B15? Will see how the optimized value looks like. The random seeds changed the value of iSp quite a lot in my previous optimizations, so we could select one that's close to 3.

eloch216: Sounds good. Can you push the changes you made to the script? I am not currently making any changes to this branch.

yufenghe22: The optimized iSp with the [1,5] bound came back to be reasonable (see the results in the shared box folder), so I did not modify the bounds. I added an option to run the script on biocluster. By default, it's `off` so it's good to run on local computer.

justinmcgrath:
If you haven't seen this paper, it is tangentially relevant to this thread: https://www.sciencedirect.com/science/article/pii/S0378429026002236.

The three models they examine all have some sort of issue responding to water availability, but in different ways. None of them capture spatial variability either.

This looks like everything works, but maybe a smooth minimum to make it continuous would help the solver when values are near REW.

inst/extdata/parameterize_soybean.R
```diff
@@ -164,7 +252,10 @@ independent_arg_names <- c(
...
   # Maintenance respiration for leaf and root
   'mrc_leaf',
-  'mrc_root'
+  'mrc_root',
+
+  # Specific leaf area
+  'iSp'
 )
 
 independent_args <- soybean$parameters[independent_arg_names]
```
yufenghe22: should this be soybean2?

inst/extdata/parameterize_soybean.R
yufenghe22: same here. Should be `soybean2`

yufenghe22: same here

yufenghe22: and here

inst/extdata/parameterize_soybean.R
eloch216: Yes, good catch.

I should probably rewrite this a bit to make it easier to change the reference model. E.g. something like `reference_model <- soybean2` and then `independent_args <- reference_model$parameters[independent_arg_names]`, etc.

Eventually we will replace `soybean` with `soybean2` and only have one soybean model, but for now we have to deal with two separate models.

eloch216:
> If you haven't seen this paper, it is tangentially relevant to this thread: https://www.sciencedirect.com/science/article/pii/S0378429026002236.
> 
> The three models they examine all have some sort of issue responding to water availability, but in different ways. None of them capture spatial variability either.

I hadn't seen this. This is a cool paper, thanks for sharing it.

In Table 1, BioCro is described as including "water, nitrogen, and temperature stress" and using the Richards equation for soil water dynamics. I am not sure I agree with this. We don't really have anything explicitly about nitrogen stress. And I am not sure what our previous water modules were based on. (They used v 3.1.3 so it doesn't include the new ones based on DSSAT.)

justinmcgrath:
> I hadn't seen this. This is a cool paper, thanks for sharing it.

I wonder how much the water handling changes made recently alter the water sensitivity they saw in Figure 3. In 2012, unirrigated yield reported by USDA was very low compared to 2011 and 2013. Strangely BioCro and WOFOST both dipped for a chuck of years surrounding 2012. Nothing responded in 2002 or 2003. It seems like they're responding mostly to temperature and not water status.

eloch216:
These changes have been pretty thoroughly tested now and seem to be working well. I will merge this.

PR #323, opened 2026-05-22
Update partitions

yufenghe22:
Updated partitioning coefficients with the bilinear stomataWS function. 
I used my optimization scripts hosted [in this repo](https://github.com/Matthews-Research-Group/soil-tipping-bucket/tree/main/run/optimization). Two files were needed for this optimization: [run_optimization_BioCroOnly.R](https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/main/run/optimization/run_optimization_BioCroOnly.R) and [obj_function.R](https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/main/run/optimization/obj_function.R). 

I also tried the `parameterize_soybean.R`, but I could not get a result as good as this one (see result below). Specifically, the `parameterize_soybean.R` would underestimate `leaf biomass`. There must be some small inconsistency between the two.

For the purpose of moving forward with the bilinear stomataWS function, I think we can go ahead and adopt this result and then deal with the `parameterize_soybean.R` in a different PR.
image

## Update:
I also re-ran the StomataWS parameter optimization, and they changed a little bit.


eloch216:
> I also tried the `parameterize_soybean.R`, but I could not get a result as good as this one (see result below). Specifically, the `parameterize_soybean.R` would underestimate `leaf biomass`. There must be some small inconsistency between the two.
> 
> For the purpose of moving forward with the bilinear stomataWS function, I think we can go ahead and adopt this result and then deal with the `parameterize_soybean.R` in a different PR.

Sounds good to me. I will take a look at the script you used and try to figure out what's different between the two scripts.

eloch216 (approved):

PR #324, opened 2026-05-25
Water mass balance tests

eloch216:
This PR is a follow-up to issue #320.

The main goal was to add tests to check whether our water accounting is accurate when using the new soil water modules. The new tests can be found in `tests/testthat/test.cumulative_modules.R`.

To help implement the tests, I added a new module called `BioCro:total_soil_water` that just calculates the total soil water across all the layers. Like the `BioCro:total_biomass` module, this isn't needed all the time, so I didn't add it to the model definition. I tried to channel my inner @leavesandlemmas while writing this one.

I also added some new outputs to the `BioCro:cumulative_water_dynamics` module. Now it can be used to get the cumulative water losses to drainage, surface runoff, etc. This simplifies the test because it eliminates the need to manually integrate the rates.

While working on this, I found it confusing that some of the rate quantities had "rate" in their names while others did not. There were also several units used across them: `Mg / ha / hr`, `mm / hr`, and `cm / hr`. So, here I standardized them as much as I can. For example, `drain` has become `drainage_rate` to indicate this is a rate, and its units are now `Mg / ha / hr`. When it is integrated, the "rate" is lost from its name, so `drainage` represents the cumulative drainage (in `Mg / ha`).

The only exception to the renaming and converting is `precip`. For consistency with the others, it should become `precipitation_rate` with units of `Mg / ha / hr`, and then `precipitation` should be the cumulative amount. But, this would require changing a bunch of historic BioCro weather data sets that use `precip` with units of `mm / hr`. I didn't want to deal with that so I left `precip` alone.

There are some changes to stored module and model test results, but these are just related to the changes in name and units for a few quantities. I have confirmed that no behavior is altered, and the `soybean2` simulations are otherwise unchanged.


eloch216:
Looks like there is a test failure that only occurs on Mac:

```
[87](https://github.com/biocro/biocro/actions/runs/26423398167/job/77782383048?pr=324#step:10:189)
── Failure ('test.cumulative_modules.R:185:13'): the soybean2 model accounts for all carbon when using `BioCro:partitioning_growth_calculator` as its partitioning growth calculator module ──
all(soil_evaporation_rate >= 0) is not TRUE
```

It seems strange that the soil evaporation rate is non-negative on Windows and Linux but not Mac.

tests/testthat/test.cumulative_modules.R
```diff
@@ -105,6 +141,85 @@ test_soybean_carbon_accounting <- function(partitioning_calculator) {
...
+
+        # Check that all water input rates are non-negative (except unmet
+        # demand, which should be non-positive)
+        with(soybean_res, {
+            expect_true(all(irrigation_rate >= 0))
+            expect_true(all(precip >= 0))
+            expect_true(all(unmet_demand_rate <= 0))
+        })
+
+        ## Uncomment this when debugging test failures to check where a rate
```
eloch216, line 214: @yufenghe22 Could you try running the test locally with these lines un-commented? It should make a figure showing `soil_evaporation_rate` grouped by its sign. That might help debug this issue.

src/module_library/total_soil_water.h
```diff
@@ -0,0 +1,131 @@
...
+    {
+    }
+    static string_vector get_inputs();
+    static string_vector get_outputs();
+    static std::string get_name() { return "total_soil_water"; }
+
+   private:
+    // Definition of a struct that represents a soil layer
+    struct soil_layer {
+        double const thickness;      // cm
```
leavesandlemmas, line 50: Is it necessary to have the member fields const and also the std::vector of soil_layer const?

src/module_library/total_soil_water.h
```diff
@@ -0,0 +1,131 @@
...
+        double const water_content;  // m^3 / m^3
+    };
+
+    // Pointers to input parameters
+    double const& soil_depth_1;
```
leavesandlemmas, line 55: It is possible to swap this for a single std::vector<soil_layer> if the soil_layer holds either references or pointers. I can propose code for that if you want. The main advantage is that it is easy to change the number of layers later, but I suppose that's not much of a concern.

leavesandlemmas (approved):

src/module_library/total_soil_water.h
eloch216: I prefer to make variables `const` whenever possible. This is a habit I picked up from Scott R. I find that it helps to clarify intentions and prevent mistakes (where later code might accidentally modify something that shouldn't have been modified).

Is there a reason why it would be better for these to not be `const`?

leavesandlemmas:
> Looks like there is a test failure that only occurs on Mac:
> 
> ```
> [87](https://github.com/biocro/biocro/actions/runs/26423398167/job/77782383048?pr=324#step:10:189)
> ── Failure ('test.cumulative_modules.R:185:13'): the soybean2 model accounts for all carbon when using `BioCro:partitioning_growth_calculator` as its partitioning growth calculator module ──
> all(soil_evaporation_rate >= 0) is not TRUE
> ```
> 
> It seems strange that the soil evaporation rate is non-negative on Windows and Linux but not Mac.

Is it basically zero ?  Floating point arithmetic error ?

src/module_library/total_soil_water.h
eloch216: My preference is to keep module class initializers simple, in the sense that they just use `get_input` and `get_op`. Then they can largely be ignored when reading a module because all the important calculations are done in the `do_operation` method.

I also think it's beneficial to keep the `get_inputs` and `get_outputs` methods as simple initializer lists of strings with unit comments. Then they can be parsed easily by our unit documenting script.

If I understand what you're proposing, it would probably involve specifying a number of layers somewhere, automatically generating input and output quantity names using that number, and also initializing a `std::vector` of soil layers, which would need to hold references rather than values. I think this would interfere with the simplicity described above.

Sometimes it is worth taking a more "engineered" approach to module design. But I don't foresee a need to change the number of soil layers very often. And I think these changes would make the module more difficult to follow. So I would prefer to keep it as-is, unless there is an issue with it (besides parts of it being repetitive).

I would argue that it's also easy to change the number of layers by adding or removing some quantities. It's just slightly less fast.

src/module_library/total_soil_water.h
eloch216: Along these lines, I also like to use the ternary operator to avoid patterns like

```c++
double var = default_value;

if (something) {
  var = other_value;
}
```

which can instead become

```c++
double const var = something ? other_value : default_value;
```

This is another thing picked up from Scott R.

tests/testthat/test.cumulative_modules.R
yufenghe22: Two plots were generated. It seems the left figure shows that there're points with `soil_evaporation_rate<0`, but I can't see it from the plot.
image

yufenghe22: Looks like a floating issue. There's ONE hour that is non-positive:
```
[1] soil_evaporation_rate: -2.775558e-17
[1] fraction_doy: 250.2917
```

yufenghe22 (approved):

tests/testthat/test.cumulative_modules.R
eloch216: > Is it basically zero ? Floating point arithmetic error ?

> Looks like a floating issue. There's ONE hour that is non-positive:

Great, thanks for checking. I will just change the test a bit so small negative values don't trigger failures.

eloch216:
Looks like the tests are all passing now so I'll merge this

src/module_library/total_soil_water.h
leavesandlemmas: > I prefer to make variables `const` whenever possible. This is a habit I picked up from Scott R. I find that it helps to clarify intentions and prevent mistakes (where later code might accidentally modify something that shouldn't have been modified).
> 
> Is there a reason why it would be better for these to not be `const`?

Here, I suppose it doesn't matter too much since soil_layer is basically just a pair of values, but marking the member fields const makes the soil_layer more difficult to reuse. The member fields can't be modified for any reason, even if it would make sense to. But marking soil_layer const means it can't be modified in that scope but it could be modified in some other context. 

I agree that preventing unintended modification is a good idea, but I am not sure spamming const is always the best way to achieve that. Unfortunately, it seems the most common strategy to prevent unintended modification in c++ is to make member fields private so that they can only be modified in certain ways by methods. But you and justin don't seem to like that.

src/module_library/total_soil_water.h
leavesandlemmas: > If I understand what you're proposing, it would probably involve specifying a number of layers somewhere, automatically generating input and output quantity names using that number, and also initializing a std::vector of soil layers, which would need to hold references rather than values. I think this would interfere with the simplicity described above.

One of the issues with biocro modules is that they aren't that modular because the modules typically make assumptions that extend outside of their `do_operation`.  We have not one module that assumes there's a specific number of soil layers, but a dozen or so.  The simplicity comes at a cost of making the modules more interdependent than they have to be.

> Sometimes it is worth taking a more "engineered" approach to module design. But I don't foresee a need to change the number of soil layers very often. And I think these changes would make the module more difficult to follow. So I would prefer to keep it as-is, unless there is an issue with it (besides parts of it being repetitive).
> 
> I would argue that it's also easy to change the number of layers by adding or removing some quantities. It's just slightly less fast.

Well I don't think it would make the code harder to follow, and it would prevent unintentional index errors (it's really easy to screw up an inline'd for loop if you forget to update one of the indices). I don't see as much benefit for the unit document script since it has been out of date with develop for most of the time I worked on biocro and it doesn't actually verify the units are correct anyways.  

But I suppose your judgement of what is easy to follow and easy to modify is more applicable than mine given you will be the person to make any future changes.

src/module_library/total_soil_water.h
yufenghe22: I guess a general good practice is to avoid `const` at the construct definition, but use `const` at the instance level.

src/module_library/total_soil_water.h
eloch216: > I don't see as much benefit for the unit document script since it has been out of date with develop for most of the time I worked on biocro

The online documentation is only for the latest release of BioCro, since it is aimed at users rather than developers. To see the documentation for the `develop` branch (or a feature branch), it's always necessary to generate it locally.

src/module_library/total_soil_water.h
```diff
@@ -0,0 +1,131 @@
...
+    // Lambda for accumulating soil water
+    auto accum_soil_water = [](double const acc, soil_layer const layer) {
+        return acc + layer.water_content * layer.thickness;  // cm
+    };
+
+    // Total water in all layers
+    double const total_water = std::accumulate(
+        soil_layers.begin(), soil_layers.end(),
+        0.0,
+        accum_soil_water);  // cm
```
justinmcgrath, line 124: total_water = soil_depth_1 *  soil_water_content_1 +
        soil_depth_2 * soil_water_content_2 + 
        soil_depth_3 * soil_water_content_3 +
        soil_depth_4 * soil_water_content_4 +
        soil_depth_5 * soil_water_content_5 +
        soil_depth_6 *  soil_water_content_6;

That appears to do the same thing with half the lines of code, no struct and no ancillary functions. Functions that work on variable length objects would make sense if that were possible here, but the layers are fixed.

Is the logic correct here, or am I misunderstanding it. The variable is called depth in the pointers, but then it's called thickness in the function. Is it actually the thickness?

eloch216: > Is the logic correct here, or am I misunderstanding it. The variable is called depth in the pointers, but then it's called thickness in the function. Is it actually the thickness?

Sorry for the super late reply here. The "soil depth" quantities are actually the thickness of each layer, not the depth. I'm not sure why they are mis-named.

PR #326, opened 2026-06-16
Change neg partitioning

eloch216:
This PR is intended to address an issue that can occur in the soybean model when there is severe water stress late in the season.

When the soil water content is very low, daytime canopy assimilation rates are greatly reduced, and the net daily canopy assimilation can become negative. If this happens late in the season when most carbon is being partitioned to the seeds (`kGrain ~ 1`), most of the respiratory losses are subtracted from the seed biomass. This can cause the seed mass to become negative, which is obviously unrealistic.

One way to solve this issue is to change the behavior of the `partitioning_growth_calculator` module so that when canopy assimilation is negative, respiratory losses are subtracted proportional to the mass of each tissue, rather than using the DVI-dependent partitioning factors. This makes it impossible for any tissue mass to become negative, since it is an exponential decay.

This approach may not accurately represent what happens to a crop under severe water stress. But I think it is more realistic than the previous version, which cannot possibly agree with reality.

The figure below compares the old and new versions of the partitioning growth module for an extreme case of late-season drought. With the new module, there is no longer a negative `Grain` mass.

Picture1

Code I used:

```r
library(BioCro)
library(PhotoGEA)
library(lattice)

drivers <- soybean_weather[['2002']]
drivers[drivers$doy >= 220, 'precip'] <- 0.0

res <- with(soybean2, {run_biocro(
  initial_values,
  parameters,
  drivers,
  direct_modules,
  differential_modules,
  ode_solver
)})

PhotoGEA::pdf_print(
  lattice::xyplot(
    Grain + Leaf + Rhizome + Root + Shell + Stem ~ fractional_doy,
    data = res,
    type = 'l',
    auto = TRUE,
    grid = TRUE,
    xlab = 'Day of year (2002)',
    ylab = 'Biomass (Mg / ha)'
  )
)

PhotoGEA::pdf_print(
  lattice::xyplot(
    soil_water_content + StomataWS ~ fractional_doy,
    data = res,
    type = 'l',
    auto = TRUE,
    grid = TRUE,
    xlab = 'Day of year (2002)',
    ylab = '(dimensionless)',
    ylim = c(0, 1)
  )
)

PhotoGEA::pdf_print(
  lattice::xyplot(
    canopy_assimilation_rate ~ fractional_doy,
    data = res,
    type = 'l',
    auto = TRUE,
    grid = TRUE,
    xlab = 'Day of year (2002)',
    ylab = 'Canopy net CO2 assimilation rate (Mg / ha / hr)'
  )
)

PhotoGEA::pdf_print(
  lattice::xyplot(
    precip ~ fractional_doy,
    data = res,
    type = 'l',
    auto = TRUE,
    grid = TRUE,
    xlab = 'Day of year (2002)',
    ylab = 'Precipitation rate (mm / hr)'
  )
)

daily_assim <- tapply(res$canopy_assimilation_rate, res$doy, sum)

PhotoGEA::pdf_print(
  lattice::xyplot(
    as.numeric(daily_assim) ~ as.numeric(names(daily_assim)),
    type = 'b',
    pch = 16,
    grid = TRUE,
    xlab = 'Day of year (2002)',
    ylab = 'Net daily canopy CO2 assimilation (Mg / ha)'
  )
)
```


eloch216:
For years with normal levels of precipitation, the new partitioning module produces similar results to the old one. Even so, I think we would need to re-parameterize the `soybean2` model if we want to use this new version.

The figure below shows the regression test results for each of the crop models.

Picture2

eloch216:
In this PR, I also added a test to check that no soybean tissue mass becomes negative when there is severe water stress. This test was initially failing when I first added it, but it now passes with the new partitioning module. The test can be found in `tests/testthat/test.cumulative_modules.R`.

leavesandlemmas:
Why not have high water stress drive respiration rates to zero?

eloch216:
> Why not have high water stress drive respiration rates to zero?

I hadn't considered that. Is there evidence for it? Naively, I wouldn't expect water stress to reduce respiration costs. If the leaves are still alive, they must be respiring, right?

I'm not sure if I understand how you would implement this. Would you want to multiply `RL` by `StomataWS` so it is reduced to zero when stress is high?

leavesandlemmas:
> I hadn't considered that. Is there evidence for it? Naively, I wouldn't expect water stress to reduce respiration costs. If the leaves are still alive, they must be respiring, right?

I guess it depends on how severe the water stress. If plant is dying and leaves are wilting, metabolism will slow so respiration would drop.  What happens under non-fatal droughts is harder to say with any certainty.  This study found a reduction in leaf respiration but only for some species that they measured https://www.sciencedirect.com/science/article/pii/S0098847214001993  but I haven't read this closely. Respiration is hard to measure.

> if the leaves are still alive, they must be respiring, right?

Not necessarily, there's several experiments I thought a lot about in my phd work that involved putting plants in complete darkness. In particular, [Weber et al. 2018](dx.doi.org/10.1111/nph.14987) who put saplings of four different species into complete darkness for weeks and found that many survived up to 9 nine weeks in complete darkness. They didn't measure respiration, but starch and sugar concentrations drop zero after ~ 3 weeks or so. Weber et al talk a lot about how there must be some hidden source of carbon to supply respiration, but probably the plants just stopped respiring much after 3 weeks, and metabolism mostly shut down. Their data fits this interpretation pretty well (no need for dark carbon). 

Respiration is supplied by starch / sucrose so it stops when the supply of those run out whenever there's a deficit (plus maybe some breakdown of proteins or fatty acids, but fatty acids aren't at high concentrations in leaf tissues and breaking down proteins for energy is a bit like burning your furniture for heat in winter). So a more principled way to drive respiration to zero is to have a pool of starch that runs out under deficits. Then the nighttime respiration isn't reduced. Unfortunately, adding that isn't straightforward with BioCro's current structure.  But even if it were, the FvCB model doesn't really model low assimilation rates well anyways (and almost certain not under extreme water stress).

leavesandlemmas (approved):
Code looks fine to me.

eloch216:
> > I hadn't considered that. Is there evidence for it? Naively, I wouldn't expect water stress to reduce respiration costs. If the leaves are still alive, they must be respiring, right?
> 
> I guess it depends on how severe the water stress. If plant is dying and leaves are wilting, metabolism will slow so respiration would drop. What happens under non-fatal droughts is harder to say with any certainty. This study found a reduction in leaf respiration but only for some species that they measured https://www.sciencedirect.com/science/article/pii/S0098847214001993 but I haven't read this closely. Respiration is hard to measure.

That's interesting. It's hard to say whether we can generalize from 3-year-old evergreens to annual crops.

Out of curiosity, I tried calculating the response of assimilation to water stress for a few different (low) values of Qabs. For low (but nonzero) light levels, there is a dependence of An on StomataWS, with severe water stress causing an increase in An (AKA a reduction in dark respiration). This occurs without any changes in RL -- it's just a consequence of stomatal closure.

image

Code I used:

```r
library(BioCro)
library(PhotoGEA)
library(lattice)

rc <- module_response_curve(
  'BioCro:c3_assimilation',
  within(soybean2$parameters, {
    atmospheric_pressure = 101325
    Catm = 420
    gbw = 1.2
    rh = 0.6
    temp = 28
    Tleaf = 30
  }),
  expand.grid(
    StomataWS = seq(0, 1, by = 0.01),
    Qabs = seq(0, 50, by = 5)
  )
)

PhotoGEA::pdf_print(
  lattice::xyplot(
    Assim ~ StomataWS,
    group = Qabs,
    data = rc,
    type = 'l',
    auto = TRUE,
    grid = TRUE,
    main = 'Water stress response of An for low Qabs'
  )
)
```

> So a more principled way to drive respiration to zero is to have a pool of starch that runs out under deficits. Then the nighttime respiration isn't reduced. Unfortunately, adding that isn't straightforward with BioCro's current structure. But even if it were, the FvCB model doesn't really model low assimilation rates well anyways (and almost certain not under extreme water stress).

A starch pool would be nice, but as you know, adding that would be a big undertaking. So, in the current version of BioCro, respiration is allowed to continue even past the point where starch reserves would have been depleted. This isn't quite realistic. But I think we could just keep it in mind as a limitation of the model -- that assimilation/growth/yield predictions will be less accurate in cases of severe long-term water stress. This is probably a limitation of most crop growth models. At the very least, they are trained using data from living crops, not those facing death from drought.

leavesandlemmas:
The same problem would occur in biocro if night were prolonged.

eloch216:
> The same problem would occur in biocro if night were prolonged.

Good point -- we will have to be careful modeling soybean growth on Venus or during a nuclear winter 😂

leavesandlemmas:
Well photoperiod experiments are a thing... besides the night is longer later in the season and/or at higher latitudes.  In the same way that biocro cannot really model changes in planting density without refitting, it couldn't model photoperiod experiments if the nighttime periods are too long. Given that the goal of biocro is to accurately simulate canopy photosynthesis, it probably doesn't matter then there's no real need to fix the negative masses at the end of the season either.

justinmcgrath (approved):
Looks good. Thanks.

eloch216:
I am running the parameterization script. When it's done, I can update `soybean2` and merge this.

yufenghe22:
> I am running the parameterization script. When it's done, I can update `soybean2` and merge this.

Sorry. I wanted to let you know that I'm following the conversation on respiration in case we want to change RL. Looks like there will be no further changes on that. I will submit jobs for optimization on the cluster.

eloch216:
> > I am running the parameterization script. When it's done, I can update `soybean2` and merge this.
> 
> Sorry. I wanted to let you know that I'm following the conversation on respiration in case we want to change RL. Looks like there will be no further changes on that. I will submit jobs for optimization on the cluster.

Thanks. I tried running the parameterization script in `inst/extdata` and got some odd results that didn't look very good. I found a mistake in the script that was causing `Root` to be excluded from the objective function. I just fixed it and will try running it again. Hopefully it produces similar results to your script now.

yufenghe22:
> > > I am running the parameterization script. When it's done, I can update `soybean2` and merge this.
> > 
> > 
> > Sorry. I wanted to let you know that I'm following the conversation on respiration in case we want to change RL. Looks like there will be no further changes on that. I will submit jobs for optimization on the cluster.
> 
> Thanks. I tried running the parameterization script in `inst/extdata` and got some odd results that didn't look very good. I found a mistake in the script that was causing `Root` to be excluded from the objective function. I just fixed it and will try running it again. Hopefully it produces similar results to your script now.

Great. Hope it fixes the issue. I submitted three jobs with different Random Seed numbers several hours ago. Will post the results once it's done.

eloch216:
The results from `inst/extdata/parameterize_soybean.R` were not that great, especially for 2005 where there is a big under-prediction for final seed biomass (see below).

I took a very close look at the scripts in `soil-tipping-bucket` (`run/optimization/run_optimization_BioCroOnly.R` and `run/optimization/obj_function.R`) and found a few other differences. Some of the parameter bounds were slightly different; I changed these but they probably don't have much of an impact.

More important, I noticed that weights are not applied to the LAI values. For example, compare the seed error term:

https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/bc55640a09e9606e4aa2608eee5898dea870d2b6/run/optimization/obj_function.R#L108

to the LAI error term:

https://github.com/Matthews-Research-Group/soil-tipping-bucket/blob/bc55640a09e9606e4aa2608eee5898dea870d2b6/run/optimization/obj_function.R#L122

Effectively, the LAI weights are fixed to 1. This is equivalent to standard deviation of 0.36786944 when using the "logarithm" weighting method. In `inst/extdata/parameterize_soybean.R`, I had been using stdev values from the `BioCroValidation::soyface_biomass` data set. With these values, some of the weights were larger than 1 and some were smaller. I have changed the script so they are all set to 1 and will try running it again.

image

image

yufenghe22:
Got the results back. Looks like the case with random number 2345 has the best results overall (https://uofi.box.com/s/lmn0ayx1pxmq15ffhan6ss9ji16vwmfl). So I also attached the corresponding parameter file [soybean2_updated.R](https://uofi.box.com/s/qlq6pbwml6qx2unq3dhoh815meyb3umb), which can be used to replace the `soybean2.R` if you agree that case is good to use.

> Effectively, the LAI weights are fixed to 1. This is equivalent to standard deviation of 0.36786944 when using the "logarithm" weighting method. In `inst/extdata/parameterize_soybean.R`, I had been using stdev values from the `BioCroValidation::soyface_biomass` data set.

Good catch. I'd be surprising to see if this impacted the results significantly. Year 2002 does not have standard deviation for the observation data, soI did not follow the original weights that are based on the inverse of std.

eloch216:
> Got the results back. Looks like the case with random number 2345 has the best results overall (https://uofi.box.com/s/lmn0ayx1pxmq15ffhan6ss9ji16vwmfl). So I also attached the corresponding parameter file [soybean2_updated.R](https://uofi.box.com/s/qlq6pbwml6qx2unq3dhoh815meyb3umb), which can be used to replace the `soybean2.R` if you agree that case is good to use.

These look good. I will update the model definition here and then merge this PR.

> > Effectively, the LAI weights are fixed to 1. This is equivalent to standard deviation of 0.36786944 when using the "logarithm" weighting method. In `inst/extdata/parameterize_soybean.R`, I had been using stdev values from the `BioCroValidation::soyface_biomass` data set.
> 
> Good catch. I'd be surprising to see if this impacted the results significantly. Year 2002 does not have standard deviation for the observation data, soI did not follow the original weights that are based on the inverse of std.

You are right, changing the LAI weights did not have a large impact.

But, I finally figured out the issue with the script -- it was just a mistake in the code that generates the final plots! During the optimization, the precipitation data from Gray et al. is used for the 2005 weather data. But, for generating the final plots, the default precipitation data was used. I changed this, and now the plots from `inst/extdata/parameterize_soybean.R` look good.

In the future, we can now use `inst/extdata/parameterize_soybean.R` for reparameterizations. This will make our workflow more transparent and repeatable. Any changes to the optimization should be made in that script so it stays up-to-date.

yufenghe22:
> But, I finally figured out the issue with the script -- it was just a mistake in the code that generates the final plots!

Glad you figured it out! I think the soil-tipping-bucket repo can be retired after the soil water paper is completed. There are a couple of other scripts for the simulations used in the paper, and I've started organizing them in a better way.

Issue #327, opened 2026-07-28
potential convergence issue in c3CanAC and improvement


yufenghe22:
The current solver for gs-Tleaf in [c3CanAC](https://github.com/biocro/biocro/blob/f0294f8dfd08e4966a245d69d9ce1b7f97164b2f/src/module_library/c3CanAC.cpp#L83) is a fixed-point method. If I remember it correctly, the reason for using this simple method is to be consistent with legacy BioCro's behavior. However, it's not a safe method.
It's probably a rare case, but I have encountered an issue with maxed-out iteration in my BioCro-ePhotosynthesis simulation (likely due to low soil water content), which would trigger an error like:
```
Error in as.data.frame(.Call(R_run_biocro, initial_values, parameters,  :
  Caught exception in R_run_biocro: c3Canopy solver reports failed convergence. Termination flag:
    Reached the maximum number of iterations.
```
One improvement could be done is to adopt the dekker method as used in [c3photoC](https://github.com/biocro/biocro/blob/f0294f8dfd08e4966a245d69d9ce1b7f97164b2f/src/module_library/c3photo.cpp#L161). Note that this may still not work for a un-solvable problem.

justinmcgrath:
Is this different from when we stopped using fixed point iteration earlier? It's not a great solution in general, so I don't see a reason to use it anywhere.

leavesandlemmas:
> Is this different from when we stopped using fixed point iteration earlier? 

Yes, #166 and #208 removed the fixed point iteration from the Ci calculation and pseudo Penman-Monteith equation for leaf temperature, respectively. But the Ci equation is solved conditional on a known leaf temperature. And the leaf temperature solution for the pseudo penman monteith equation is conditional on Ci (which is then used to compute Gs). Both now use Dekker's method.  @yufenghe22  is referring to the fixed point iteration used to find the Gs that solves both.  It was and still is done using fixed point iteration. PR #274 added checks to ensure that a failed convergence is detected as well as more iterations than the original.

   

> It's probably a rare case, but I have encountered an issue with maxed-out iteration in my BioCro-ePhotosynthesis simulation (likely due to low soil water content), which would trigger an error like:

What is the residual in those cases? 

> One improvement could be done is to adopt the dekker method as used in [c3photoC](https://github.com/biocro/biocro/blob/f0294f8dfd08e4966a245d69d9ce1b7f97164b2f/src/module_library/c3photo.cpp#L161). Note that this may still not work for a un-solvable problem.

[...]

PR #328, opened 2026-07-31
Dekker method for gs iteration in c3CanAC 

leavesandlemmas:
Here's a pr to make the change suggested in #327.  

## Summary of Changes

Replaces fixed_point iteration with the dekker root-finding method for the coupled stomatal-conductance / leaf-energy-balance solve at all four leaf photosynthesis call sites.  Modifies the one-dimensional root-finding library's error reporting. 

I ran the tests to ensure that the solvers still work, but many of the tests fail due to slight differences. I haven't updated the tests, but someone else should make that judgement call.  I have not tested the efficiency (e.g. how iterations is needed or the effect on speed of a simulation). 

### 1. Switch gs solver from fixed_point to dekker
+ The equation solver looks for the fixed point of the current system (the lambda function was rewritten so that the zero is the fixed point).
+ I picked `Gs_min` and `100 * gsw_estimate + 0.1` as the bracket points. A good upper bound was tricky to find but this version does not encounter any invalid_brackets when running tests.  The trick is to get an upper that's wide enough. 
+ Applied to CanAC.cpp, c3CanAC.cpp, c3_leaf_photosynthesis.cpp, and c4_leaf_photosynthesis.cpp. 

### 2. Modifications to the root-finding library (src/math/roots/onedim/roots.h)

I made some modifications to the root-finding library for diagnostic purposes. Basically I wanted to see what the brackets were when they cause the invalid_bracket error. Unfortunately that required a minor redesign of the result_t struct and how methods populate it. It's not as clean as I might like (I let claude draft it), but it works. It's also helpful to have a print out for the residual and root and iteration count whenever there's a failed termination.  

+ Added result_t::message(), a member function that reports root, iteration, residual, and the termination reason in one string.
+ Added result_t::bracket, an optional<pair<graph_t, graph_t>> populated only by bracketing methods. When present, message() includes the bracket.  
+ Added a CRTP bracket() hook to root_finding_method<Method> (default nullopt), overridden in the 8 bracketing methods (bisection, regula_falsi, ridder, illinois, pegasus, anderson_bjorck, dekker, dekker_newton) to report their own left/right (or best/contrapoint) state - non-bracketing methods needed no changes.
+ Updated error-message call sites updated (8 files): CanAC.cpp, c3CanAC.cpp, c3_leaf_photosynthesis.cpp, c4_leaf_photosynthesis.cpp, c3photo.cpp, c4photo.cpp, boundary_layer_conductance.cpp, leaf_energy_balance.cpp


eloch216:
I started running the `inst/extdata/parameterize_soybean.R` script to test this. The optimizer tries out many odd parameter values, resulting in a lot of `c3_leaf_photosynthesis solver reports failed convergence with termination flag: Reached the maximum number of iterations` errors on the `develop` branch. So far, after 23 iterations, there are no error messages yet!

eloch216:
I took a look at the crop model outputs to see how much they had changed. There are some VERY substantial changes to the soybean model outputs. This is surprising to me, since the soybean models did not report any convergence failures before. If there weren't any convergence failures before, and there aren't any now, shouldn't the simulation results be very close?

Picture1

eloch216:
It looks like there are very large differences in canopy transpiration rate with the new Dekker solver. This would be expected if there are changes to the leaf-level stomatal conductance values. These are depleting the soil water content, causing water stress to be high, and reducing assimilation.

Picture2

eloch216:
Canopy transpiration is higher with the Dekker method, but stomatal conductance is not. This is puzzling.

image

eloch216:
Leaf temperatures are not very much different:

image

leavesandlemmas:
Hmm, wonder if there's multiple solutions. Multiple fixed points would not be obvious using fixed point iteration since it would converge to stable fixed points  so maybe there's a stable fixed point and unstable one. The secant method could converge to either. A bracketing method only works if there's an odd number of fixed points inside the bracket, so that suggests 3 fixed points.

eloch216:
The residuals returned by both the Dekker and fixed-point method are all exactly zero, is this normal / believable?

image

leavesandlemmas:
> The residuals returned by both the Dekker and fixed-point method are all exactly zero, is this normal / believable?

Hmm, if the starting guess is the correct solution, then maybe so. But it's worth exploring more.

eloch216:
> > The residuals returned by both the Dekker and fixed-point method are all exactly zero, is this normal / believable?
> 
> Hmm, if the starting guess is the correct solution, then maybe so. But it's worth exploring more.

Hmm, just checked the iterations too. They are all zero, which does not seem believable:

image

eloch216:
The residuals and iterations all being zero is partially a result of the regression test only storing / checking every 24th value (which happens to be at night). I was using the regression test script to generate these comparisons.

If I just run the Dekker version of the model and plot every time point, there are nonzero residuals and iterations.

Picture3

eloch216:
Now I'm comparing the full model output at hourly time step. It's useful to compare behavior at different hours of the day. The daytime transpiration rates are similar, but nighttime transpiration with the Dekker solver is much higher, even though nighttime stomatal conductance is not much different. Maybe the transpiration model is very sensitive to small values of stomatal conductance? I think there might be a `1 / gs` somewhere in it. So even tiny changes to a low `gs` could have a big impact on transpiration. I will need to double check this.

image

image

image

eloch216:
Aha! I think I figured out what's going on.

I think something weird happens when the Dekker method converges after 0 iterations. In this case, the final function evaluation is for the upper bound. So the transpiration rate (which is updated as a side-effect of `func`) is not accurate.

If I add another line to `c3_leaf_photosynthesis` (see below), the new simulation results are identical to the old ones for `soybean2`:

```c++
    // Throw exception if not converged
    if (!is_successful(result.flag)) {
        throw std::runtime_error(
            "c3_leaf_photosynthesis solver reports failed convergence:\n    " +
            result.message());
    }

    // Re-run the function to make sure results are up-to-date
    double unused = func(photo.Gs);
```

Does this make sense? I haven't looked at the Dekker code itself to see if my guess for its operation is true. This also might explain how the residual is exactly zero (if it is not updated properly).

Is there something we can change in the Dekker method itself to prevent this from happening? Otherwise, we need to be very careful about functions that update things as side-effects. It seems like this issue could also affect `c3photoC`, `c4photoC`, and the evapotranspiration code, which also use the Dekker method.

eloch216:
If the Dekker method converges after 0 iterations, that means the function's value at the `initial_guess` is close enough to zero (within the absolute tolerance). In that case, there is no need to even evaluate the function at the endpoints, right? Maybe this behavior could be enforced for bracketing methods (if the initial guess is a solution, don't evaluate the function at the endpoints).

leavesandlemmas:
> If the Dekker method converges after 0 iterations, that means the function's value at the `initial_guess` is close enough to zero (within the absolute tolerance). In that case, there is no need to even evaluate the function at the endpoints, right? 

The iteration count > 0 means the initial conditions allowed for at least one iteration.  Here's the relevant code for the dekker method's initialization (there's a second one).  It actually checks whether any of the 3 points given forms a bracket. but the side effects create problems.   The algorithms assume that the function does not have any side effects so that the pair (x, y) can be passed around or swapped without worrying about whether external values are updated or what the order of evaluation is. So here's the contrapoint is the last point to be evaluated so the transpiration is populated by that rather than the first initial guess, even though the algorithm will terminate on the initial guess.
 
```cpp
 template <typename F>
    bool initialize(F&& fun, double a, double b, double c)
    {
        best.x = a;
        best.y = fun(a);

        last.x = b;
        last.y = fun(b);

        contrapoint.x = c;
        contrapoint.y = fun(c);

        if (same_signs(best.y, contrapoint.y)) {
            if (same_signs(best.y, last.y)) {
                flag = Flag::invalid_bracket;
                return false;
            }

            std::swap(last, contrapoint);
        }

        if (smaller(contrapoint.y, best.y)) {
            std::swap(best, contrapoint);
        }

        if (is_zero(best.y)) {
            flag = Flag::residual_zero;
            return false;
        }

        return true;
    }

```

I am thinking the simplest fix is  just reevaluate the lambda function at the root after the root solver terminates to ensure the correct side-effects.  I am not sure of a fix to the algorithms that would account for side-effects. The algorithms all implicitly assume a mathematical function (which can't have side-effects) so not sure that modifying the algorithms is the best solution. Another possibility is to rewrite the lambda function using some other struct / design so that there are no side-effects. But that requires more work.

leavesandlemmas:
Sorta surprised we havent had problems with side-effects before (or maybe we have had, but just hadn't noticed?)

eloch216:
I think we have had problems with side-effects in the past. I could search through and find the details if you are interested.

It seems very easy to forget to re-evaluate the lambda after the loop converges. I think it would be better to make sure the solvers behave in a more "friendly" way, since most of the applications of the solvers use functions with side-effects.

See this PR: https://github.com/biocro/biocro/pull/329

It seems there are issues with both the fixed-point and Dekker solvers, but not with others.

eloch216:
> I am thinking the simplest fix is just reevaluate the lambda function at the root after the root solver terminates to ensure the correct side-effects. I am not sure of a fix to the algorithms that would account for side-effects.

The re-evaluation could be done inside the solve method, so it always happens and we don't need to remember it. It's not the most efficient solution but it's simple.

yufenghe22:
> > I am thinking the simplest fix is just reevaluate the lambda function at the root after the root solver terminates to ensure the correct side-effects. I am not sure of a fix to the algorithms that would account for side-effects.
> 
> The re-evaluation could be done inside the solve method, so it always happens and we don't need to remember it. It's not the most efficient solution but it's simple.

Can we modify the [make_result](https://github.com/biocro/biocro/blob/91fbda2e362c165302e61e905d8ed3ad675adb69/src/math/roots/onedim/roots.h#L257) function to accept a `func` like:
```
    template <typename F>
    result_t make_result(size_t iteration, F&& func)
```

and then call the `func` one more time to get the final values?

yufenghe22:
Since the side-effects are not due to the solver itself but due to the evaluation functions, it's probably making more sense and easier to just add another evaluation at the end of root solution.
This applies to `c3CanAC`, `CanAC`, `c3_leaf_photosynthesis`, `c4_leaf_photosynthesis`, `c3photoC`, and `c4photoC` when their residual lambdas mutate output state. `leaf_energy_balance` is already safer because it calculates its final outputs explicitly from `result.root` rather than depending on the last residual evaluation.
For example, in `c3photoC`, we need a line like this after  `double const Cc = result.root;                                         // micromol / mol`
```
    // Synchronize FvCB_res, BB_res, Gs, and Assim with the accepted root.
    check_assim_rate(Cc);
```
I can make these changes if this looks good for everyone.

justinmcgrath:
When this was first done, I made the point that the design did not represent the abstraction, it was unnecessarily complicated, and it was likely to cause problems with use. I cannot think of any reason that side effects would be necessary here, and in general they should be avoided. The solutions here are workarounds to accommodate the fundamentally incorrect design.

eloch216:
> it's probably making more sense and easier to just add another evaluation at the end of root solution.

This isn't my favorite solution because it's something we need to remember to do any time we use one of the solvers.

Check out this PR, which alters the solvers a tiny bit so they check each input value (initial guess and/or bracket bound) and immediately stop if one of them is a root. It makes the solvers more efficient, and solves the problem occurring here: https://github.com/biocro/biocro/pull/330

As Justin said, it would also be best to not use a function with side-effects, since they are confusing. (Scott also suggested this.) Note that we already take this approach with the code in `leaf_energy_balance.cpp`. (There is a lambda, but it only captures by value so it does not alter any pre-existing variables when it is evaluated.)

I could work on switching the other lambdas to use this style instead. Then we would no longer use side-effects, but the solver algorithms would also be robust to any accidental use of side-effects in the future.

yufenghe22:
> It makes the solvers more efficient, and solves the problem occurring here: #330

It's pretty cool to make the solvers more "intelligent". However, I wonder what's the convention like for general solvers like Boost. I assume they would not care about the side-effects. It's still the users' awareness to make sure the lambda function is properly designed or the final output is synchronized with the `root`.

eloch216 (approved):
This should be good to go once the checks are done running. It is a nice improvement to no longer have so many convergence failures!

yufenghe22 (approved):

PR #329, opened 2026-07-31
Add new tests of the 1D root finders

eloch216:
This PR adds some new tests for the 1-D root solvers.

Specifically, each solver is now expected to behave such that when a root is found, the final function evaluation must be one where that root value was passed to the function. This must be the case even when the initial guess is correct.

This is intended for "test-driven development" because the test does not currently pass.

The following code can be used to test just the root finders:

```r
library(BioCro)
module <- 'BioCro:root_onedim_test'
cases <- cases_from_csv('BioCro:root_onedim_test', 'biocro/tests/module_test_cases/')
sapply(cases, function(x) {test_module(module, x)})
```

Right now, the code above produces the following output:

```
Differences found for Module `BioCro:root_onedim_test` test case `Kepler Equation Test Case`:

                 Quantity Expected     Actual
1 fixed_point_last_x_diff        0 0.08904276


Differences found for Module `BioCro:root_onedim_test` test case `Initial guess is correct (otherwise based on case above)`:

               Quantity Expected Actual
1 dekker_sg_last_x_diff        0      2

[1] "Module `BioCro:root_onedim_test` test case `Kepler Equation Test Case`: calculated outputs do not match expected outputs"                               
[2] "Module `BioCro:root_onedim_test` test case `Initial guess is correct (otherwise based on case above)`: calculated outputs do not match expected outputs"
```

So, there is an issue with the fixed-point solver (when the initial guess is not correct) and with the Dekker solver (when the initial guess is correct).

Since the root occurs for `x = 1`, these results mean that the fixed-point solver most recently evaluated the function at `x = 1 + 0.08904276 = 1.08904276` and the Dekker method most recently evaluated it at `x = 1 + 2 = 3`. For the test case, the upper limit for the bracket is 3, which seems to explain the results from the Dekker method.


eloch216:
I added a "meta-check" for the solvers to make sure the test cases are properly defined.

Otherwise, it would be possible for one of the `_x_last_diff` values to be nonzero, but the test would pass as long as it was nonzero in the case definition file.

eloch216:
Realized I should also check for issues when one of the bracket values is a root. Now we have problems with most of the methods:

```

Differences found for Module `BioCro:root_onedim_test` test case `Kepler Equation Test Case`:

                 Quantity Expected     Actual
1 fixed_point_last_x_diff        0 0.08904276


Differences found for Module `BioCro:root_onedim_test` test case `Initial guess is correct (otherwise based on Kepler case above)`:

               Quantity Expected Actual
1 dekker_sg_last_x_diff        0      2


Differences found for Module `BioCro:root_onedim_test` test case `Lower bracket is root (otherwise based on Kepler test above)`:

                      Quantity Expected  Actual
1  anderson_bjorck_last_x_diff        0 2.00000
2        bisection_last_x_diff        0 2.00000
3           dekker_last_x_diff        0 2.00000
4    dekker_newton_last_x_diff        0 2.00000
5        dekker_sg_last_x_diff        0 2.00000
6      fixed_point_last_x_diff        0 1.96649
7         illinois_last_x_diff        0 2.00000
8          pegasus_last_x_diff        0 2.00000
9     regula_falsi_last_x_diff        0 2.00000
10          ridder_last_x_diff        0 2.00000
11          secant_last_x_diff        0 2.00000


Differences found for Module `BioCro:root_onedim_test` test case `Upper bracket is root (otherwise based on Kepler test above)`:

                 Quantity Expected     Actual
1 fixed_point_last_x_diff        0 0.08904276

[1] "Module `BioCro:root_onedim_test` test case `Kepler Equation Test Case`: calculated outputs do not match expected outputs"                                      
[2] "Module `BioCro:root_onedim_test` test case `Initial guess is correct (otherwise based on Kepler case above)`: calculated outputs do not match expected outputs"
[3] "Module `BioCro:root_onedim_test` test case `Lower bracket is root (otherwise based on Kepler test above)`: calculated outputs do not match expected outputs"   
[4] "Module `BioCro:root_onedim_test` test case `Upper bracket is root (otherwise based on Kepler test above)`: calculated outputs do not match expected outputs"   
```

PR #330, opened 2026-07-31
More checks for roots in initializer methods

eloch216:
I figured out how to check the lower bracket, upper bracket, and starting guess to see if they are roots. As soon as a root is found, the solver terminates. This makes the solver slightly more efficient, and addresses the issues raised in PR #329 and PR #328 .

I will also do this for the other solvers, but I have to step away from my computer for a little while.


eloch216:
Here I have updated the `initialize` method for the relevant solvers so that each provided input value is checked to see if it is a root. If it is, the process is immediately ended without evaluating the function again. This ensures that the most recent function evaluation is always at the root. Note that this is not a major change, since most of the initializers already checked for roots. The difference is that the checks now happen sooner.

The fixed-point iteration solver required a slightly different approach. Here, the solver now returns `x` rather than `y = fun(x)` as the root. This also ensures that the most recent function evaluation is always at the root.

With these changes, the tests introduced in PR #329 are satisfied. It also reduces the number of test failures compared to PR #328. Here, the regression tests for miscanthus, willow, and `soybean` all pass. Changes are needed for `soybean2` but they are minor roundoff errors that don't substantially change the biomass predictions. (See below.)

image

leavesandlemmas (approved):

PR #331, opened 2026-08-03
No lambda side effects

eloch216:
This PR rewrites the lambdas used to solve photosynthesis equations so they have no side effects.

It also introduces a new free function (`solve_c3_gs`) that can be called by both `c3_leaf_photosynthesis` and `c3CanAC`, ensuring they use the same solver parameters and reducing some duplicated code. Likewise for `solve_c4_gs`.

While working on this, I discovered that different upper bracket bounds had been used in `c3_leaf_photosynthesis` and `c3CanAC`. Now they both use the same limits. This caused very tiny changes to the output of `c3_canopy` (which now uses the same bounds as `c3_leaf_photosynthesis`). Likewise for `c4_canopy`.


yufenghe22 (approved):
Looks good to me. I think these changes also improve the readability.

PR #332, opened 2026-08-03
Relax Euler requirement and update some units

eloch216:
I have been pondering whether or not the new soil water modules require an Euler solver with a step size of 1.

I had previously taken a hard-line stance, believing that an Euler solver was required. After some time, I have convinced myself that this isn't the case. I will explain my reasoning below.

----

Say we have a simple ODE representing exponential decay: $\frac{dy}{dt} = -k \cdot y$ where $k > 0$ is the decay constant. As pseudocode, this could be written as `rate = -k * y;`.

Suppose we are solving this with an Euler solver using a step size of 1. As code, this could be written as:

```
rate = -k * y_0;
delta = rate * 1;
y_1 = y_0 + delta;
```

If `k` is large enough such that `abs(delta) > y_0`, the solver will run into problems, exhibiting oscillations instead of exponential decay. The best way to fix this would be to simply use a smaller time step. However, we could do something like this in the code:

```
rate = -k * y_0;
delta = rate * 1;
if (y_0 + delta <= 0) {
    delta = -0.99 * y_0;
}
y_1 = y_0 + delta;
```

This is a "fix" that would prevent the oscillatory behavior when using a solver with timestep of 1. However, we could also look at it as simply defining a new differential equation, which could be written piecewise as $\frac{dy}{dt} = -k \cdot y_0$ if $k < 1$ and $\frac{dy}{dt} = -0.99 \cdot y$ if $k \geq 1$.

If this new model agrees with experimental data, then it is reasonable to use it, even though its origins were as a "fix" to a numerical issue.

I think that's what's happening with the DSSAT code. Some of it was motivated by thinking of an Euler solver -- especially the many checks to see if a total change over an hour will exceed a threshold. But we can just think of these limits as being part of the model equations. In that case we are free to use any solver we wish. The checks aren't always guaranteed to work for every solver -- e.g. an Euler solver with a 2 hour time step might not work well. But that's a separate issue.


eloch216:
I also noticed that some of the outputs returned by soil-water-related differential modules were not actually derivatives -- they were differences. It is easy to fix this by just dividing the difference by the time interval over which the difference occurs. The interval is always 1 for these modules, so this has no numerical impact at the moment, but it's important from a conceptual point of view.

leavesandlemmas:
> However, we could also look at it as simply defining a new differential equation, which could be written piecewise as [...] .

It seems to me that this is only true if the timestep is always exactly one, so I don't see how that weakens the euler step requirement.  The maximum `delta` would be different for different integration explicit schemes too. So you would get a different piecewise differential equation.

>  if this new model agrees with experimental data, then it is reasonable to use it, even though its origins were as a "fix" to a numerical issue.

> I think that's what's happening with the DSSAT code. Some of it was motivated by thinking of an Euler solver -- especially the many checks to see if a total change over an hour will exceed a threshold. 

I don't know if they were thinking of a euler method per se. It is possible to phrase any model with a timestep baked in it without explicit reference to a differential equation (like the logistic map). Like BioCro assumes the model has the form $\dot yt = f(t , y)$ (vectorfields/differential equations) but you could instead have models of the form $y(t + 1) = F( y(t) )$ where they advance the system's state by a fixed timestep (an iterated map instead of a vectorfield/differential equation). The checks ensure that the mapping sends a valid state to another valid state, but the resulting map may or may not have a differential equation (like the logistic map). Though obviously a discrete time model has to agree with some continuous time model since time is continuous. 

> But we can just think of these limits as being part of the model equations. In that case we are free to use any solver we wish.

I don't think this is true for differential equations because the timestep at which invalid states are generated varies with the integration method and with the stiffness of the equations.  It is true if the model's equations are fundamentally thought of as difference equations / iterated maps though. 

In your example, the discrete time model $y \mapsto ay$ for  $0 < a<1$ gives the same behavior as the continuous time model. But we could think of  $y \mapsto ay$ as fundamental where $a$ has some implicit timestep dependence, just fit $a$ to the data.

eloch216:
> > However, we could also look at it as simply defining a new differential equation, which could be written piecewise as [...] .
> 
> It seems to me that this is only true if the timestep is always exactly one, so I don't see how that weakens the euler step requirement. The maximum `delta` would be different for different integration explicit schemes too. So you would get a different piecewise differential equation.

I agree that choosing a different time step or integration method would result in a different limit on `delta`. But, my perspective is that once a limit is placed on `delta`, a new differential equation has been implicitly defined. We are free to solve that differential equation using any method we wish. If its predictions sufficiently agree with reality, it can be considered to be a useful scientific model.

> I don't know if they were thinking of a euler method per se. It is possible to phrase any model with a timestep baked in it without explicit reference to a differential equation (like the logistic map). Like BioCro assumes the model has the form $\dot yt = f(t , y)$ (vectorfields/differential equations) but you could instead have models of the form $y(t + 1) = F( y(t) )$ where they advance the system's state by a fixed timestep (an iterated map instead of a vectorfield/differential equation).

I agree that they weren't thinking of the Euler method, and instead they had defined an iterative model that proceeds in daily time steps.

My point here is that if you have a set of differential equations and then apply an integration method (such as Euler method with a time step of one hour) to it, you have created an iterative model. This can be thought of as a method for converting a differential equation to an iterative model. But the process can also be reversed -- if you can find a differential equation and integration method that reproduce the original iterative model, you have effectively converted it to a differential equation.

There is not a unique differential equation corresponding to each iterative model. E.g. if I assume an Euler solver with a time step of thirty minutes, I might end up with a different differential equation than if I had assumed an Euler solver with a time step of one hour. But the point still stands that a differential equation can be created, and now it exists independently of the process by which it was "reverse engineered" from an iterative model.

That's what I believe we have done here. We have taken the original DSSAT model, which was an iterative model using a daily time step, and "reverse engineered" a set of differential equations. The differential equations would only agree perfectly with the iterative model when using a particular solver. But that's okay as long as they continue to make reasonable predictions with other solvers.

yufenghe22:
Not sure if I understand this properly, but do we need to check if different time steps and RK4 can produce similar results assuming the 1-hour Euler is the benchmark since it has shown to produce reasonable/realistic results? Then we are confident the "underlying" ODE is solver-independent.

leavesandlemmas:
I am confused. I thought the euler requirement was required for those modules was that they had checks and other logic built on the implicit time step. You're saying they are now time-step independent ??

eloch216:
> I am confused. I thought the euler requirement was required for those modules was that they had checks and other logic built on the implicit time step. You're saying they are now time-step independent ??

The option to designate a module as requiring an Euler solver was originally added because of the `BioCro:thermal_time_senescence` and `BioCro:thermal_time_and_frost_senescence` modules. These are a bit weird. At each time point, they keep a record of how much biomass was added to each tissue through growth. When the threshold for senescence is reached, they look back in time and start "undoing" the biomass that was added through growth by subtracting it from each tissue. Because they need to keep track of the past state at each time step, they absolutely only make sense with an Euler solver. Obviously there are a lot of issues with this approach to senescence, and nowadays we would not permit such modules to be added to the library. But, before Megan added the logistic senescence module, this was how BioCro had always modeled senescence.

In this discussion, I am arguing that simply having checks that were derived by assuming a particular time step does not automatically restrict a model to only Euler solvers. There is some precedent for that philosophy. The `BioCro:partitioning_growth` module has a rhizome check that only makes sense in an Euler context, but that module is not marked as requiring an Euler solver.

If the DSSAT-inspired modules are marked as requiring an Euler solver, this will cause issues going forward. These modules will be used in all of our models because they are much more reliable than our previous treatment of soil water. So this would effectively prevent us from ever using another solver. At the same time, models with substrate carbon pools seem to not work well with Euler solvers.

eloch216:
> Not sure if I understand this properly, but do we need to check if different time steps and RK4 can produce similar results assuming the 1-hour Euler is the benchmark since it has shown to produce reasonable/realistic results? Then we are confident the "underlying" ODE is solver-independent.

I tried checking this. There is a `c3photoC` convergence error when the Euler step size is 2.0, which prevents the simulation from running. It seems that the soil water content in the top layer does not converge until the Euler step size is below 0.5. But the impact on biomass is very small. See figure below.

I'm not sure this is very conclusive. In my opinion, it just means that the numerical solution to the system of ODEs has not converged to the real one when the Euler step size is 1.0. No set of ODEs is "solver-independent." I don't know if this result can really tell us whether it's okay to relax the Euler requirement.

Soil water content figure:

Picture1

Grain figure:

Picture2

Code I used:

```r
# Load libraries
library(BioCro)
library(PhotoGEA)
library(lattice)

# Helping function for running the soybean2 model with a particular ODE solver
test_ode_solver <- function(os, os_name) {
  tmp_res <- tryCatch(
    run_biocro(
      soybean2$initial_values,
      soybean2$parameters,
      soybean_weather[['2002']],
      soybean2$direct_modules,
      soybean2$differential_modules,
      os
    ),
    error = function(e) {
      print(os_name)
      print(e)
      run_biocro(
        soybean2$initial_values,
        soybean2$parameters,
        soybean_weather[['2002']][1, ],
        soybean2$direct_modules,
        soybean2$differential_modules,
        os
      )
    }
  )
  
  within(tmp_res, {solver = os_name})
}

# Define ODE solver scenarios
scenarios <- list(
  list(default_ode_solvers[['homemade_euler']],                                'homemade_euler'),
  list(within(default_ode_solvers[['boost_euler']], {output_step_size = 2.0}), 'boost_euler step size 2.0'),
  list(default_ode_solvers[['boost_euler']],                                   'boost_euler step size 1.0'),
  list(within(default_ode_solvers[['boost_euler']], {output_step_size = 0.5}), 'boost_euler step size 0.5'),
  list(default_ode_solvers[['boost_rk4']],                                     'boost_rk4'),
  list(default_ode_solvers[['boost_rkck54']],                                  'boost_rkck54')
)

# Run scenarios and get full results
full_res_list <- lapply(scenarios, function(x) {do.call(test_ode_solver, x)})

full_res <- do.call(rbind, full_res_list)

# Make sure traces are ordered as they were defined
snames <- sapply(scenarios, function(x) {x[[2]]})

full_res$solver = factor(full_res$solver, levels = snames)

# Helping function for plotting results
plot_var <- function(varname) {
  pdf_print(
    xyplot(
      full_res[[varname]] ~ full_res[['fractional_doy']],
      group = full_res[['solver']],
      type = 'l',
      auto = TRUE,
      xlab = 'Day of year',
      ylab = varname,
      par.settings = list(
        superpose.line = list(col = multi_curve_colors())
      )
    ),
    width = 11
  )
  
  pdf_print(
    xyplot(
      full_res[[varname]] ~ full_res[['fractional_doy']] | full_res[['solver']],
      type = 'l',
      auto = TRUE,
      xlab = 'Day of year',
      ylab = varname
    ),
    width = 11
  )
}

# Plot results
to_plot <- c(
  'soil_evaporation_rate',
  'soil_water_content_1',
  'soil_water_content',
  'Grain'
)

lapply(to_plot, plot_var)
```

leavesandlemmas:
> a `c3photoC` convergence error when the Euler step size is 2.0, which prevents the simulation from running

That seems fishy to me. The c3photoC 's solution should not depend on the timestep, but maybe the euler method has generated some invalid state?

> No set of ODEs is "solver-independent."

The ODEs are independent of the method used to compute solutions, no?  Like for a vector field $\dot y  = f(t, y)$ , $f$ is independent of the method used to approximate a solution $y(t)$, right?  You mean that the approximate solution is not independent of the method used to compute it ? 

What is the euler requirement mean mathematically exactly? That the differential module must return a new valid $y$ instead of a $\dot y $ or that it stores state ? Like mathematically how would I tell ?  

>   The BioCro:partitioning_growth module has a rhizome check that only makes sense in an Euler context, but that module is not marked as requiring an Euler solver.

This check seems like it would not reliably prevent negative rhizomes.  I think the issue is that the ODEs may be totally valid but the approximate solution from an integration method generates invalid states (like negative biomass). But it is possible that the true solution to the ODE does not generate an invalid state, so it is only an issue with the method used to compute the approximate solution.  But it seems like the solver should have the invalid states, since its a problem with the computed solution rather than the solution defined by the ODE.  

I agree that we can reverse engineer ODEs which agree with an iterative map, but define a solution for all time points rather than just a discrete set of times. However, it seems to me that much of the logic that tries to prevent invalid states conflicts with the fact that the ODE/vector field function is supposed to return $\dot y$ rather than $y$ at the next time point. 

> At the same time, models with substrate carbon pools seem to not work well with Euler solvers.

Yeah, pretty much any transport model will be stiff in a way that the current biocro model is not. In fact, the basic model is not that stiff at all.

yufenghe22:
> I'm not sure this is very conclusive. In my opinion, it just means that the numerical solution to the system of ODEs has not converged to the real one when the Euler step size is 1.0. No set of ODEs is "solver-independent." I don't know if this result can really tell us whether it's okay to relax the Euler requirement.

Thanks. I think this is good evidence that the system is convergent as the timestep is reduced. I would not worry much about the top-soil water content being non-converged given that the biomasses look good. 

I repeated your case with Euler of step size of 2 and the error message says `Function returned a nonfinite value: NaN, Inf, -Inf.>`.
My guess is that it's from this part : `update(days_stage2_op, delta_days_stage2 / timestep);`. When the rate of change is negative, using a larger step can overshoot because the solver applies twice the calculated one-hour change. This can cause the resulting `days_stage2` to be negative. Then because it's used under a square root function, that leads to an `NAN`.

yufenghe22:
> What is the euler requirement mean mathematically exactly? That the differential module must return a new valid 
> 𝑦
>  instead of a $\dot y $ or that it stores state ? Like mathematically how would I tell ?

The issue is that `delta_days_stage2` is calculated by first constructing a next state over a fixed one-hour interval, so it is tied to a one-hour future target, rather than being derived directly as a local/pure instantaneous rate. One-hour Euler exactly reconstructs that target, and that's why initially we thought the Euler with one-hour step is a must-have. Other solvers can integrate on this `derived ODE`. They simply will not reproduce the original one-hour mapping exactly. Ed's tests suggest that smaller-step Euler, RK4, and RKCK54 can integrate it successfully and approach a stable solution (probably indicating a unique underlying ODE?).

DSSAT does not have this rate of change. It only integrates by a fixed interval (daily). The `delta_days_stage2` was later introduced/reverse-engineered to fit BioCro's solver scheme.

yufenghe22 (approved):

eloch216:
I am going to eventually merge this because it has been approved, but we can continue discussing a bit since I think there are some unresolved questions. See the final point below.

> > a `c3photoC` convergence error when the Euler step size is 2.0, which prevents the simulation from running
> 
> That seems fishy to me. The c3photoC 's solution should not depend on the timestep, but maybe the euler method has generated some invalid state?

> I repeated your case with Euler of step size of 2 and the error message says `Function returned a nonfinite value: NaN, Inf, -Inf.>`. My guess is that it's from this part : `update(days_stage2_op, delta_days_stage2 / timestep);`. When the rate of change is negative, using a larger step can overshoot because the solver applies twice the calculated one-hour change. This can cause the resulting `days_stage2` to be negative. Then because it's used under a square root function, that leads to an `NAN`.

I think Yufeng figured this out. It's not really a solver convergence issue. It's because an invalid result is returned from the soil evaporation model.

> > No set of ODEs is "solver-independent."
> 
> The ODEs are independent of the method used to compute solutions, no? Like for a vector field y ˙ = f ( t , y ) , f is independent of the method used to approximate a solution y ( t ) , right? You mean that the approximate solution is not independent of the method used to compute it ?

Yes, that's what I meant.

> What is the euler requirement mean mathematically exactly? That the differential module must return a new valid $y$ instead of a $\dot y$ or that it stores state? Like mathematically how would I tell?

The option to flag a module as requiring an Euler solver was added by me, and I don't think I had a super clear list of scenarios in mind where it would be appropriate. As I mentioned above, it was introduced to prevent the `BioCro:thermal_time_senescence` and `BioCro:thermal_time_and_frost_senescence` modules from being used with other solvers since they would produce some nonsense results with other solvers.

In response to your question, I have been thinking about it a bit more. I don't think the requirement has a mathematical meaning (or should have one). In other words, you cannot tell just by considering the equations used in a module.

I also think it is unrelated to the issue of returning $\dot y$ vs. a new valid $y$. This is settled by the definition of a differential module -- it must always return a time rate of change. We would need to introduce new module types if we wanted to allow modules to return $\Delta y$ or a new value of $y$.

Instead, I think the Euler requirement is purely related to the code implementation of a set of equations. If the code is written in such a way that it will return total nonsense when it is used with a different solver, then an Euler solver is required. I'm not sure if I feel confident enough to give a general set of criteria that could be used to identify such modules. I do think that a module that stores an ordered sequence of previous states would require an Euler solver. Maybe one way to phrase this would be that if a module requires that it is only evaluated once per time step, it requires an Euler solver. I am just not sure if this is the only scenario that would require an Euler solver.

As a side note, I think that most (or all) modules that require an Euler solver _could_ be re-written in such a way that they are compatible with any solver. However, it may not always be practical to do so.

If we can figure out a clear statement about this, it could be added to the BioCro guidelines to help clarify this in the future. I could do that in this PR before it is merged. Right now, my suggestion would be something like:

"If a module's code implementation is written in such a way that it will only return meaningful results when used with an Euler solver, it can be flagged as requiring an Euler solver. Most commonly, this occurs when a module stores an ordered sequence of previous state values under the assumption that the module is only evaluated once per time step. Such module implementations should be avoided whenever possible."

eloch216:
I added something to the contribution guidelines about module types. It can be refined later as needed.

I also added a check to detect the issue Yufeng identified above. Now, when I try to use a 2 hour time step with an Euler solver, the simulation fails partway through with this error message:

```
The ODE solver reports the following:
boost::numeric::odeint::integrate_const encountered an error and has returned a partial result:
Thrown in soil_evaporation_ritchie: days_stage2_next is negative.
```

This is more informative than a convergence error message, since the problem is not truly related to the `Ci` solver in `c3photoC`.

When the checks are done, I will merge this.

PR #333, opened 2026-08-03
Update solar_position_michalsky.h

eloch216:
A while ago I noticed a tiny error in the solar position module. It has no impact on the outputs, but I would like to fix it anyway.

### DIFF
--- a/src/module_library/solar_position_michalsky.h
+++ b/src/module_library/solar_position_michalsky.h
@@ -198,7 +198,7 @@ void solar_position_michalsky::do_operation() const
 
     // Unpack the doy and hour in UTC
     double const fractional_doy_utc = fractional_doy - time_zone_offset / hr_per_day;  // days
-    double const doy_utc = std::floor(fractional_doy_utc / hr_per_day);                // days
+    double const doy_utc = std::floor(fractional_doy_utc);                             // days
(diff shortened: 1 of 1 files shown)


src/module_library/solar_position_michalsky.h
```diff
@@ -198,7 +198,7 @@ void solar_position_michalsky::do_operation() const
 
     // Unpack the doy and hour in UTC
     double const fractional_doy_utc = fractional_doy - time_zone_offset / hr_per_day;  // days
-    double const doy_utc = std::floor(fractional_doy_utc / hr_per_day);                // days
+    double const doy_utc = std::floor(fractional_doy_utc);                             // days
```
yufenghe22, line 201: At least for the `doy_utc`, this has a big impact, which also impacts `hour_utc` significantly. It's good that you caught this.

yufenghe22 (approved):

src/module_library/solar_position_michalsky.h
eloch216: It does change `doy_utc` and `hour_utc` significantly, but it doesn't change any of the module outputs. Note that all the checks are passing, but no changes were required to any of the stored module or model tests. I guess that's because there are so many periodic functions in the solar position calculations.

PR #334, opened 2026-08-07
Update soybean script and parameterization

eloch216:
One last change related to the Dekker solver.

Because there are fewer convergence failures from the `gs` solver, this makes more regions of parameter space available to optimizers when parameterizing a model. This could potentially allow a better parameterization to be found. While working on this, I figured I should also update the script to follow the new approach we have been using with LD11 -- now the evolutionary optimizer is followed up by a Nelder-Mead optimizer.

I ran the updated parameterization script with three seeds (1234, 2345, 3456). One nice result is that I didn't encounter any convergence failures at all. That's really great and it seems they have been pretty much eliminated.

I was able to get a slightly better parameterization. The objective function value when evaluated with the original parameters was 0.1137801. After the 2000 iterations of the evolutionary optimizer, the values were 0.131258, 0.116578, and 0.121750. So none of these were an improvement over the original parameterization. But, after the Nelder-Mead part, we got down to 0.119645, 0.101495, and 0.1071855. So two of the seeds improved on the original parameterization. I choose the one with the lowest objective function value (seed 2345) to use.

It is a little bit different from the original one, but not hugely different:

image

I also updated the script to include comparison plots of ambient vs. elevated CO2. It seems there is still an overestimation of yield for elevated CO2.

soybean_validation_original

soybean_validation_reparameterized


yufenghe22:
PR #328 is a merge to the `develop`. Current PR #334 is merged to a new branch. Has the PR #334 adopted the new changes from PR #328?

inst/extdata/parameterize_soybean.R
```diff
@@ -100,14 +109,21 @@ if (!dir.exists(OUTPUT_DIR)) {
...
+
+# Specify Rdata file names
+RDATA_FILE      <- file.path(OUTPUT_DIR, 'optim_res.Rdata')
+RDATA_FILE_NMKB <- file.path(OUTPUT_DIR, 'optim_res_nmkb.Rdata')
 
 # Get Catm values for 2002 and 2005
 Catm_2002 <- with(BioCro::catm_data, {Catm[year == '2002']})
+Catm_2004 <- with(BioCro::catm_data, {Catm[year == '2004']})
 Catm_2005 <- with(BioCro::catm_data, {Catm[year == '2005']})
+Catm_2006 <- with(BioCro::catm_data, {Catm[year == '2005']})
```
yufenghe22: `Catm_2006` is incorrect.

yufenghe22 (approved):
> I ran the updated parameterization script with three seeds (1234, 2345, 3456). One nice result is that I didn't encounter any convergence failures at all. That's really great and it seems they have been pretty much eliminated.

This is really great. I think it suggests the underlying problems are solvable under varying environmental conditions, which is a big relief. For the eCO2, it can be left like this unless someone needs to use it now. There should be acclimations on the Vcmax and Jmax. I'm drafting the soil water paper, and will update the eCO2 parameterization later.

inst/extdata/parameterize_soybean.R
eloch216: Fixed!

eloch216:
> PR #328 is a merge to the `develop`. Current PR #334 is merged to a new branch. Has the PR #334 adopted the new changes from PR #328?

Yes, this branch is based on the one in PR #328

eloch216:
I will merge this once the checks have finished

yufenghe22:
For the record, here's what I get about biomasses at eCO2. I considered the following acclimations, which however does not seem to improve the overestimation much for eCO2. These acclimation parameters are based on Andrew (for RL) and Lisa's (for Vcmax and Jmax) papers.
```
 if(predefine_eCO2){
    #pre-define
    soybean_parameters0$RL_at_25    = soybean_parameters0$RL_at_25 * (1+0.37*0.42)
    soybean_parameters0$Vcmax_at_25 = soybean_parameters0$Vcmax_at_25*0.85
    soybean_parameters0$Jmax_at_25  = soybean_parameters0$Jmax_at_25*0.95
    soybean_parameters0$Catm  = 570
  }
```
image

PR #335, opened 2026-08-14
Misc small

eloch216:
I think we are almost ready for a new release. This PR just makes a few small last-minute changes:

1. The bounds for the `grc` parameters in `parameterize_soybean.R` were changed. This will prevent future problems. There's no need to re-run the script because the current parameter values are still within the new bounds.

2. I fixed a "maybe uninitialized" compiler warning that I've been seeing.

3. I saw a weird message in a recent `R CMD check` result on Ubuntu about `src/.deps` being included with the package. Not sure why this has started happening, but I tried adding it to `.Rbuildignore`.

PR #336, opened 2026-08-15
fix some compile issues on Mac

yufenghe22:
This PR fixes the macOS compilation error and related warnings reported in PR #335.
## Changes

- Replace variable-length arrays in `soil_water_flow_functions.cpp` with
  `std::vector`.
- Make the fixed six-layer count in `multi_layer_soil_profile` a `constexpr`
  constant.
- Make the `zero_finding_method` friend declarations consistently use `struct`.
- Treat the BioCro's copy of the Boost directory as a system include directory, suppressing
  clang's warnings originating from third-party Boost headers.

The change from `-I` to `-isystem` only affects compiler diagnostics and header
dependency tracking; Boost remains available from the same directory.

## Update
Looks like `-isystem` is not allowed here. So I just changed back to `-I`.

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -1,4 +1,4 @@
-PKG_CPPFLAGS+=-I../src/inc -DR_NO_REMAP
+PKG_CPPFLAGS += -isystem ../src/inc -DR_NO_REMAP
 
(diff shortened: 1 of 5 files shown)


eloch216:
Thanks! It looks like the boost "extension" warnings also disappeared.

PR #338, opened 2026-08-18
Release v3.4.0

eloch216:
It's been a while since our last release and there have been several significant changes made on the `develop` branch, so it's time for a new one!


eloch216:
See this page for details about making a new release: https://biocro.org/BioCro-documentation/latest/bookdown/contributing-to-biocro.html#new-release

eloch216:
See also the companion PR targeted at `develop`: https://github.com/biocro/biocro/pull/339

inst/extdata/optimize_StomataWS_linear.R
```diff
@@ -0,0 +1,262 @@
...
+# this has two steps: 1. we fit to each Gs point by adjusting the StomataWS value
+# to achieve a "perfect" fit with module c3_leaf_photosynthesis;
+# 2. we create a linear function between the optimized StomataWS and REW.
+#
+# When the script is done running, the new values of StomataWS_gradient and
+# StomataWS_intercept will be printed to the R terminal. If necessary, the
+# stored values in data/soybean2.R should be updated to the new ones, and a note
+# about the reparameterization should be included in man/soybean2.R
+#
+# Last used with BioCro version with commit: db12f4a0
```
yufenghe22: L15 seems outdated and unnecessary. Can be deleted.

inst/extdata/parameterize_soybean.R
```diff
@@ -10,56 +10,140 @@
...
+library(dfoptim)
 library(lattice)
 library(parallel)
 library(PhotoGEA)
 
-# Specify some settings
-NCORES  <- detectCores() - 1 # number of cores to use for parallel operation
-ITERMAX <- 2000              # number of optimizer iterations
+# Check the BioCroValidation version
+expected_version  <- '0.3.0-2'
```
yufenghe22, line 56: For this script, do we need a version check on the BioCro's version?

yufenghe22 (approved):

justinmcgrath:
Is there a more informative name than soybean2?

There a core directory under the module library now. This will make this difficult. People tend to organize things into hierarchies because hierarchies are simple, and we operate under the assumption that if we put complicated things into hierarchies they will become simple things. But biological systems are not hierarchies and they cannot be sensibly organized into to them. It is trying to impose a structure on the system that does not exist, and eventually it leads to needless complexity in the design. BioCro was organized like this previously, and the particulars of that implementation made it impossible to swap components. Here it won't make it impossible, but it will make things complicated.

If there is a reason to categorize modules, that can be done separately, in a file that tags modules with categories.

leavesandlemmas:
> There a core directory under the module library now. This will make this difficult. People tend to organize things into hierarchies because hierarchies are simple, and we operate under the assumption that if we put complicated things into hierarchies they will become simple things. But biological systems are not hierarchies and they cannot be sensibly organized into to them. It is trying to impose a structure on the system that does not exist, and eventually it leads to needless complexity in the design. BioCro was organized like this previously, and the particulars of that implementation made it impossible to swap components. Here it won't make it impossible, but it will make things complicated.

This directory was added because many of the files in the `module_library` do not contain any modules at all. So the directory isn't meant to impose a hierarchy on modules, but separate modules from non-modules.  The core pieces of the photosynthesis calculations are outside of modules, for instance. FvCB model isn't implemented as a module. 

I found it confusing as a new developer to figure out which file I needed to look at in order to understand what the actual mathematical model was (or where it was defined in the code).

justinmcgrath:
Everything in these directories is an equation that describes how to get some state variable from other state variables. Biological systems are complicated, and for anyone new it will take time to learn the code. Putting the code into separate directories doesn't make that easier. It makes it harder, by forcing the new person need to learn more directories, which have no sensible organization. And it looks like there are now new name spaces too. This is only added complexity. "Core" is a meaningless description here. What is a new person to make of it?

justinmcgrath:
The things that don't fit the concept of state equations well are AuxBioCro.cpp, module_library.h, and module_library.cpp. AuxBiocro has been an issue for a long time, and it's a lot of work to fix, and this organization this doesn't address it. There has been no good solution identified for sensibly putting module_library.h and module_library.cpp somewhere else. If there are other files here that are not state equations, then it would make sense for them to be somewhere else. Otherwise, this should be a flat directory, and any organization can go in documentation or other files.

leavesandlemmas (approved):

inst/extdata/optimize_StomataWS_linear.R
eloch216: Good catch

inst/extdata/parameterize_soybean.R
eloch216: I don't think we really need to. In most cases, the latest version of the script should be compatible with the latest version of BioCro. There are scenarios where this won't be true, or maybe where someone has switched branches but not yet reinstalled BioCro. But specifying a BioCro version here will be another thing that will be easy to overlook when making a new release.

eloch216:
> Is there a more informative name than soybean2?

How about `soybean_sw` where `sw` refers to "soil water"? The soil water modules are the main difference between `soybean` and `soybean2`.

> Otherwise, this should be a flat directory, and any organization can go in documentation or other files.

I can look into flattening it. If it seems like it will take a while to sort out, though, I think this should be left as a task for the next release.

yufenghe22:
> There has been no good solution identified for sensibly putting module_library.h and module_library.cpp somewhere else.

Just curious, what's preventing moving these files into a folder like `interface`? In my experience, I have to memorize the prefixes when looking for certain files. They way that how we use prefixes is some sort of hierarchies, imo.

justinmcgrath:
> > There has been no good solution identified for sensibly putting module_library.h and module_library.cpp somewhere else.
> 
> Just curious, what's preventing moving these files into a folder like `interface`? In my experience, I have to memorize the prefixes when looking for certain files. They way that how we use prefixes is some sort of hierarchies, imo.

As in biocro/src/interface? I don't know that that is more intuitive than having them in module_library. What is the thinking for putting them there? I don't remember the details but I recall having them in module_library had to do with allowing module libraries in separate packages.

There is certainly a hierarchy in the the overall package, and there are cases where hierarchies are good representations of what you're working on. However, in general, the equation sets for a model cannot be organized into a meaningful hierarchy. One can obviously put them into a hierarchy that makes sense for what's there. The problem is that later someone will certainly want to make a change that doesn't match those arbitrary divisions. Then they have to jump through hoops to work around the hierarchy, when if it were flat they could get what they wanted directly. My comment was about this specific folder, not the package as a whole. It may be sensible to move the R_* files into a different directory if that's what you mean. It doesn't seem like a big concern to have them where they are and so I've never thought about it. If there are concrete ideas for that, we can see if it makes sense, but it also seems like little benefit for the effort.

leavesandlemmas (approved):

justinmcgrath (approved):

justinmcgrath:
I didn't see any INFO messages from win-builder, but some are listed in cran-comments.md. Did you get these from a different system?

```
## R CMD check results
There were no ERRORs, WARNINGs, or NOTEs.

There were 2 INFOs:

- > checking C++ specification ... INFO
  > specified C++17

  A C++ specification is needed to ensure the source code is properly compiled

- > checking for GNU extensions in Makefiles ... INFO
  > GNU make is a SystemRequirements.

  GNU make extensions are used in `src/Makevars`.
```

eloch216:
> I didn't see any INFO messages from win-builder, but some are listed in cran-comments.md. Did you get these from a different system?
> 
> ```
> ## R CMD check results
> There were no ERRORs, WARNINGs, or NOTEs.
> 
> There were 2 INFOs:
> 
> - > checking C++ specification ... INFO
>   > specified C++17
> 
>   A C++ specification is needed to ensure the source code is properly compiled
> 
> - > checking for GNU extensions in Makefiles ... INFO
>   > GNU make is a SystemRequirements.
> 
>   GNU make extensions are used in `src/Makevars`.
> ```

Every operating system and version of R seems to produce slightly different messages. The ones in our `cran-comments.md` are from running `R CMD check` on "windows-latest using R version devel" (results here: https://github.com/biocro/biocro/actions/runs/32448163922/job/96671665766?pr=338). I would expect it to be the same as `win-builder`, but I guess it's not.

I don't really know which OS/R is best to use. I think it seems safer to address these in our comments just in case. I think it would be more of an issue for them to see an `INFO` and us not write anything about it.

justinmcgrath:
I've submitted it.

justinmcgrath:
It's approved and will be on CRAN soon.

PR #340, opened 2026-08-19
Updates to 3.4.0

eloch216:
This implements some changes that were requested in PR #338


eloch216:
I also noticed that there is a new `INFO` popping up the in the `R CMD check` results since our last submission to CRAN, so I added it to `cran-comments.md`.

eloch216:
I'm using the `R CMD check` results from "windows-latest using R version devel" as a benchmark for long examples.

On branch `release-v3.4.0`, it reports:

```
  * checking examples ... [24s] OK
  Examples with CPU (user + system) or elapsed time > 5s
                      user system elapsed
  partial_application 7.41   0.01    7.44
```

After marking some parts of that example as `donttest`, that message has disappeared. But now there is a new one:

```
  * checking examples ... [27s] OK
  Examples with CPU (user + system) or elapsed time > 5s
                         user system elapsed
  crop_model_definitions 5.59   0.05    5.66
```

Not sure why the total time would have increased, or why the `crop_model_definitions` example would now be long enough to trigger a message when it wasn't before. I think the "example timing" checks are not very repeatable. Anyway, I can try to shorten that example too just to make sure.

eloch216:
I no longer see any messages about examples taking too much time in any of the GitHub action check results

src/module_library/canopy_light_helpers.cpp
```diff
@@ -2,7 +2,7 @@
 #include <stdexcept>  // std::out_of_range
 #include <utility>    // std::move
 #include <string>
-#include "canopy_light_distribution.h"
+#include "canopy_light_helpers.h"
 namespace core
```
justinmcgrath: These should not be in a different namespace.

eloch216: Done!

eloch216: A few thoughts about namespaces:

1. At some point, there could potentially be naming conflicts if someone is working entirely in C++ and compiles two separate module libraries. For example, it's possible that each library could have a different definition of a function with the same name and arguments. Ensuring that all classes, functions, etc are in a namespace would help with that. But this doesn't seem like an immediate danger.

2. Currently, all the modules are in the `standardBML` namespace. I could see a potentially benefit for introducing another namespace for any code that is related to a module but is not a module. But we would need to decide on a good name for it, and ensure it is used consistently. I think `core` was an attempt at this, but only a few classes/functions were included in it.

eloch216: Side note: All modules are in the `standardBML` namespace, but there are also a few non-module things in there. So it is not currently a purely module namespace.

leavesandlemmas: Why not have a `biocro` namespace ?

justinmcgrath: > A few thoughts about namespaces:
> 
>     1. At some point, there could potentially be naming conflicts if someone is working entirely in C++ and compiles two separate module libraries. For example, it's possible that each library could have a different definition of a function with the same name and arguments. Ensuring that all classes, functions, etc are in a namespace would help with that. But this doesn't seem like an immediate danger.
> 
>     2. Currently, all the modules are in the `standardBML` namespace. I could see a potentially benefit for introducing another namespace for any code that is related to a module but is not a module. But we would need to decide on a good name for it, and ensure it is used consistently. I think `core` was an attempt at this, but only a few classes/functions were included in it.

Each module library has its own name space, and it's set when you set up a package from SkelBML. If people are using that template the first one shouldn't be a problem. We started using a namespace at that level in anticipation of potential conflicts with multiple libraries.

Namespaces are meant to prevent name clashes. Using them for organization in my mind only complicates things. I looked to see what people thought of using them for organization, and bad practice is the overall agreement: https://stackoverflow.com/questions/4207606/sub-namespaces-for-organization-bad-practice. I don't see the point of separating set of equations by whether we've written them in Module classes or in some other way. What is being addressed here? Maybe there is some other way to do it.

leavesandlemmas: FWIW, I introduced the namespace `core`, along with several other namespaces, to prevent name clashes, and not for organization. The file directories were introduced for organization, namely to separate code that might be used in several places (e.g. the `math` directory) and to files organized thematically (e.g., all root-finding methods are separated into distinct files but grouped into a single folder). 

I have encountered name clashes in BioCro before and using namespaces greatly reduces that possibility. I also introduced namespaces to allow shorter identifier names with less risk of name clashes. E.g., `result_t` struct  in root finding code has a namespace so that one is free to reuse that elsewhere while avoiding something longer like `root_finding_result_type` or whatever. 

I stand by what I said before about the reason for having non-module files separated from modules. It took me ages to locate the FvCB equations and understand how they were used. It was very much not obvious that that the file `FvCB_assim.h` defines the FvCB model while `FvCB.h` is simply a wrapper module. More over it's not particularly obvious to me that `c3photo.h` would define and solve the leaf-level balance equations. And I had the benefit of already knowing what BioCro was probably doing mathematically before looking at any code. Certainly, some arbitrariness is unavoidable. And yes, better documentation would help, but file paths are a form of documentation.

justinmcgrath: > Side note: All modules are in the `standardBML` namespace, but there are also a few non-module things in there. So it is not currently a purely module namespace.

This isn't a problem for conflicts. Namespaces are not meant to organize types, and it would be unusual for a namespace to contain only objects (in the general programming sense, not the narrow C++ sense) of one type. The namespace around the module libraries lets authors pick any names for any of the things in their module libraries, Modules or otherwise. Including different types in the namespace is necessary, because even if we found a different place for the module_library.cpp and module_library.h files, they'd need to be in the standardBML namespace, even though they are not Modules, or they would conflict with any other module library.

justinmcgrath: > Why not have a `biocro` namespace ?

If the framework were coupled with some other code, adding a namespace would eliminate the possibility of conflicts. I've coupled BioCro to another C++ program and the issue never came up, so there was never a need to do it. I doubt that the names in the framework would ever conflict with another project, but the change is simple, and that is exactly the sort of thing namespaces are intended for.

For the module libraries, it was clear early on that if you wanted to use two libraries in one C++ project there would necessarily be a conflict, so the namespace was added.

eloch216:
@justinmcgrath Do you have any more changes to suggest? If not, I can merge this and we can submit to CRAN.

justinmcgrath (approved):

justinmcgrath:
Sorry. Forgot to approve it. I did it now.

src/module_library/canopy_light_helpers.cpp
eloch216: > > Side note: All modules are in the `standardBML` namespace, but there are also a few non-module things in there. So it is not currently a purely module namespace.
> 
> This isn't a problem for conflicts. Namespaces are not meant to organize types, and it would be unusual for a namespace to contain only objects (in the general programming sense, not the narrow C++ sense) of one type. The namespace around the module libraries lets authors pick any names for any of the things in their module libraries, Modules or otherwise. Including different types in the namespace is necessary, because even if we found a different place for the module_library.cpp and module_library.h files, they'd need to be in the standardBML namespace, even though they are not Modules, or they would conflict with any other module library.

That makes sense. Then I guess another related point is that some functions/classes/etc are _not_ in the `standardBML` namespace. Should everything in `src/module_library` be in the namespace?

eloch216: > It was very much not obvious that that the file `FvCB_assim.h` defines the FvCB model while `FvCB.h` is simply a wrapper module. More over it's not particularly obvious to me that `c3photo.h` would define and solve the leaf-level balance equations.

I'm not sure if I entirely understand the problem here. Once you open `FvCB.h`, it should be pretty clear from the descriptive comment and the `#include` statement that the actual model equations can be found in `FvCB_assim.h` (and `FvCB_assim.cpp`). Likewise for `c3photo.h` -- its purpose should be pretty clear once you look at it.

It's hard for the name of a file to ever completely describe what's in it when we are dealing with biological models and I'm not sure we need to have that as a goal for ourselves.

eloch216:
> Sorry. Forgot to approve it. I did it now.

Great! I think we have done all the checks that are available to us, and the `release-v3.4.0` branch can be submitted to CRAN. Can you do it tomorrow?

Instructions are here: https://biocro.org/BioCro-documentation/latest/bookdown/contributing-to-biocro.html#cran-submission

src/module_library/canopy_light_helpers.cpp
justinmcgrath: > > > Side note: All modules are in the `standardBML` namespace, but there are also a few non-module things in there. So it is not currently a purely module namespace.
> > 
> > 
> > This isn't a problem for conflicts. Namespaces are not meant to organize types, and it would be unusual for a namespace to contain only objects (in the general programming sense, not the narrow C++ sense) of one type. The namespace around the module libraries lets authors pick any names for any of the things in their module libraries, Modules or otherwise. Including different types in the namespace is necessary, because even if we found a different place for the module_library.cpp and module_library.h files, they'd need to be in the standardBML namespace, even though they are not Modules, or they would conflict with any other module library.
> 
> That makes sense. Then I guess another related point is that some functions/classes/etc are _not_ in the `standardBML` namespace. Should everything in `src/module_library` be in the namespace?

I can't think of a reason to not have everything in the namespace. What things aren't in it though?

leavesandlemmas: > > It was very much not obvious that that the file `FvCB_assim.h` defines the FvCB model while `FvCB.h` is simply a wrapper module. More over it's not particularly obvious to me that `c3photo.h` would define and solve the leaf-level balance equations.
> 
> I'm not sure if I entirely understand the problem here. Once you open `FvCB.h`, it should be pretty clear from the descriptive comment and the `#include` statement that the actual model equations can be found in `FvCB_assim.h` (and `FvCB_assim.cpp`). Likewise for `c3photo.h` -- its purpose should be pretty clear once you look at it.
?
> It's hard for the name of a file to ever completely describe what's in it when we are dealing with biological models and I'm not sure we need to have that as a goal for ourselves.

I suppose it's hard for me to determine if its clear or not now. After two years of navigating through these files, I know where stuff is and the comment is plenty informative. But I remember being confused because my reading of the two biocro papers made me think that these should be modules, then I'd look at the module and find the module just calls some function defined somewhere else.  I now understand why that's the case, but it was not obvious to me when I started. The core models of BioCro predates the concept of modules so the modules mostly wrap the code for those models.

I got echoes of that experience trying to explain to Hridi what files she would need to look at to understand BioCro's the canopy light distribution model. It's tricky to figure out what file one needs to look at to decide what could cause the model to behave in some way. So I wasn't really suggesting that the files necessarily have longer names (though  `c3photo` . My point was that separating files based on aspects of the code's design (like module vs. not-a-module) automatically creates a list that communicates what is a module and what isn't (without even having to open the file and examine its contents).

leavesandlemmas: > I can't think of a reason to not have everything in the namespace. What things aren't in it though?

I had AI put together a list of identifiers that aren't in any namespace inside the `module_library` directory. There's quite a few things, and that's why I was putting new code into a namespace. 

| File | Identifiers (functions, structs, objects, etc.) |
|:---|:---|
| `AuxBioCro.h` | `ET_Str`, `ws_str`, `soilML_str`, `dbp_str`, `seqRD_str`, `seqRootDepth()`, `rd_str`, `rootDist()`, `frostParms`, `nitroParms`, `crop_phenology` |
| `BioCro.h` | `watstr()`, `SoilEvapo()`, `soilML()` |
| `CanAC.h` | `CanAC()` |
| `c3CanAC.h` | `c3CanAC()` |
| `c3photo.h` | `check_c3_assim_rate()`, `c3photoC()`, `check_c3_gs()`, `solve_c3_gs()`, `solo()` |
| `c4photo.h` | `collatz_assim()`, `check_c4_assim_rate()`, `c4photoC()`, `check_c4_gs()`, `solve_c4_gs()` |
| `c3_temperature_response.h` | `c3_temperature_response_parameters`, `c3_param_at_tleaf`, `c3_temperature_response()` |
| `ball_berry_gs.h` | `ball_berry_gs()` |
| `boundary_layer_conductance.h` | `leaf_boundary_layer_conductance_campbell()`, `leaf_boundary_layer_conductance_nikolov()`, `canopy_boundary_layer_conductance_thornley()` |
| `canopy_photosynthesis_outputs.h` | `canopy_photosynthesis_outputs` |
| `collatz_photo.h` | `collatz_result`, `collatz_photo()` |
| `conductance_helpers.h` | `g_to_mass()`, `g_to_molar()`, `sequential_conductance()` |
| `conductance_limited_assim.h` | `conductance_limited_assim()` (two overloads) |
| `eC4photo.h` | `eC4photoC()` |
| `FvCB_assim.h` | `FvCB_outputs`, `FvCB_assim()` |
| `leaf_energy_balance.h` | `energy_balance_outputs`, `check_leaf_temp()`, `leaf_energy_balance()` |
| `photosynthesis_outputs.h` | `photosynthesis_outputs` |
| `rasmussen_specific_heat.h` | `rasmussen_specific_heat_of_air()` |
| `respiration.h` | `growth_resp()`, `growth_resp_Q10()`, `maintenance_resp_Q10()` |
| `soil_evaporation_functions.h` | `surface_albedo()`, `reference_evapotranspiration()`, `potential_soil_evaporation()`, `evap_str`, `ritchie_s1_to_s2()` |
| `soil_water_flow_functions.h` | `max_soil_layers` (const), `infilWater_str`, `upwardFlo_str`, `tileDrain_str`, `infil()`, `satflo()`, `up_flow()`, `tile_flow()` |
| `stomata_outputs.h` | `stomata_outputs` |
| `temperature_response_functions.h` | `arrhenius_exponential()`, `Q10_temperature_response()`, `eyring_response()`, `peaked_arrhenius_response()`, `johnson_eyring_williams_response()`, `polynomial_response()` |
| `water_and_air_properties.h` | `molar_volume()`, `saturation_vapor_pressure()`, `TempToDdryA()`, `dry_air_density()`, `TempToLHV()`, `water_latent_heat_of_vaporization_henderson()`, `TempToSFS()`, `TempToCp()`, `vapor_density_from_pressure()` |

justinmcgrath: We used the class name "Module" as jargon for "set of state equations". The idea was that because the class imposes some C++ specific structure, that a specific name distinguished it from sets of equations in general. However, we tend to say "module" lowercase, for any set of equations in our code. So everything in that directory is a module (except AuxBioCro and the module factories), but only some things are Modules.

As far as wrappers, a common naming convention is concept.cpp and concept_impl.cpp, where impl means implementation. The free functions are our implementations, where they're used. Maybe that naming convention helps. I suspect that for some things it will break a bit, as I can see someone using the same X_impl.cpp in a couple different places. A table documenting these relationships would work. Maybe some other naming convention would work.

Everything I see in that list would appropriately be in the library namespace. I can't think of any situation when something in a module library would not belong in the library namespace, so we can make this the standard approach, and hopefully it's something that can be included in SkelBML so that there isn't heavy overhead to it. I wouldn't consider it a priority. We've already protected the obvious conflicts, and it's unlikely that different libraries would have the same names. Coupling inside C++ itself is also unlikely. So there's little chance of a conflict overall, and there isn't a pressing need to fix everything. We can start doing it for new things and when there aren't higher priorities we can fix what's here.

justinmcgrath: > Each module library has its own name space, and it's set when you set up a package from SkelBML. If people are using that template the first one shouldn't be a problem.

I checked skelBML and I'm wrong about this. There are some namespace things, but I don't see where it's putting the module_library classes in a namespace. It looks like we did this for standardBML but not skelBML.

eloch216: > > Each module library has its own name space, and it's set when you set up a package from SkelBML. If people are using that template the first one shouldn't be a problem.
> 
> I checked skelBML and I'm wrong about this. There are some namespace things, but I don't see where it's putting the module_library classes in a namespace. It looks like we did this for standardBML but not skelBML.

`skelBML` does introduce a namespace for modules. But, you won't see it just by looking at the `src` directory of the base repository. It will appear after the setup script has been run. You can find the raw files in the `script/templates` directory. In these files, instances of `%1$s` will be replaced by the user-supplied module library name.

So, for example, the example module file includes the package name in the header guard, and via a namespace: 

https://github.com/biocro/skelBML/blob/2c73dbb05b19071fd195132f961e83d7d827733e/script/templates/example_module.h#L1-L7

Of course, there is no way to guarantee that a user will continue to put new modules in the namespace.

PR #342, opened 2026-08-25
Consolidate atmospheric scattering calculations

eloch216:
On `develop`, there are currently four modules that each calculate the fraction of direct and diffuse light in solar radiation at the Earth's surface:
- `BioCro:shortwave_atmospheric_scattering`
- `BioCro:c3_canopy`
- `BioCro:c4_canopy`
- `BioCro:canopy_light_distribution`

This situation can be simplified by only making the calculations in one of the modules (`BioCro:shortwave_atmospheric_scattering`). Then the others can take `irradiance_direct_fraction` and `irradiance_diffuse_fraction` as inputs instead of calculating them from other quantities.

That's what I've done here. Note that with this change, there is no longer a need for the `atmosphere_scattering` struct.


eloch216:
Some of the stored test results needed to be updated, but only for very minor reasons. The stored willow and miscanthus results only needed some new columns to be added, to address these warnings:

```
Warning ('test.CropModels.R:97:1'): (code run outside of `test_that()`)
The `miscanthus_x_giganteus` simulation result contains columns that are not in the saved result. Is this intentional? Extra columns: irradiance_diffuse_fraction, irradiance_diffuse_transmittance, irradiance_direct_fraction, irradiance_direct_transmittance.

Warning ('test.CropModels.R:97:1'): (code run outside of `test_that()`)
The `willow` simulation result contains columns that are not in the saved result. Is this intentional? Extra columns: irradiance_diffuse_fraction, irradiance_diffuse_transmittance, irradiance_direct_fraction, irradiance_direct_transmittance.
```

The module test cases were failing because of the stringent tolerances we use for these tests. E.g., here is an example test failure message that occurred before I updated the stored results:

```
Differences found for Module `BioCro:canopy_light_distribution` test case `cumulative lai is 6`:

                   Quantity  Expected    Actual
1      shaded_absorbed_ppfd  26.49458  26.49458
2 shaded_absorbed_shortwave  22.40140  22.40140
3       shaded_incident_nir 103.74427 103.74427
4      shaded_incident_ppfd  31.17009  31.17009
5      sunlit_absorbed_ppfd 581.03069 581.03067
6 sunlit_absorbed_shortwave 166.70474 166.70473
7       sunlit_incident_nir 246.61887 246.61886
8      sunlit_incident_ppfd 683.56552 683.56549
```

yufenghe22 (approved):
The PR title is confusing. It's really about reduce/remove code repetition or redundancy.

eloch216:
> The PR title is confusing. It's really about reduce/remove code repetition or redundancy.

Good point. I will make a few other similar PRs after this one and will keep that in mind for the titles. I think of it as "consolidating" the calculations from multiple places to a single place.

justinmcgrath (approved):
This looks good. I am surprised to see structs in here used simply to cluster together data. That's not what structs are for, and it's antithetical to the approach we use. I don't know how they got added.

eloch216:
> This looks good. I am surprised to see structs in here used simply to cluster together data. That's not what structs are for, and it's antithetical to the approach we use. I don't know how they got added.

When I started working on BioCro, there were already structs being used for this purpose. For example, there are several in `AuxBioCro.h` that have been around forever: `ET_Str`, `ws_str`, `soilML_str`, etc...

For a free function that needs to return more than one numerical output, it seems like a good option. The other main option would be to accept pointers or references as inputs, and modify those variables as a side effect. But this is generally more confusing than using a struct.

Unless I am misunderstanding something, I think it is common to use structs this way. It is a design pattern called a "Data Transfer Object." E.g. https://gist.github.com/MangaD/7bbb9c1b30f5bdfc31f1b99c227e5c12

leavesandlemmas:
> When I started working on BioCro, there were already structs being used for this purpose. For example, there are several in `AuxBioCro.h` that have been around forever: `ET_Str`, `ws_str`, `soilML_str`, etc...

And also the c3photo and c4photo functions, the FvCB function, the ball_berry_gs function and the c3canAC and canAC functions also all return similar data structs.

justinmcgrath:
Structs are useful for defining a new aggregate type where the components make sense as a whole in a way that they don't individually.

Points are a common example.

struct Point2D {
 double x;
 double y;
}

If you remove or add to that struct it isn't the same thing any more. It is there to facilitate a mental model of the program, not simply to group two variables together. People use arrays and containers for that.

The way that these have been use in the BioCro code, simply to give an aggregate return type, can be convenient especially since C++ lacks that syntax. I wouldn't tend to do that for these particular cases, because arrays and containers handle that situation perfectly. However, that use of structs isn't that bad of a thing. But here it's not simply for aggregate return types; it's building classes out of these in a way that makes the code far more complicated and adds nothing to it. There's more to read and understand (structs in structs!?), and those additions decrease flexibility, not increase it. This code does what it did before, just with more parts, and it doesn't represent our mental model of mathematical models.

eloch216:
> The way that these have been use in the BioCro code, simply to give an aggregate return type, can be convenient especially since C++ lacks that syntax. I wouldn't tend to do that for these particular cases, because arrays and containers handle that situation perfectly. However, that use of structs isn't that bad of a thing.

I'm not sure I agree that arrays and containers handle the situation perfectly.

I think `std::vector<double>` is a pretty bad way to return multiple numeric values, since the caller of the function needs to know the particular order of the returned values. If the order changes, this can cause problems that are tough to troubleshoot.

I think `std::map<std::string, double>` or `std::unordered_map<std::string, double>` are better because elements can be accessed by a name, but they are still not perfect. Here the issue is that the caller might try to use an undefined key to access an element of the map. This will cause a runtime error.

A `struct` that's used as a "data transfer object" gives the advantage of accessing elements by name, as in `map` or `unordered_map`. Additionally, if the caller tries to access a member that isn't defined in the struct, this will cause a compilation error, which is faster and easier to detect than an "undefined map key" runtime error.

> But here it's not simply for aggregate return types; it's building classes out of these in a way that makes the code far more complicated and adds nothing to it. There's more to read and understand (structs in structs!?), and those additions decrease flexibility, not increase it. This code does what it did before, just with more parts, and it doesn't represent our mental model of mathematical models.

I agree that some of the structs here are extra complicated, especially when there are structs within structs. It will be possible to simplify this stuff. I am planning to do it in small pieces, so each PR is easier to follow and there is less chance of changing any behavior. This PR is one example.

PR #343, opened 2026-08-26
Consolidate shortwave calculations

eloch216:
Similar to PR #342, this one aims to consolidate some calculations that are repeated across multiple modules. Currently, the calculation of direct/diffuse PPFD/PAR/NIR incident on the canopy occurs in five different modules:
- `BioCro:incident_shortwave_from_ground_par`
- `BioCro:c3_canopy`
- `BioCro:c4_canopy`
- `BioCro:canopy_light_distribution`
- `BioCro:ten_layer_canopy_properties`

In this PR, it is now only calculated in one place: `BioCro:incident_shortwave_from_ground_par`.

Note that with this change, there is no longer a need for the `canopy_light::from_solar` constructor.

As with the other PR, the changes to test files here are only due to small rounding errors, and the addition of new inputs/outputs.


eloch216:
BTW, I have at least two more similar PRs in mind. I probably won't work on them right away, though.

yufenghe22 (approved):

Issue #346, opened 2026-09-03
c4 photosynthesis does not respond to water stress


femiguez:
The miscanthus model (C4 photosynthesis) does not appear sensitive to water stress. For example, see the following code:

```r
miscanthus_x_giganteus$parameters$wsFun <- 0

result0 <- run_biocro(
  miscanthus_x_giganteus$initial_values,
  miscanthus_x_giganteus$parameters,
  get_growing_season_climate(weather$'2005'),
  miscanthus_x_giganteus$direct_modules,
  miscanthus_x_giganteus$differential_modules,
  miscanthus_x_giganteus$ode_solver
)

miscanthus_x_giganteus$parameters$wsFun <- 1

result1 <- run_biocro(
  miscanthus_x_giganteus$initial_values,
  miscanthus_x_giganteus$parameters,
  get_growing_season_climate(weather$'2005'),
  miscanthus_x_giganteus$direct_modules,
  miscanthus_x_giganteus$differential_modules,
  miscanthus_x_giganteus$ode_solver
)

miscanthus_x_giganteus$parameters$wsFun <- 2

result2 <- run_biocro(
  miscanthus_x_giganteus$initial_values,
  miscanthus_x_giganteus$parameters,
  get_growing_season_climate(weather$'2005'),
  miscanthus_x_giganteus$direct_modules,
  miscanthus_x_giganteus$differential_modules,
  miscanthus_x_giganteus$ode_solver
)

miscanthus_x_giganteus$parameters$wsFun <- 3

result3 <- run_biocro(
  miscanthus_x_giganteus$initial_values,
  miscanthus_x_giganteus$parameters,
  get_growing_season_climate(weather$'2005'),
  miscanthus_x_giganteus$direct_modules,
  miscanthus_x_giganteus$differential_modules,
  miscanthus_x_giganteus$ode_solver
)

miscanthus_x_giganteus$parameters$wsFun <- 4

result4 <- run_biocro(
  miscanthus_x_giganteus$initial_values,
  miscanthus_x_giganteus$parameters,
  get_growing_season_climate(weather$'2005'),
  miscanthus_x_giganteus$direct_modules,
  miscanthus_x_giganteus$differential_modules,
  miscanthus_x_giganteus$ode_solver
)

library(arsenal)

cmp12 <- comparedf(result1, result2)
cmp23 <- comparedf(result2, result3)
cmp34 <- comparedf(result3, result4)
```
[...]

PR #347, opened 2026-09-08
Remove `wsFun`

eloch216:
In response to issue #346 , this PR removes the `wsFun` input because it is not actually used for anything. While searching through the code, I also noticed an unused function that uses `wsFun` (`compute_wsPhoto`) so I removed this as well.


eloch216:
Tagging @femiguez

justinmcgrath (approved):
wsPhotoCol isn't used anywhere either?

eloch216:
> wsPhotoCol isn't used anywhere either?

Nope. It was stored as the `rcoefPhoto` member of the `soilML` return value, but that member was never accessed anywhere.

PR #348, opened 2026-09-08
Add checks to make sure RH is not a percentage

eloch216:
This PR adds some checks to make sure RH is not expressed as a percentage, since that can cause calculation issues. See this comment for more information: https://github.com/biocro/biocro/issues/344#issuecomment-5592169144

To address this, I got a list of all modules that take `rh` as an input. Many of them use either `ball_berry_gs`, `leaf_energy_balance`, or both, so adding checks to these functions cover several modules:
- `BioCro:ball_berry`
- `BioCro:c3_assimilation`
- `BioCro:c3_canopy`
- `BioCro:c3_leaf_photosynthesis`
- `BioCro:c4_assimilation`
- `BioCro:c4_canopy`
- `BioCro:c4_leaf_photosynthesis`
- `BioCro:ten_layer_c3_canopy`
- `BioCro:ten_layer_c4_canopy`
- `BioCro:leaf_evapotranspiration`
- `BioCro:leaf_evapotranspiration_check`

Then there were a few others that needed separate checks:
- `BioCro:leaf_gbw_nikolov`
- `BioCro:rh_to_mole_fraction`
- `BioCro:soil_evaporation`
- `BioCro:soil_evaporation_ritchie`
- `BioCro:two_layer_soil_profile`
- `BioCro:water_vapor_properties_from_air_temperature`

PR #349, opened 2026-09-09
Update R-CMD-check-backend.yaml

eloch216:
This PR attempts to address the issue described here: https://github.com/biocro/biocro/pull/348#issuecomment-5592615789

### DIFF
--- a/.github/workflows/R-CMD-check-backend.yaml
+++ b/.github/workflows/R-CMD-check-backend.yaml
@@ -196,7 +196,7 @@ jobs:
         if: ${{ !inputs.dry-run }}
         uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: any::rcmdcheck
+          extra-packages: any::rcmdcheck, local::.
           needs: check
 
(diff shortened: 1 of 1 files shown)


eloch216:
I tried two approaches here and they were not helpful.

Looking through old records, this happened in the past on MacOS: https://github.com/biocro/biocro/pull/273#issuecomment-3992690602

Back then, the only solution I could find was to increase the minimum tested version of MacOS: https://github.com/biocro/biocro/pull/275

@yufenghe22 had suggested it could be related to the new release of R (https://github.com/biocro/biocro/pull/348#issuecomment-5593104735). That sounds plausible, but I think the issue would have fixed itself by now if that were the root cause.

I will see if it helps to increase the minimum supported version to R 4.2.0. It's annoying to do this simply because the online testing setup has changed, since the package is surely still compatible with R version 4.1.0. But without a way to reliably test, the safest thing to do is change our supported versions. R version 4.2.0 was released in April 2022 so hopefully most people are using newer versions now anyway.

PR #351, opened 2026-09-16
Reduce Module boilerplate and scope of maintenance changes.

justinmcgrath:
The heavy repetition in Module definitions is cumbersome, distracting, and doesn't match our mental model for Modules very well. I figured there was a way to do it with macros but I couldn't sort it out, yet AI handled it easily.

This PR has what I hope is a working example. The top of module_macros.h describes usage, but in pseudocode it's this.

define inputs
define outputs
pass those to a class definer which does all the boilerplate of mapping pointers to keys, defining get_inputs, get_outputs, etc.
define do_operation.

I changed thermal_time_bilinear as an example. That one includes do_operation() in the header, but as now, you could put that in its own file.

I view this as filling in the inability to have static virtual functions in base classes. This macro wraps the Module class in a template class (template in the general sense, not C++ template).

This does a few things I like.

- There's no repetition. It would be great if we could just define do_operation and there were a way smart enough to get the inputs and outputs from that, but this is the second best option.
- It reduces these files to something that represents very well our concept of modules: inputs, outputs and the equations that relate them.
- By wrapping each class in a macro, we bring all of the class-level changes to a single place. If we change something like how we map the pointers, that change is done in a single place instead of for every Module file.

There are some downsides I see.

- There's another layer of complexity - we add this macro class.
- My personal experience is that I'm not very familiar with macros and I expect that's similar for others. This usage I found easy enough to follow; it's identical to using function pointers. But, if someone wants to modify this, they need to figure it out.

To me, the substantially reduced repetition and the potential to reduce maintenance if we change the class design outweigh the added wrapper. There could be some costs I'm not considering though.

As an example before doing anything in earnest, I put the macros in their own file. I would say that these should go in module.h though, as they are part of that class design. This PR would not be merged. We'd add the macros to `framework` and update the modules in `biocro`.


justinmcgrath:
One thing I wasn't sure about was allowing the C++ name and the map key to be different. For inputs, we usually use the same names. For outputs usually append "_op". I think there are a couple places where we don't do that though, and the version in this PR allows something like this.

```c++
#define THERMAL_TIME_BILINEAR_INPUTS(X)                                                \
    X(some_cpp_name, "some_different_key_name")                /* days */                      \
```

I'm not sure why we'd want that. We could have the macros enforce the scheme we're using now, and then we could have macros like this.

```c++
#define THERMAL_TIME_BILINEAR_INPUTS(X)                                                \
    X(fractional_doy)                /* days */                                                           

#define THERMAL_TIME_BILINEAR_OUTPUTS(X)                                               \
    X(TTc)                                                                /* degrees C * day / hr */
```

It's simpler to define, but I wasn't sure that we wanted to enforce that naming scheme.

justinmcgrath:
Maybe we could put the units in the macro too, and store that with the class. I don't know if that's useful.

yufenghe22:
Looks like a good idea. There will be lots of work to apply the macro to all modules, and some modules may need more changes to fit the macro. Shall we do it one module by one module or let AI handles it for all?

justinmcgrath:
We could change a handful to see if it's what we like, then once we decide on a plan change the rest. We should decide whether we want to be able to use different names and include units before editing every module, since changing them later requires editing every  module file.

eloch216:
This is pretty interesting and seems like it could be a good idea. A few comments/thoughts:

1. As I understand it, these macros are a "convenience" and would not actually required to define a module class, correct? I just want to double check because we have some modules that might not be compatible with them, such as `BioCro:ten_layer_c3_canopy`.

2. If we decide to use this for most modules, we will need to adjust `script/create_quantity_table.R` to make sure it can find the units in the macro code. At the moment, it only looks for units in the `get_inputs()` and `get_outputs()` functions.

3. If we decide to use this, we may also want to adjust the `module_write` R function to make sure it uses the macros. (Or maybe use the macros by default but still provide an option to write a module without them?)

4. I think it would be okay (and probably helpful) for the macros to force the C++ name and the map key to be identical. Most of them are already identical, so it's confusing when I encounter one of the few cases where they are different. From memory, I don't recall all the cases where different names are used. I believe at least one of them renames the `temp` quantity (representing ambient air temperature as measured by a weather sensor) to something more useful like `air_temperature`. Probably a better approach would be to just rename this to `air_temperature` everywhere. (`temp` is my least favorite name for a quantity because it is so commonly used in coding as the name for a "temporary" variable.)

5. A potential downside is that module writers might not really understand how a module class is designed "underneath" the macros. This could be explained in documentation, though. There could be an example showing a single module written with and without the macros.

eloch216:
> Maybe we could put the units in the macro too, and store that with the class. I don't know if that's useful.

I think this could be helpful. At the very least, it should (mostly) prevent people from writing a module without specifying units. (I say "mostly" because someone could still put fake or empty units but hopefully they won't!)

Other advantages could be:

1. The `module_info` R function could return units (right now it cannot do that).

2. The `get_quantity_units` script could assemble the units by querying each module via `module_info` rather than scraping the code for unit comments.

3. Checking units could be part of validating a dynamical system. E.g. if one module thinks mass is in kg but another needs it to be in Mg, then they are not really valid when working together. If we did this, I would just want to use simple string comparison. With this test, `kg/s`, `kg / s`, `kg s-1`, and `kg s^(-1)` would all be different units, which would be a little annoying. But making or using an intelligent unit interpreter seems like a lot of work. I would save that issue for 20 years down the road when all other issues have been addressed.

PR #352, opened 2026-09-18
Use peaked Arrhenius temperature response for Vcmax and Jmax

djaiswal:
## Summary

This PR changes the temperature responses of `Vcmax` and `Jmax` in the C3 photosynthesis modules from an Arrhenius equation to a peaked Arrhenius equation (Medlyn et al. 2002), so that these rates can decline above an optimum temperature.

It builds on #273, which added `peaked_arrhenius_response()` to `temperature_response_functions.h` and applied it to other C3 parameters; this PR applies the same function to `Vcmax_norm` and `Jmax_norm` in `c3_temperature_response.cpp`. It also relates to the concerns raised in #272 about consistency between peaked temperature response functions: by reusing the single existing `peaked_arrhenius_response()` helper rather than introducing a new formulation, `Vcmax` and `Jmax` now share the same peaked form as the other parameters.

## Changes

- `BioCro:c3_parameters`, `BioCro:c3_assimilation`, `BioCro:c3_leaf_photosynthesis`, `BioCro:c3_canopy`, and `BioCro:ten_layer_c3_canopy` now use `peaked_arrhenius_response()` for `Vcmax_norm` and `Jmax_norm`.
- **Input quantities renamed:** the modules no longer accept `Vcmax_Ea` and `Jmax_Ea`. They instead require

  | Old input  | New inputs                          |
  |------------|-------------------------------------|
  | `Vcmax_Ea` | `Vcmax_Ha`, `Vcmax_Hd`, `Vcmax_S`   |
  | `Jmax_Ea`  | `Jmax_Ha`, `Jmax_Hd`, `Jmax_S`      |

  where `Ha` is the enthalpy of activation (J / mol), `Hd` the enthalpy of deactivation (J / mol), and `S` the entropy (J / K / mol). Setting `Hd = Inf` and `S = -Inf` makes the deactivation term exactly 1, so the response reduces to the previous Arrhenius equation with `Ha` playing the role of the old `Ea`.

- **Everything in this repository has already been converted.** Every occurrence of `Vcmax_Ea` / `Jmax_Ea` was replaced by `Ha` with the same value, plus `Hd = Inf` and `S = -Inf`. This covers the `soybean`, `soybean_sw`, and `willow` model definitions (`data/*.R`), `man/soybean.Rd`, `tests/testthat/test.c3photo.R`, and the stored module test cases. Because the conversion is exact, the outputs of these crop models are unchanged.

- **Users with their own parameter lists** (outside this repository) that supply `Vcmax_Ea` or `Jmax_Ea` to any of the affected modules will need to make the same substitution: rename `*_Ea` → `*_Ha`, and add `*_Hd = Inf` and `*_S = -Inf` to keep the current behaviour, or supply finite `Hd` and `S` values to enable the high-temperature decline.
- Stored module test cases for the five affected modules were converted to the new inputs. Only the auto-generated placeholder cases (all parameters = 1) have changed expected outputs (by ≤ 0.03 %); all "typical soybean" / "Tobacco" cases are numerically unchanged.
- `NEWS.md` updated under "Minor User-Facing Changes".

Two commits adding stored peaked test cases and a dedicated unit-test file were reverted on this branch (`d230fac1`, `2c46aeee`); they can be restored in a follow-up if we want explicit coverage of the peaked regime.

## Testing

`R CMD check` on the branch (R 4.6.1, macOS arm64): **Status: OK** — 0 errors, 0 warnings, 0 notes; `[ FAIL 0 | WARN 0 | SKIP 2 | PASS 2470 ]` (the 2 skips are pre-existing).

## References

- Medlyn, B. E. et al. (2002) *Plant, Cell & Environment* 25, 1167–1179. https://doi.org/10.1046/j.1365-3040.2002.00891.x
- #273 — More temp responses
- #272 — Potentially inconsistent peaked temperature response functions


justinmcgrath:
The changes look good. The organization of this does not follow the rest of the design though. Parameters are passed through c3_photo to this temperature response. c3_photo needs a value for vcmax and jmax. Those calculations would come before c3_photo and pass those values. This could be in c3_leaf_photosynthesis::do_operation, but if it it were a Module, then the sorts of changes being made here wouldn't force the parameter changes on people when this gets updated, and then it would also be easier to swap temperature response approaches.

justinmcgrath (approved):

justinmcgrath:
Maybe it's not straightforward to make it a module because of the simultaneous equations.

yufenghe22:
What are our thoughts on using the peaked version from Medlyn, B. E. et al. (2002) Plant, Cell & Environment 25, 1167–1179. https://doi.org/10.1046/j.1365-3040.2002.00891.x?

Our default Arrhenius parameters come from Carl's paper, which is not for soybean.

justinmcgrath:
> What are our thoughts on using the peaked version from Medlyn, B. E. et al. (2002) Plant, Cell & Environment 25, 1167–1179. https://doi.org/10.1046/j.1365-3040.2002.00891.x?
> 
> Our default Arrhenius parameters come from Carl's paper, which is not for soybean.

Is this question for me? Carl's paper has the advantage of using a low-Rubisco-content line, so that the entire A-Ci curve is carboxylation-rate limited. But obviously it can't represent a peak, and the species is different. Looking at Harley, Weber and Gates, where Medlyn et al. got their soybean data, their tables 5 and 6 don't show a peaked Vcmax response though. So I don't see how Medlyn et al. fit a peaked equation to it. I know they refit the curves, so maybe that produced different values. At this point though, I'm skeptical that the peaked parameters for soybean represent what happens over 40 degrees C. 

If you want to consider temperatures above 40 degrees C, it would be important to sort that out. However, below that temperature neither of these sources indicate a peaked response, and I suspect it isn't critical which is used. Sensitivity analysis for the particular conditions evaluated would indicate whether it's important, and if so, then you could compare the parameters to see the consequences of using one over the other.

djaiswal:
> What are our thoughts on using the peaked version from Medlyn, B. E. et al. (2002) Plant, Cell & Environment 25, 1167–1179. https://doi.org/10.1046/j.1365-3040.2002.00891.x?
> 
> Our default Arrhenius parameters come from Carl's paper, which is not for soybean.

This change accommodates cassava photosynthesis measurements up to 45°C. Southern India commonly experiences temperatures in this range, and prior measurement limits did not cover it — this PR extends support to capture accurate readings under these field conditions.

justinmcgrath:
I think that the peaked Arrhenius equation is the best for this application, but by chance I came across this approach that hasn't come up yet in these temperature response calculations and I hadn't heard of it, so I thought I'd share.

https://www.sciencedirect.com/science/article/pii/0022519381902460

It's similar to Eyring in that it formulates the response based on biochemistry. The difference is that it includes a low temperature deactivation, which may be useful for species where low temperature inhibits a process beyond simply slowing it.

djaiswal:
> I think that the peaked Arrhenius equation is the best for this application, but by chance I came across this approach that hasn't come up yet in these temperature response calculations and I hadn't heard of it, so I thought I'd share.
> 
> https://www.sciencedirect.com/science/article/pii/0022519381902460
> 
> It's similar to Eyring in that it formulates the response based on biochemistry. The difference is that it includes a low temperature deactivation, which may be useful for species where low temperature inhibits a process beyond simply slowing it.

Could you please share the PDF directly? I am unable to download it from the link.

yufenghe22 (approved):

eloch216:
This looks good to me!

> Two commits adding stored peaked test cases and a dedicated unit-test file were reverted on this branch (`d230fac1`, `2c46aeee`); they can be restored in a follow-up if we want explicit coverage of the peaked regime.

I think it's good practice to add new tests whenever possible, so I took a look at these commits.

I think the dedicated unit-test file is a bit more complicated than necessary, especially since it includes an R implementation of the peaked Arrhenius equation (essentially duplicating C++ code).

I do like the new test cases and would recommend re-adding them. I think this can be merged once the cases have been re-added.

> The organization of this does not follow the rest of the design though.

> Maybe it's not straightforward to make it a module because of the simultaneous equations.

To be clear, the changes in this PR do not alter any of the code organization or design. Our photosynthesis calculations have always used hard-coded temperature responses.

As you mentioned, this is due to the simultaneous equations, and the same issue occurs for other parts of the code. For example, a user cannot swap out the Ball-Berry equation for the Medlyn equation. Right now, there are several models that are all "bundled together" inside the C3 canopy model and cannot be individually swapped out: FvCB model, Penman-Monteith equation, Ball-Berry model, Nikolov model for leaf boundary layer conductance, Stefan-Boltzmann law for long-wave radiation losses, particular equations for temperature response of various parameters. (This is just off the top of my head so may not be exhaustive.)

I think this will always be challenging because the simultaneous equation solver may need different bounds and starting guesses for the unknown parameters depending on which model components are used.

A workaround could be to have several different photosynthesis options. E.g. one with Ball-Berry and peaked Arrhenius, one with Medlyn and peaked Arrhenius, one with Ball-Berry and regular Arrhenius, etc. It's not "truly" modular but would get the job done.

But, addressing this hasn't been a priority so far.

eloch216:
> > I think that the peaked Arrhenius equation is the best for this application, but by chance I came across this approach that hasn't come up yet in these temperature response calculations and I hadn't heard of it, so I thought I'd share.
> > https://www.sciencedirect.com/science/article/pii/0022519381902460
> > It's similar to Eyring in that it formulates the response based on biochemistry. The difference is that it includes a low temperature deactivation, which may be useful for species where low temperature inhibits a process beyond simply slowing it.
> 
> Could you please share the PDF directly? I am unable to download it from the link.

[1-s2.0-0022519381902460-main.pdf](https://github.com/user-attachments/files/32622102/1-s2.0-0022519381902460-main.pdf)

leavesandlemmas:
> Stefan-Boltzmann law for long-wave radiation losses

Are there other models for blackbody radiation losses?

>  think this will always be challenging because the simultaneous equation solver may need different bounds and starting guesses for the unknown parameters depending on which model components are used.

One way around that issue is to have bounds and starting guesses be part of the "problem statement" that the algorithm takes as input. So swapping the ball-berry for the medlyn equation also swaps starting guesses. 

It seems to me that the biggest barrier to using modules for the simultaneous equations is that modules can't accept other modules as inputs (or return as outputs). Otherwise you could have a leaf_photosynthesis_module that requires a stomata model, a biochemical model (fvcb, von caemmerer for c4, etc.) and so on. 

I think the closest workaround is the multilayer canopy modules uses class templates and class inheritance to make a meta module.

eloch216:
> > Stefan-Boltzmann law for long-wave radiation losses
> 
> Are there other models for blackbody radiation losses?

I'm glad someone noticed I included this one in the list.

The Stefan-Boltzmann law is indeed the only way to calculate net blackbody radiation. However, applying the model requires values of the leaf and sky emissivities. These are hard to know, and the emissivity of the sky is particularly variable, depending on the solar zenith angle, cloud cover, humidity, etc. Because of this, at least one situation-specific approximation to the Stefan-Boltzmann model has been developed. Such approximations are based on the same law but may have different functional forms.

I only recently learned about this while documenting the `reference_evapotranspiration()` function, which is part of the new soil evaporation calculations based on DSSAT. See equations 44-45 and other nearby text in the "net long-wave radiation" section from https://doi.org/10.1061/9780784408056.ch04

Right now, I believe we assume both emissivities are 1. I suspect that this may overestimate net longwave radiation, and we may want to try a different approach some day.

